PR #480, opened 2017-07-04
DISCUSSION ONLY : refactor JS code in playQueue view

biconou:
This is only a port of the PR I created on Libresonic to propose a discussion on JS code architecture.


jooola:
@biconou @muff1nman This PR is really getting old, why not close it ?

PR #477, opened 2017-07-04
New JukeBox

biconou:
Guidelines for Contributing
---------------------------

Libresonic development is a community project, and contributions are welcomed. Here are a few guidelines you should follow before submitting:

  1.  **License Acceptance** All contributions must be licensed as [GNU GPLv3](https://github.com/Libresonic/libresonic/blob/develop/LICENSE.txt) to be accepted. Use [`git commit --signoff`](https://jk.gs/git-commit.html) to acknowledge this.
  2.  **No Breakage** New features or changes to existing ones must not degrade the user experience. This means do not introduce bugs, remove functionality, or make large changes to existing themes/UI without prior discussion in an Issue.
  3.  **Coding standards** Language best-practices should be followed, comment generously, and avoid "clever" algorithms. Refactoring existing messes is great, but watch out for breakage.
  4.  **Be bold!** Without contributions, this project will vanish. If you just want to help out, try [submiting a patch](https://github.com/Libresonic/libresonic/issues?q=is%3Aissue+is%3Aopen+label%3Apatches-welcome) for an open Issue.
  5.  **Stay relevant** Issues or commentary that is off-topic or tangential to Libresonic development is subject to moderation. Questions should be focused on improving documentation to solve a problem. Visit [Reddit](https://www.reddit.com/r/libresonic) or [IRC](http://webchat.freenode.net?channels=%23libresonic) for community discussion.

PR #490, opened 2017-07-14
New Jukebox (with rebrand to airsonic)

muff1nman:
cc @biconou 

This PR serves two purposes:

- allows us to spend more time on #477 if necessary by updating it with the rebrand effort. (Sorry @biconou, I haven't had a chance to retest after the latest changes)
- Also proves that we can pull in significant changes that were made prior to #489 with minimal effort.


muff1nman:
@biconou I've tested this and it seems to work other than the fact audio is a bit distorted. But I imagine any issues there will probably need to be done in the library.

Also any reason this doesn't use pulse audio? Just curious more than anything else.

Finally, I was still bothered by the playqueue interface. I think https://github.com/muff1nman/airsonic/pull/1 is a better abstraction. What do you think?

biconou:
@muff1nman "audio is a bit distorted" that sounds weird. Hope we will have any occasion to understand why. For me the sound is perfectly clear and fluent. But for the moment we don't have a lot of beta testers. Wait and see. 
Can you precise your question about PulseAudio ? Did you mean the jukebox does not see PulseAudio server as a playback device ? 
This is an interesting topic but I don't know PulseAudio. And I must confess I also have to investigate more on the exact behavior of JavaSound API in different linux audio configurations.
I think that having a smart jukebox player is a long run job but I am interested in it and it will be very valuable for Airsonic.

Finally, I like your proposition to avoid polluting a domain object with implementing the playlist interface (nice peace of java).
Can you merge it ?

muff1nman:
@biconou merged that one tweak into this PR.

I'm not completely sure, but it seems like it was using the more lower level alsa interface as the device names seemed to be alsa names. I'm not sure it matters, but the first question that popped up in my head when the audio didn't seem quite right is if using the pulse audio server directly would produce different results.

Now with all that said, my setup is a bit strange because I also have my pulse audio system hooked up to qemu to run windows in parallel. So perhaps that was causing issues and I might have to try a different machine..

biconou:
@muff1nman Yes, javasound targets Alsa devices. 
I've read that it exists a bridge that let you see the PulseAudio server as an Alsa device and I suppose you use it to test (right ?).
I've also read that supporting natively PulseAudio in JavaSound is possible and I realized that I am interested in PulseAudio as it could be a solution to play sound remotely (one of my goals).

biconou:
```
    private void updateJukebox(Player player, int offset) throws Exception {
        if (player.getTechnology() == PlayerTechnology.JUKEBOX) {
            jukeboxService.updateJukebox(player, offset);
        } else if (player.getTechnology() == PlayerTechnology.JAVA_JUKEBOX) {
            jukeboxJavaService.updateJukebox(player, offset);
        }
    }
```

muff1nman:
Yes, my thought is that really we're leaking too much from jukeBoxService. Why don't we just have one jukeBoxService instead of two?

biconou:
My Idea (from the beginning) is that the new jukebox technology could replace the older one when it is stable and reach enough (what could take a significantly long time). Thus I decided to create a new jukebox service because the older one is too messy and brought me too many constraints.
Finally I'm still not happy with this new jukebox service and I'd like to see it dead. 
As I explained before, I'm convinced that a real clean solution would involve larger modifications including playQueueService and even Rest API. 

For the moment, we could get rid of  the above messy if-based-code present in PlayQueueService this way : 
Create a "proxy" JukeboxService that delegates to a JavaJukeboxService or a LegacyJukeboxService. 
The new JukeboxService is the only one seen by the PlayQueueService. It should implement play,stop, etc ... functions as in the Rest API. 
In the future, the LegacyJukeboxService will stay as is. The JavaJukeboxService could be improved till it is reach enough to deprecate the LegacyJukeboxService.

What do you think ?

muff1nman:
Yes that sounds like a good plan

biconou:
@muff1nman Ok, I'll implement that and other ideas will come next.

biconou:
@muff1nman I just create a PR to show a refactor of jukeboxService. 
Do you agree with it ? (take a look at JukeboxService and PlayQueueService changes).
If it's ok for you I suggest the following steps. 
- I'll do a lot of tests and fix what would not work
- I'll merge it in master
- I'll improve org.airsonic.player.controller.RESTController#jukeboxControl to avoid using PlayQueueService.

Next step will then be : 
The new jukebox will only be controlled through the rest api (get rid of DWR).

biconou:
I finished the job on this feature. I close this PR. Thanks to @muff1nman for reviewing/supporting/challenging

gerroon:
Is this supposed to work over network via Pulseaudio? The reason I am asking is that I am getting wonky results if the Pulseaudio's default server is a another pc on the network. 

Basically it plays one song and it does not play the next in the same album. I get sound over the network when it plays it. 

I cant test this locally because the server is headless.

thanks

biconou:
The support of pulseaudio is not yet ready. It might work but I have not tested it.

gerroon:
@biconou biconou  it kind of works, it works for the first song but fails if I change to another song, until I restart the server.

PR #491, opened 2017-07-14
Readme history updates

muff1nman:
Requires #489 

Updates history section with reason for forking into airsonic.

Suggestions welcome - I want to keep it as neutral/amiable as possible while still keeping all the important bits. Perhaps we should remove the IRC converstation history? Not sure if its necessary but I think it details best how Libresonic's goals differ from Airsonic's.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,10 +25,20 @@ History
 
 The original [Subsonic](http://www.subsonic.org/) is developed by [Sindre Mehus](mailto:sindre@activeobjects.no). Subsonic was open source through version 6.0-beta1, and closed-source from then onwards.
 
-Airsonic is maintained by [Eugene E. Kashpureff Jr](mailto:eugene@kashpureff.org). It originated as an unofficial("Kang") of Subsonic which did not contain the Licensing code checks present in the official builds. With the announcement of Subsonic's closed-source future, a decision was made to make a full fork and rebrand to Airsonic.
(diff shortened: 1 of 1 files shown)


jooola:
I suggest to remove the IRC conversations... We don't need this neither Eugene !
I don't think this will have any benefits !

README.md
```diff
@@ -25,10 +25,32 @@ History
...
 
-Airsonic is maintained by [Eugene E. Kashpureff Jr](mailto:eugene@kashpureff.org). It originated as an unofficial("Kang") of Subsonic which did not contain the Licensing code checks present in the official builds. With the announcement of Subsonic's closed-source future, a decision was made to make a full fork and rebrand to Airsonic.
-
-Airsonic will strive to maintain compatibility and stability for Subsonic users, including a clean upgrade path. New features and refactoring are welcomed as a Pull Request on GitHub.
-
+Libresonic was created and maintained by [Eugene E. Kashpureff Jr](mailto:eugene@kashpureff.org). It originated as an unofficial("Kang") of Subsonic which did not contain the Licensing code checks present in the official builds. With the announcement of Subsonic's closed-source future, a decision was made to make a full fork and rebrand to Libresonic.
+
+Around July 2017, it was discovered that Eugene had different intentions/goals for the project than some contributors had. An IRC conversation details this best:
+
+```
```
jooola: I think we don't need the IRC history.

CHANGELOG.md
```diff
@@ -62,7 +62,7 @@ v6.1.beta1
 v6.1-alpha1
 -----------
 
-  * Search+replace subsonic-->libresonic
+  * Search+replace subsonic-->airsonic
```
jooola: Something is wrong here ! We need to keep libresonic name in this file !

jooola:
I think this is now a great start, we can always update this with more infos later !

PR #497, opened 2017-07-22
Add project official docker image.

Tapchicoma:
Tag for docker image is set from project version.
To build docker image for the project run:
- mvn package
- mvn -N dockerfile:build

Created docker image can be uploaded to the docker hub:
- mvn -N dockerfile:push

Signed-off-by: Yahor Berdnikau <egorr.berd@gmail.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,27 @@
+FROM alpine:3.6
+
+LABEL description="Airsonic is a free, web-based media streamer, providing ubiquitious access to your music." \
+      url="https://github.com/airsonic/airsonic"
+
+ENV UID=1001 GID=1001 AIRSONIC_PORT=4040 AIRSONIC_DIR=/airsonic
+
+WORKDIR $AIRSONIC_DIR
+
+COPY dockerfiles/run.sh /usr/local/bin/run.sh
+COPY airsonic-main/target/airsonic.war airsonic.war
+
+RUN apk --no-cache add \
+    ffmpeg \
+    lame \
+    su-exec \
+    libressl \
+    ca-certificates \
+    tini \
+    openjdk8-jre \
+    && chmod +x /usr/local/bin/run.sh
+
+EXPOSE $AIRSONIC_PORT
+
+VOLUME $AIRSONIC_DIR/data $AIRSONIC_DIR/music $AIRSONIC_DIR/playlists $AIRSONIC_DIR/podcasts
+
+ENTRYPOINT ["run.sh"]
--- a/dockerfiles/run.sh
+++ b/dockerfiles/run.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+mkdir -p $AIRSONIC_DIR/data/transcode
+ln -s /usr/bin/ffmpeg $AIRSONIC_DIR/data/transcode/ffmpeg
+ln -s /usr/bin/lame $AIRSONIC_DIR/data/transcode/lame
+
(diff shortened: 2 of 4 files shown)


Tapchicoma:
I've took [Wonderfall dockerfile](https://github.com/Wonderfall/dockerfiles/tree/master/libresonic) as base inspiration.

My idea is that on each release also docker image can be pushed to the docker hub.
It also can be used during development to run integration tests.

dockerfiles/run.sh
```diff
@@ -0,0 +1,22 @@
...
+     -Dserver.contextPath=/ \
+     -Dlibresonic.home=$AIRSONIC_DIR/data \
+     -Dlibresonic.host=0.0.0.0 \
+     -Dlibresonic.port=$AIRSONIC_PORT \
+     -Dlibresonic.contextPath=/ \
+     -Dlibresonic.defaultMusicFolder=$AIRSONIC_DIR/musics \
+     -Dlibresonic.defaultPodcastFolder=$AIRSONIC_DIR/podcasts \
+     -Dlibresonic.defaultPlaylistFolder=$AIRSONIC_DIR/playlists \
+     -Djava.awt.headless=true \
+     -jar libresonic.war
```
muff1nman: lets use the new airsonic options here. I.e. `airsonic.home` and `airsonic.defaultMusicFolder`.

Also `libresonic.port`, `libresonic.contenxPath` and `libresonic.host` arent used. Those options are covered by the spring boot `server.port`, `server.contextPath`, etc

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+LABEL description="Airsonic is a free, web-based media streamer, providing ubiquitious access to your music." \
+      url="https://github.com/airsonic/airsonic"
+
+ENV UID=1001 GID=1001 AIRSONIC_PORT=4040 AIRSONIC_DIR=/airsonic
+
+WORKDIR $AIRSONIC_DIR
+
+COPY dockerfiles/run.sh /usr/local/bin/run.sh
+COPY libresonic-main/target/libresonic.war libresonic.war
+
```
muff1nman: Unless I forgot to do it, the maven war should be named `airsonic.war`

dockerfiles/run.sh
Tapchicoma: Sorry, forget to check master branch changes. I was expected that there is still develop branch that has latest changes.

Fixed in updated commit.

Dockerfile
Tapchicoma: Fixed in updated commit.

muff1nman:
Any reason we didn't use https://github.com/spotify/dockerfile-maven? The readme on https://github.com/spotify/docker-maven-plugin states that it is recommended.

Tapchicoma:
ehm, plugin used is `dockerfile-maven`

muff1nman:
ah, oops. I see that now.

I think this is good. Only other thing is that we will need to create the docker hub account.

muff1nman:
The one other thing to think about here, is what the future will look like for all of our supported release targets. I think at some point we considered moving non-war targets to separate repos (https://github.com/Libresonic/libresonic/pull/223). Perhaps we need to do that here? Or maybe we should go back to a submodule per release target paradigm like before? The biggest benefit of submodules I'm seeing is that we could make it easier to perform releases such that everything is built from the same commit.

Tapchicoma:
Don't know about maven (I am not experienced maven user), but in gradle you can create a special task that will build all artifacts and upload them to the desired destinations.

BTW what do you think about migrating to gradle build system?

muff1nman:
See #506 for how I think it would be best to integrate with maven.

Issue #481, opened 2017-07-05
New logo/icon


jooola:
As part of the rebranding process #476  i would like to propose some new logo's and icon's.

===========================
#### Dark
![airsonic_blue_dark_1400x512](https://user-images.githubusercontent.com/19195485/27872857-79d77044-61aa-11e7-992f-9e73f94475e0.png)
![airsonic_dark_1400x512](https://user-images.githubusercontent.com/19195485/27872858-79d91638-61aa-11e7-8f66-930c4069b3be.png)

===========================
#### Light
![airsonic_blue_light_1400x512](https://user-images.githubusercontent.com/19195485/27872868-8220bbde-61aa-11e7-90f3-d178a50fbdd7.png)
![airsonic_light_1400x512](https://user-images.githubusercontent.com/19195485/27872869-822758ea-61aa-11e7-8e8a-894495aafcdd.png)

===========================
#### Icons
![airsonic_blue_512x512](https://user-images.githubusercontent.com/19195485/27872874-8d0a8098-61aa-11e7-8256-a98dbf546c08.png)
![airsonic_dark_512x512](https://user-images.githubusercontent.com/19195485/27872875-8d2904be-61aa-11e7-83a9-2297b9d33a86.png)
![airsonic_red_512x512](https://user-images.githubusercontent.com/19195485/27872876-8d29dd08-61aa-11e7-83eb-87a9433fa79e.png)

===========================
#### Buttons
![airsonic_blue_512x512](https://user-images.githubusercontent.com/19195485/27872946-ca25aaac-61aa-11e7-9364-73ca3439f5f5.png)
![airsonic_dark_512x512](https://user-images.githubusercontent.com/19195485/27872948-ca3f1f00-61aa-11e7-8357-c579defab52e.png)
![airsonic_red_512x512](https://user-images.githubusercontent.com/19195485/27872949-ca45bb3a-61aa-11e7-984b-afb9ba13d1a9.png)

jooola:
Some variant:

![airsonic_blue_dark_1400x512](https://user-images.githubusercontent.com/19195485/27994244-0e1aae4e-64ba-11e7-85f0-5a7193fc3985.png)
![airsonic_blue_light_1400x512](https://user-images.githubusercontent.com/19195485/27994246-0e223010-64ba-11e7-9759-c9adfe322c97.png)
[...]

PR #502, opened 2017-07-31
Rebrand: New logo #481

jooola:
PR to change logo with the voted logo in #481 

Some screenshots : 

capture d ecran de 2017-07-31 17-10-29
capture d ecran de 2017-07-31 17-10-45


muff1nman:
Perhaps one small request: Can you make the png images for the login page and the main logo in the upper left higher res? They could use a bit more crispness I think.

jooola:
Sorry for the PR spam... See #503

PR #506, opened 2017-08-04
Docker maven tweaks

muff1nman:
Builds on Tapchicoma's work in #497 as I wasn't quite happy with the original docker image and also wanted to integrate with maven a bit better.

Specifically, changes I made on the docker side are as follows:

- Removed user manipulations - we can use `-u` flag for that
- Allow user to override command, provide arguments and override java opts
- Reordered some Dockerfile statements to be more condusive to caching

I am running this as follows (note uid 1000 matches my local host uid):

```
mkdir {music,podcasts,playlists,data}
docker run \
	-u 1000 \
	-v `pwd`/music:/airsonic/music:z \
	-v `pwd`/podcasts:/airsonic/podcasts:z \
	-v `pwd`/playlists:/airsonic/playlists:z \
	-v `pwd`/data:/airsonic/data:z fabb0c26b48a
```

Things todo are to add some documentation and integrate this into the travis build steps.


install/docker/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+    libressl \
+    ca-certificates \
+    tini \
+    openjdk8-jre
+
+COPY run.sh /usr/local/bin/run.sh
+
+RUN chmod +x /usr/local/bin/run.sh
+
+COPY target/dependency/airsonic-main-6.3-SNAPSHOT.war airsonic.war
```
Tapchicoma: Just my sidenote: instead of hardcoding version of war here better to have something like this:

```
ARG AIRSONIC_VERSION
COPY target/dependency/airsonic-main-${AIRSONIC_VERSION}.war airsonic.war
```
and set this version via plugin, see plugin [MR](https://github.com/spotify/dockerfile-maven/pull/41)

Also this `AIRSONIC_VERSION` can be used in `LABEL`

muff1nman: Good point. I'll update the pr.

PR #518, opened 2017-08-13
Refactor to remove serviceLocator

bfkelsey:
- This removes all dependencies on deprecated ServiceLocator class.
- Also does some cleanup in the settingsService


muff1nman:
Looks good from a very quick glance. Will review in depth soon.

Issue #519, opened 2017-08-13
Discussion: What do you think of using Kotlin on new backend code?


bfkelsey:
For those that don't know Kotlin is a newer JVM language originally developed by Jetbrains. It has excellent interop with current java code and provides many features to develop cleaner code. Recently it has also become an officially supported language on Android.

Here are some pages to read about it: 
https://kotlinlang.org/docs/reference/server-overview.html
https://www.wired.com/story/kotlin-the-upstart-coding-language-conquering-silicon-valley
http://us12.campaign-archive1.com/home/?u=f39692e245b94f7fb693b6d82&id=93b2272cb6
http://talkingkotlin.com/

biconou:
I am interested in Kotlin but still did not find any spare time to try it. 
Kotlin is also known as beeing able to produce front end applications. Right ? 
At the moment, we are looking for some solution for a new UI.

bfkelsey:
@biconou I don't think Kotlin would be the best tool for a javascript app right now. For that I think typescript would be a better choice. Which after talking to @muff1nman I have started on but waiting on a JSON api to really wire it up. 

Anyways you should try Kotlin, it is pretty easy to learn after using java and has excellent support in intellij.

biconou:
@bfkelsey  Well I'm both interested in TypeScript for frontend and a bit of Kotlin for backend.
As you mentionned, java/kotlin integration is fluent so, I think that, using Kotlin progressively could be possible. 
I am testing a bit of webpack at the moment. Did ya consider using it to package typescript compiled code ?

biconou:
@bfkelsey  What do you exactly mean by 
`waiting on a JSON api to really wire it up`

bfkelsey:
It is using angular cli so it provides a webpack build. While I could use xml in angular it is kind of a pain to set up and since @muff1nman is working on adding a json airsonic api I could wait for that. So currently I have a basic layout and I am just using sample data.

bfkelsey:
I would like to add a few more features to what I have, then I can make the code available somewhere.

Issue #474, opened 2017-06-24
Airsonic rest api


issuemover631:
<a href="https://github.com/muff1nman"><img src="https://avatars2.githubusercontent.com/u/1531528?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [muff1nman](https://github.com/muff1nman)**
_Saturday Jun 24, 2017 at 15:53 GMT_
_Originally opened as https://github.com/Libresonic/libresonic/pull/474_

----

Still a work in progress, but just wanted to open a PR for visibility.

TODO:
- [ ] Use `@RequestParam` instead of `request.getParameter`
- [ ] Add proper documentation on swagger api output
- [ ] Move generated swagger json to something other than /v2/api-docs
- [ ] Perhaps generate swagger ahead of time to avoid having springfox dependencies in the war?
- [ ] Address authentication

----

_**[muff1nman](https://github.com/muff1nman)** included the following code: https://github.com/Libresonic/libresonic/pull/474/commits_

muff1nman:
I am having second thoughts on swagger. This video shares many of my concerns: https://www.youtube.com/watch?v=k5ncCJBarRI&t=3712s. The points against swagger starts 27 minutes in.

muff1nman:
My biggest concern against spring rest docs is that it requires writing tests in order to get the documentation and it seems more tailored to a level 3 rest api rather than our level 2? 1.5 maybe? rest api that doesn't have any tests.

Perhaps this is the time to just do it right, but it will probably take a bit more time...

heyarne:
In order to keep difference between the airsonic and subsonic api to a minimum I have written the following mail to @sindremehus on May 14:

> Hello Sindre Mehus,
> 
> My name is Arne Schlüter and I’m part of the community around Airsonic. A great part of the code of Airsonic stems from your work on Subsonic, which we’re very grateful for.
> 
> I’m writing this mail mostly on my own; while I’ve talked to other people in the community, it’s an open source project without a strict hierarchical leadership, so please take these words as only mine. I do not claim to be representing the Airsonic community, I just think it might be a good idea to talk to you.
> 
> While maintaining the code and extending the codebase we realised that some details in the current API specification could in our eyes be improved. This boils down to two things:
> 
[...]

PR #522, opened 2017-08-16
Start adding api controllers

bfkelsey:
This can be used as a discussion for #519 and if it is desired it can be merged.
- This adds a couple of the current api endpoints as json endpoints.
- This also adds the ability to use kotlin

### DIFF
--- a/airsonic-main/pom.xml
+++ b/airsonic-main/pom.xml
@@ -15,6 +15,7 @@
     <properties>
         <metrics.version>3.1.0</metrics.version>
         <chameleon.version>1.2.0-RELEASE</chameleon.version>
+        <kotlin.version>1.1.4</kotlin.version>
     </properties>
 
     <dependencies>
@@ -279,6 +280,23 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>com.nhaarman</groupId>
+            <artifactId>mockito-kotlin-kt1.1</artifactId>
+            <version>1.5.0</version>
+            <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.jetbrains.kotlin</groupId>
+                    <artifactId>kotlin-reflect</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.jetbrains.kotlin</groupId>
+                    <artifactId>kotlin-stdlib</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
         <!-- UPNP / DLNA -->
         <dependency>
             <groupId>org.fourthline.cling</groupId>
@@ -475,6 +493,32 @@
             <version>2.1.0</version>
             <scope>runtime</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-stdlib-jre8</artifactId>
+            <version>${kotlin.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-stdlib</artifactId>
+            <version>${kotlin.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-reflect</artifactId>
+            <version>${kotlin.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jetbrains</groupId>
+            <artifactId>annotations</artifactId>
+            <version>13.0</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-test-junit</artifactId>
+            <version>${kotlin.version}</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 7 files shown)


biconou:
Sorry for asking but I feel confused (and I like to be part of the discussion).
What about #474 ? 
As I remember, @muff1nman previously started to experiment  something but it seams he gave it up for the moment. Right ? 
Then @bfkelsey is proposing a brand new implementation of a new rest/json api using spring controllers coded in Kotlin. Right ?

airsonic-main/src/main/kotlin/org/airsonic/player/controller/api/MusicFoldersController.kt
```diff
@@ -0,0 +1,29 @@
...
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.ResponseBody
+import javax.servlet.http.HttpServletRequest
+
+@Controller
+@RequestMapping(value = "api/musicfolders",
+        consumes = arrayOf("application/json"),
+        produces = arrayOf("application/json"))
+class MusicFoldersController @Autowired constructor(val securityService: SecurityService,
+                                                    val settingsService: SettingsService) {
```
biconou: Is that a Spring Rest controller implemented in Kotlin ?

biconou:
I like the Idea of having a separate class for each api function instead of the legacy all in one RestController class.

airsonic-main/src/main/kotlin/org/airsonic/player/controller/api/MusicFoldersController.kt
bfkelsey: Yup. It is pretty simple.

bfkelsey:
@biconou I think he was having some trouble with swagger working correctly. 

I added this more as an example of using Kotlin in this project but we could go forward with using these controllers if everyone thinks it is a good solution. Then adding more functionality incrementally.

muff1nman:
Couple thoughts here:

I'm warming up to the idea of using kotlin, but perhaps with a few suggestions:
- Start with it in the tests first so people can play with it in the codebase to see if its the right move before moving it into the actual codebase.
- Don't mix kotlin and java in the same package. Although I'm sure this wouldn't cause a problem technically, I think from a uniformity standpoint it would be good to have packages be either all java or all kotlin. No mixing.

My gut tells me Kotlin is still quite a new language and we should proceed cautiously, but after reading about it for several days it does seem its picking up some steam. If we start with it in the tests, we can work with it there for a while and then reevaluate moving it into the actual code base at a later time.

@bfkelsey I'm still working on the new rest api. Making some slow progress with spring rest docs. Once I have a concrete example, I'll send a PR and you guys can tell me what you think. Also, I don't want to hold you up on getting what you need so here is a suggestion: You can use json to interact with the current subsonic 1.15 endpoint and then when the new rest api is ready, switch over. The assumption is that the new rest api won't be that different initially, but is instead something we'll evolve over time.

@biconou I like the idea of splitting the rest controller up and will take that into account. If we break it up into several controllers, it probably makes sense to move it to a new package.

biconou:
@muff1nman I feel the same about Kotlin.

@muff1nman @bfkelsey I like the Idea that you both are working on a new API in different ways in parallel and we'll merge ideas/techniques and take final architectural decisions in a second time. I think that you are both on the way to propose excellent technical solutions and we don't have to matter the API spec itself for the moment.

airsonic-main/pom.xml
```diff
@@ -538,6 +572,60 @@
...
+                                <sourceDir>${project.basedir}/src/test/kotlin</sourceDir>
+                                <sourceDir>${project.basedir}/src/test/java</sourceDir>
+                            </sourceDirs>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
```
biconou: Is this maven compiler replacement needed by kotlin ?

airsonic-main/src/main/kotlin/org/airsonic/player/controller/api/SystemController.kt
```diff
@@ -0,0 +1,19 @@
+package org.airsonic.player.controller.api
+
+import org.airsonic.player.dto.PingResponse
+import org.springframework.stereotype.Controller
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.ResponseBody
+
+@Controller
```
biconou: Wouldn't it be preferable to use real @RestController as specified by Spring Rest ? (see https://spring.io/guides/gs/rest-service/)

yoyo007:
With @RestController and lib Jackson-databind you could convert objects into json  just by returning the object just from the controller method.

airsonic-main/src/main/kotlin/org/airsonic/player/controller/api/SystemController.kt
bfkelsey: Yeah, that is a good idea.

airsonic-main/pom.xml
bfkelsey: I took that mostly from their sample for setting up a mixed java/kotlin project. I can play around with it to see if it is actually needed.

bfkelsey:
@biconou @muff1nman I guess what I would like to see from a new api would be improving security,  restful design and test coverage. 
For security, getting rid of passing the user and password as url parameters would be a good start. If we want to keep it simple just use basic auth. Otherwise go with some kind of token system.

As far as the actual design it should be improved where /albums returns a list of albums that has an id field where a user can call /albums/{id} and get all of the information for that album.

As far as Kotlin goes I think it is a great language and I really enjoy using it so I would add much more code if we did use it :-)

muff1nman:
@bfkelsey I agree with all of the points you made about the REST api. Any feedback is welcome on my first attempt in #537.

biconou:
@bfkelsey Why closing ?

bfkelsey:
@biconou I just assumed you wanted to go with what @muff1nman was doing in Java.

biconou:
No. I think the experimentation you started is nice and we should keep it as a PR on which i'll probably try to contribute later. For the moment I spend a few time on airsonic but my priority is to deliver a first version of the new jukebox on master. But I planned to study you proposition more precisely. 
Anyway, keeping this PR alive is not a bad thing as long as it is, for the moment, something experimental. 
We have both to be inspired by both muff1nman and you.

PR #537, opened 2017-08-27
[Dot Not Merge] Rest api springdoc

muff1nman:
Demonstrates how [spring rest docs](https://docs.spring.io/spring-restdocs/docs/current/reference/html5/) works. Html documentation is currently generated in `airsonic-main/target/generated-docs` after running a `mvn package`

Things still todo:
- Finish the rest of the api (will probably take a while, but wanted to get some initial feedback first)
- Setup better authentication
- Address any other todos left in the code
- Rebase/squash

Bugs:
- For 204 No Content responses, the documentation shows "null" with a content length of 4.

Although I copied the wsdl generated classes as a starting point of the rest api domain classes, as I started doing this, I think it might be better to just use the actual domain classes directly instead of having an unnecessary layer. I did this for the first converted music folders call and I think it makes sense.


yoyo007:
Personally i like the idea of using the domain classes. 
Since we are using spring security you can do this to get the username.
`Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
String currentPrincipalName = authentication.getName();`

bfkelsey:
This seems a little large for one PR. You may want to break it up into 3 prs. One with the restController refactors. One with the cleanup of the other various controller and finally one to add the new api controllers.
I would also want to break up the restController to multiple controllers. If you don't want to use Kotlin for at least the pojos you should used some kind of annotation processor to simplify those classes.

muff1nman:
@bfkelsey Yes I agree it's a bit large and can definitely split it up into multiple PRs and at the minimum organize the commits. But the biggest thing is I wanted to get some early feedback on if spring rest docs seemed useful before doing it for the whole rest api. So far I just did a couple endpoints.

I'm thinking about not adding all of those classes and just use the original domain packages. I think it would make it simpler and make it easier to move to spring data at some point.

bfkelsey:
@muff1nman It was just pretty hard for me to follow the changes. I will try to pull it down and check out the generated docs.

muff1nman:
@bfkelsey okay let me split it up. Ill also post the generated docs here so everyone can see

bfkelsey:
I was able to pull it down and the docs are pretty nice.

muff1nman:
Here is what it looks like so far. Haven't gotten very far yet, but I think one gets the idea of what everything might look like: https://s3-us-west-2.amazonaws.com/test.andrewdemaria.com/index.html.

It will take quite a bit of time to do everything because you have to write tests in order to generate the documentation.

muff1nman:
Okay I'll keep going down this path then and will start getting some smaller prs created.

muff1nman:
Closing this. Will split into different prs.

PR #538, opened 2017-08-27
Autowire dao service and component scan

yoyo007:
This change was to make all the dao and service type classes use autowired and be component scanned. It simplifies the applicationContext-service.xml down to a few lines. I like this method of working with data access layers and services its easy to make a new service or dao and right away be able to use it in another service.

Guidelines for Contributing
---------------------------

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/JWTSecurityService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/JWTSecurityService.java
@@ -8,6 +8,8 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.time.DateUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
 import org.springframework.web.util.UriComponents;
 import org.springframework.web.util.UriComponentsBuilder;
 
@@ -16,6 +18,7 @@ import java.math.BigInteger;
 import java.security.SecureRandom;
 import java.util.Date;
 
+@Service("jwtSecurityService")
 public class JWTSecurityService {
     private static final Logger logger = LoggerFactory.getLogger(JWTSecurityService.class);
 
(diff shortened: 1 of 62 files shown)


biconou:
I understand the idea but it is a lot of files change at once. Plus it is transversal. 
Thus we need to guaranty no regression.

airsonic-main/src/main/java/org/airsonic/player/service/upnp/ArtistUpnpProcessor.java
```diff
@@ -28,13 +28,15 @@
...
+import org.springframework.stereotype.Service;
 
 import java.util.List;
 
 /**
  * @author Allen Petersen
  * @version $Id$
  */
+@Service
 public class ArtistUpnpProcessor extends UpnpContentProcessor <Artist, Album> {
```
biconou: Is it a @service or a @component ? Same question for all these kind of classes in upnp package.

yoyo007: @service is a @component you really can use either in this case. Service i usually think of as business logic or layer that uses other services or dao/data layer. Component is the more generic. Service is a component and a controller is a component it is just the generic for spring.

yoyo007:
airsonic-main/src/main/java/org/airsonic/player/service/PlayerService.java
@DependsOn("liquibase") 
This is the biggest thing i wasn't sure how to test i am unfamiliar with liquibase.

muff1nman:
@yoyo007 the way to test it is to use `spring-boot:run` with an empty airsonic home. Without that, it was failing to startup.

airsonic-main/src/main/java/org/airsonic/player/service/JWTSecurityService.java
```diff
@@ -25,6 +28,7 @@
     public static final int DEFAULT_DAYS_VALID_FOR = 7;
     private static SecureRandom secureRandom = new SecureRandom();
 
+    @Autowired
     private final SettingsService settingsService;
```
muff1nman: Let's not annotate any final fields as `@Autowired`. This is a newer class where we were using `contructor-arg` so the contructor should be annotated as `@Autowired` not the fields. The spring folks recommend using autowired constructors over autowired fields, so we should move to do that as much as possible.

yoyo007:
@muff1nman In the future did you want to move primarily to constructor autowires?

muff1nman:
I think we should unless someone disagrees.

muff1nman:
Also apparently if you only have one constructor you can omit the autowired annotation. Interesting. https://docs.spring.io/spring/docs/current/spring-framework-reference/htmlsingle/#beans-autowired-annotation

Issue #535, opened 2017-08-26
error messages invisible in Back in Black theme


postags:
![2017-08-25 21_26_54-chillin - x i b sweizy gang ent - airsonic - waterfox](https://user-images.githubusercontent.com/22846348/29745486-4e4a1a28-8a70-11e7-939b-aba49323985d.png)

muff1nman:
Closing - This is a duplicate of #536 and related to #517.

postags:
It's a separate issue, just with the same illustrating image. The text in the dark grey side of the table is not visible unless highlighted.

muff1nman:
Ah I see.

hschroedl:
I played around with this a bit. Basically `head.jsp` has to be included in `error.jsp`. 
In order to force the error page I basically made some other view return the AccessDenied view, e.g. in `StarredController.java`

    return new ModelAndView("accessDenied");

and copied the contents of `error.jsp` in there. Didn't manage to get the error page any other way. Would really like to know how to do that though :man_shrugging:

hschroedl:
@muff1nman Ping, I think this can be closed ;)

PR #547, opened 2017-08-30
Fix error message readability #535

hschroedl:
This pull request addresses #535. The fix basically consists of including` head.jsp` in `error.jsp`. 

image

This has been tested with a couple of other themes (including Default, Midnight, Barent Sea and  Groove)


muff1nman:
Looks good. We just have to be careful we don't have too much logic in `head.jsp` to keep the error page minimal. Looks like right now we're doing okay there.

PR #550, opened 2017-09-02
Rest api cleanups

muff1nman:
Contains various rest api cleanups that were necessary in order to start work on a new rest endpoint.
- Rename rest controller to be clear it is for subsonic
- Change to void return type instead of returning null with ModelAndView where we are writing directly to the http response
- Moved bookmark caching logic into service layer


muff1nman:
I think these changes are pretty safe.

PR #553, opened 2017-09-02
Initial setup of rest api

muff1nman:
Hopefully this is a more review-able chunk of code than #537. The idea here is to implement all of the endpoints one by one and document them using spring rest docs rather than copy over the existing subsonic api directly.

Biggest thing I still need to do is figure out a good security method. I don't think we want to use the existing subsonic rest security scheme.

cc @bfkelsey @yoyo007

PR #564, opened 2017-09-11
Proposition based on #553 restapi initial with springrestdocs

biconou:
This PR proposes to test the effect of @RestController annotation. 
Also adds some response for the api/index resource.


muff1nman:
I'd rather use ResponseEntity as it allows more control over response codes.

Issue #455, opened 2017-06-07
Ability to ignore album subfolders such as "scans"


issuemover631:
<a href="https://github.com/codicusmaximus"><img src="https://avatars1.githubusercontent.com/u/11328458?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [codicusmaximus](https://github.com/codicusmaximus)**
_Wednesday Jun 07, 2017 at 02:34 GMT_
_Originally opened as https://github.com/Libresonic/libresonic/issues/455_

----

Feature request - add the ability in the settings to define a list of folder names / wildcards that should be ignored.  The issue is that when an Album folder contains a folder named "scans" or "extras" it shows up as a clickable item in both the Web UI and most phone clients yet when you click them they contain no music.  It might also be a good idea to simply not show folders if they contain no music.

jooola:
This can now be accomplish using the exclude pattern tool

PR #571, opened 2017-09-15
Adds a settings option to ignore symlinks during the media scan.

akpdw:
For issue 570.

Check is done for all children of the root folder on down, so if the root folder itself is a symlink, everything will still work.


biconou:
Does it work on both windows and linux ?

akpdw:
Files.isSymbolicLink() is supposed to work for symlinks in both Unix and Windows filesystems, but I haven't actually tested it anywhere but under Linux. I could probably test it out, though that would mean actually booting into Windows. :)

muff1nman:
Whats the need for symlinks in your media folders? Just not following the use case.

akpdw:
The use case is a legacy media library that is used by multiple people.  We have a bunch of files that were ripped inconsistently (and/And, spaces vs underscores, etc), and then a bunch of playlists that were made from those locations. But then it was a pain to remember which artist directory to look into for which album, so we consolidated the actual album and music files into single artist directories. But since that would break the playlists, we also symlinked the old artist directories to the new, consolidated directories.

So I can't removed the symlinks or it would break other people's old playlists. But if airsonic honors the symlinks, then it ends up having multiple copies of the symlinked albums.

cascooscuro:
This is my particular use case for symlinking.  I want the music folder structure to follow this format:
/[artist]/(year)-(album name)-(media type)-(catalog number)-(encoding)

I don't want to modify the original folders, so what I do is create the the structure with symlinks (using the metadata) pointing to the original files/folders. Airsonic is only aware of the symlinked structure.

So please, keep support for both by using some setting.

Thanks.

akpdw:
@biconou Have confirmed it works on both Windows and Linux.

@cascooscuro I have it as a top-level setting (as opposed to a per-media-folder setting), and by default it's disabled.  So if you don't do anything, then the current behavior is kept.

biconou:
I think this PR is worth merging if it is actually needed by some people. Having a wide range of ability to customize music library organization is great.
I'll start reviewing/testing this PR to merge it. I have some critics on your code.
See the review.

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -228,7 +229,7 @@ private void doScanLibrary() {
     }
 
     private void scanFile(MediaFile file, MusicFolder musicFolder, Date lastScanned,
-                          Map<String, Integer> albumCount, Genres genres, boolean isPodcast) {
+                          Map<String, Integer> albumCount, Genres genres, boolean isPodcast, boolean ignoreSymLinks) {
```
biconou: I don't like the fact that you added one more argument to the scanFile method. 
In this special case it seams useless as you can invoque the settings service from inside the scanFile method code bloc. I've worked a lot analyzing the structure of this service and my thoughts is it is messy. One of my futures projects would be to simplify it and introduce a plugin aware code organisation. Adding more and more arguments to methods generally leads to more mess in code.
Another suggestion I would make is to add some debug code to indicate that a special file/directory is skipped due to symlink.

akpdw: Eh, I just did it as a knee-jerk optimization in case settingsService.getIgnoreSymLinks() was a more expensive call than it should be. I tested it out and it's not, so no problem switching it over to your preferred syntax.

And added in the requested LOG.debug() calls.

muff1nman:
I'm not a fan of this pr as it seems too specific to be of use to anyone else. I highly doubt anyone else needs this as is. I could perhaps see a more general implementation where we allow users to plug in custom filters for whether to include a file or not, but that might require getting our plug-in architecture setup

akpdw:
My impression is that wanting to ignore symlinks is more common than you might think, but then again, it's a feature that I need, and therefore I've looked around online before for people asking that question.  So my impression may be a little biased.  :)

Anyway, though, it's easy to generalize a bit.  For instance, https://github.com/airsonic/airsonic/issues/455 exists where someone wanted to ignore specific subfolders.  I added an excludePattern option to handle that case.  (And also moved both that and the symlink method to the isExcluded() method mentioned in that bug, which I'd missed before because it's in MediaFileService instead of MediaScannerService where I'd have expected it to be, but whatever.)

muff1nman:
Okay with those latest changes this is a much more reasonable change with a bit of flexibility to it.

airsonic-main/src/main/java/org/airsonic/player/service/SettingsService.java
```diff
@@ -729,6 +737,37 @@ public void setSortAlbumsByYear(boolean b) {
...
+    private void compileExcludePattern() {
+        if (getExcludePatternString() != null && getExcludePatternString().trim().length() > 0) {
+            excludePattern = Pattern.compile(getExcludePatternString());
+            excludePatternCompiled = true;
+        }
+    }
+
+    public Pattern getExcludePattern() {
+        if (! excludePatternCompiled) {
+            compileExcludePattern();
```
muff1nman: Can't this be simplified to just a null check on excludePattern? The excludePatternCompiled boolean is redundant as currently written. One valid reason to keep the flag is if you set it in the case the pattern string is null or there is an error compiling it so as to avoid unnecessarily recompiling, but that would require changes to compileExcludePattern.

airsonic-main/src/main/java/org/airsonic/player/service/SettingsService.java
```diff
@@ -180,6 +183,8 @@
     private static final String DEFAULT_SONOS_SERVICE_NAME = "Airsonic";
     private static final int DEFAULT_SONOS_SERVICE_ID = 242;
     private static final String DEFAULT_EXPORT_PLAYLIST_FORMAT = "m3u";
+    private static final boolean DEFAULT_IGNORE_SYMLINKS = false;
+    private static final String DEFAULT_EXCLUDE_PATTERN_STRING = "";
```
muff1nman: default should be null.

akpdw:
I think that should take care of the nitpicks.  (The compile call is necessary in both getExcludePattern() and setExcludePatternString() because setExcludePatternString() is called only when the value is updated, not if it's already set at start time. That's what the excludePatternCompiled boolean was checking for before, but this version gets around it.)

muff1nman:
Looks good. Please rebase, squash and sign-off.

muff1nman:
Actually, can you make one additional change - it seems a better spot for the UI setting would be on the media folders page instead of on the General page.

akpdw:
Ok.  Moved the UI to the media folders page and everything (I think) should be rebased properly.

PR #590, opened 2017-10-07
Use `ffprobe` and not `ffmpeg` to scrape video metadata

### LINKED ISSUE #117: Use ffprobe for metadata discovery
<a href="https://github.com/tari"></img></a> **Issue by [tari](https://github.com/tari)**
_Saturday Jul 30, 2016 at 09:04 GMT_
_Originally opened as https://github.com/Libresonic/libresonic/issues/117_

----

The ffmpeg metadata parser currently runs `ffmpeg` with an input but no output file specified, which causes ffmpeg to bail out but only after printing some file information. The `ffprobe` command is provided by ffmpeg to probe the same information, but doesn't require incorrect invocation to avoid reading the entire file.

`ffprobe` also provides options for selecting the output format (it's meant to be used by external programs) which may be useful if the default ffmpeg output format changes in the future.

tari:
Fixes #117.

This changes behavior by searching for `ffprobe` in the transcode directory and falling back to using `PATH` to locate `ffprobe` if it doesn't exist in the transcode directory, beyond the obvious "uses `ffprobe` instead of `ffmpeg`".

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/metadata/FFmpegParser.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/metadata/FFmpegParser.java
@@ -19,11 +19,11 @@
  */
 package org.airsonic.player.service.metadata;
 
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.airsonic.player.domain.MediaFile;
-import org.airsonic.player.io.InputStreamReaderThread;
 import org.airsonic.player.service.SettingsService;
 import org.airsonic.player.service.TranscodingService;
-import org.airsonic.player.util.StringUtil;
 import org.apache.commons.io.FilenameUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


muff1nman:
What about people who dont have ffprobe, but have ffmpeg?

tari:
It should be uncommon to have a system in which ffmpeg is available, but not ffprobe- I'm not aware of any binary distributions that include the former only, and searching `PATH` will work fine without any extra work in most cases.

Some users (like on Windows in particular) might need to get an ffprobe binary, but that should be easy- I've just checked the Windows binary package [recommended by the docs](https://airsonic.github.io/docs/transcode/#on-windows) and it includes ffprobe for instance.

biconou:
@tari  Can you teach us what is the advantage of using ffprobe instead of ffmpeg ?

jooola:
@biconou https://github.com/airsonic/airsonic/issues/117#issue-240476789

jooola:
Need some documentation changes right ?
@tari Are you up adding some lines to the documentation repo ?

tari:
Yeah, I'll plan to write some docs in short order.

biconou:
For your information : 
A ffmpeg/ffprobe java wrapper
https://github.com/bramp/ffmpeg-cli-wrapper

jooola:
Sorry if this is totally offtopic.
@biconou would you like to create yourself  a matrix riot account, and join the airsonic channel ?
You are actually missing a lot about airsonic discussion !

biconou:
done

Issue #536, opened 2017-08-26
search breaks when the session times out


postags:
![2017-08-25 21_26_54-chillin - x i b sweizy gang ent - airsonic - waterfox](https://user-images.githubusercontent.com/22846348/29745492-79c190be-8a70-11e7-8dfb-137f9f63ef62.png)

Airsonic encountered an internal error. You can report this error in the Airsonic Forum. Please include the information below. 

```
Status     403 
Error     Forbidden Message Invalid CSRF Token '5c0b766f-42b3-4a58-b08d-2d8a389c706a' was found on the request parameter '_csrf' or header 'X-CSRF-TOKEN'. 
Path     /airsonic/search.view 
Time     Sat Aug 26 04:23:56 PDT 2017 
Exception
Java version Oracle Corporation 1.8.0_131 
Operating system Linux 4.10.0-32-generic Server Apache Tomcat/8.0.38 (Ubuntu) 
Memory Used 371 of 601 MB
```

Any search query started throwing this error, although everything else was working fine. The session time-out assessment is per plastikman187 on Reddit.

muff1nman:
I think this is related to #517

evertvp:
This is indeed the same bug as #517 as far as I can tell, but I don't think the behaviour should be to log out. The only problem is that the csrf token has expired, not the user session.

A refresh of the top frame fixes this.

muff1nman:
Okay, so while a refresh of the top frame would indeed get around this, perhaps the best thing to do here is to change /search from a GET to a POST. It doesn't change any state anyways so shouldn't require CSRF protection.

jooola:
This looks like

PR #606, opened 2017-10-29
Don't require csrf for search endpoint

### LINKED ISSUE #517: No autologout
1. Opened airsonic in a browser
2. Put notebook in a sleep mode for 20h
3. Wakeup notebook, open airsonic tab in browser and saw following error:

```
Status | 403
-- | --
Error | Forbidden
Message | Invalid CSRF Token 'e2fca6a2-faab-4e67-b832-bbee932fb787' was found on the request parameter '_csrf' or header 'X-CSRF-TOKEN'.
Path | /search.view
Time | Thu Aug 10 16:36:32 GMT 2017
Exception |  
Java version | Oracle Corporation 1.8.0_131
Operating system | Linux 4.9.16-gentoo_v1
Server | Apache Tomcat/8.5.11
Memory | Used 101 of 240 MB
```

I think proper behaviour in such case should be autologout and opening login page.

muff1nman:
Reasoning:

- It doesn't change state and is not a sensitive endpoint
- It really should be changed to GET but that is a bit more intrusive
  change that can be done at another time
- The search csrf token is stored on the top.jsp page for a long time.
  If the user keeps this tab open for a while it is possible the csrf
  token will change on their session with other requests going on such
  that the search csrf token becomes wrong/stale.

Fixes #517 and #536. Supersedes #552  

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/security/CsrfSecurityRequestMatcher.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/security/CsrfSecurityRequestMatcher.java
@@ -6,6 +6,8 @@ import org.springframework.stereotype.Component;
 
 import javax.servlet.http.HttpServletRequest;
 
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.regex.Pattern;
 
 /**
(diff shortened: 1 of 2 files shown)


jooola:
Is this less secure ?

By removing the token check we open the endpoint to the world ? Or am I totally not understanding this PR ?

muff1nman:
https://www.owasp.org/index.php/Cross-Site_Request_Forgery_(CSRF)_Prevention_Cheat_Sheet

> CSRF attacks specifically target state-changing requests, not theft of data, since the attacker has no way to see the response to the forged request.

The endpoint is still protected. CSRF is basically something that allows an attacker to trick people into submitting requests who are already logged in. In this case, we don't really care about if an attacker forces a user to make a search because it doesn't change any state on the server side. An example of a case where this is important is for example the change password form. You definitely would not want an attacker to be able to trick the user into changing their password into something the attacker could supply.

Issue #359, opened 2017-04-22
Added a property to ignore file timestamps when reindexing


issuemover631:
<a href="https://github.com/muff1nman"><img src="https://avatars2.githubusercontent.com/u/1531528?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [muff1nman](https://github.com/muff1nman)**
_Saturday Apr 22, 2017 at 16:13 GMT_
_Originally opened as https://github.com/Libresonic/libresonic/pull/358_

----

#### Background
I accidentally deleted most of my music directory. The database was still intact. I recovered the music directory by rolling back to a previous ZFS snapshot and performed a reindex. However, libresonic did not mark the deleted files as present. Turns out the file timestamp was unchanged through the ZFS restore, and so libresonic still thought the last indexing effort was still "good".

#### Changes
This adds the option to ignore file timestamps when scanning files. This can be helpful in the case of a restore as described above. There might be a better way to do this, as this was really a quick effort on my part to fix my own libresonic.

#### TODO
This does not add a UI, just a single property that can be turned on by editing the lilbresonic.properties file.

Signed-off-by: Andrew DeMaria <lostonamountain@gmail.com>

----

_**[muff1nman](https://github.com/muff1nman)** included the following code: https://github.com/Libresonic/libresonic/pull/358/commits_

eharris:
I also think this should be exposed in the admin interface, either as a one-time option when doing a manual scan/update of the library, or as an option that could be enabled for the periodic/scheduled library updates.

While I think that striving to minimize disk/system impact when performing a whole library rescan is a good goal, it seems to me that Airsonic doesn't do a very good job at being able to detect when its database is out of sync, and automatically rescan just a small portion of the library.  For instance, when I navigate in the web UI to a specific folder that I know has changes that Airsonic doesn't reflect (#1083), it still assumes its view of the library is correct rather than doing even the smallest checking to make sure that it actually is.  I can understand always bypassing doing any checks if the "Fast Access Mode" option is enabled, but when it's not, I think it should be doing a lot more in the way of lightweight and limited-scope checking to verify that the database is still consistent with reality.

jvoisin:
I do agree with @eharris , a "Force full rescan" button would be nice.

eharris:
It should be noted that cover art updating is another area that needs significant improvement, since rescans appear to ignore changes in cover art under pretty much all circumstances. See #649

eharris:
[...]

PR #607, opened 2017-10-30
Option to disable timestamps during scan

muff1nman:
I accidentally deleted most of my music directory. The database was
still intact. I recovered the music directory by rolling back to a
previous ZFS snapshot and performed a reindex. However, libresonic did
not mark the deleted files as present. Turns out the file timestamp was
unchanged through the ZFS restore, and so libresonic still thought the
last indexing effort was still "good".

This adds the option to ignore file timestamps when scanning files. This
can be helpful in the case of a restore as described above. There might
be a better way to do this, as this was really a quick effort on my part
to fix my own libresonic.

This does not add a UI, just a single property that can be turned on by
editing the lilbresonic.properties file.

@fxthomas suggested this could instead be a query parameter on the
initial issue #359. That would basically move the potential UI to the
scan page. That would be fine, but I could imagine there might be cases
where people want this setting on all the time.

Also perhaps this could be obsoleted by the larger change to improve scanning in general.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/dao/AlbumDao.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/dao/AlbumDao.java
@@ -334,12 +334,12 @@ public class AlbumDao extends AbstractDao {
     }
 
     public void markNonPresent(Date lastScanned) {
-        int minId = queryForInt("select min(id) from album where last_scanned != ? and present", 0, lastScanned);
-        int maxId = queryForInt("select max(id) from album where last_scanned != ? and present", 0, lastScanned);
+        int minId = queryForInt("select min(id) from album where last_scanned < ? and present", 0, lastScanned);
+        int maxId = queryForInt("select max(id) from album where last_scanned < ? and present", 0, lastScanned);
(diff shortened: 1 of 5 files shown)


jooola:
> Also perhaps this could be obsoleted by the larger change to improve scanning in general.

Should we consider each specific request or build larger tools that could be used by most ? This question goes also to the #567 PR.

I'm not against building specific tools for each problems ( it would anyway help some people ) but if we invest time in such things we may never see our big scan system re-factor.
But since I don't do much in the back-end, this is just a small opinon.

jvoisin:
This isn't a major change, it also improves a bit the code, I think that we should merge it, and if someone wants to add a parameter or a GUI element, be my guest.

fxthomas:
I think that we should add this to the UI. It's something we can point people at when having scanning issues, it definitely needs to be more discoverable.

eharris:
I think this should be an option in the UI as well.  Either as a single-pass recan-now option, or a toggle that can be enabled, force rescan, disable, so that if some users want it on always they can.

Issue #594, opened 2017-10-08
Mixed content errors with reverse proxy and HTTPS


tari:
Running Airsonic 10.0.1 on Linux behind an Apache reverse proxy; some views attempt to load over http and are blocked as mixed content, whereas others (like About or artist/album views) are not.

For instance, clicking Settings I get a warning in my browser console (domains changed for public consumption):

> Blocked loading mixed active content http://example.net/subsonic/musicFolderSettings

Inspecting the request as logged by the browser, the server is indeed returning a Location with HTTP:

> GET https://example.net/subsonic/settings.view? [HTTP/1.1 302  26ms]
> Location http://example.net/subsonic/musicFolderSettings

Solutions like #435 don't seem to help; my server configuration looks like this (some irrelevant parts removed):

```
Listen 443
<VirtualHost *:443>
        SSLEngine on
        SSLCertificateFile "/etc/letsencrypt/live/..."
        SSLCertificateKeyFile "/etc/letsencrypt/live/..."
        ProxyRequests off
        ProxyPreserveHost on
        SSLProxyEngine on

        Header always set Strict-Transport-Security "max-age=604800"
        RequestHeader set X-Forwarded-Proto "https"

        ProxyPass /subsonic/ http://localhost:4040/subsonic/
        ProxyPassReverse /subsonic/ http://localhost:4040/subsonic/
</VirtualHost>
```

And `server.contextPath` is set to `/subsonic`.

jooola:
Could quickly test if your context path is correctly set ?
https://airsonic.github.io/docs/configure/standalone/#context-path

tari:
I notice the docs say to set `server.context-path` and I had `server.contextPath`, but they appear to behave the same. Checking the command line of my running server, it is set as expected:

```
java -Xmx512m -Dserver.address=127.0.0.1 -Dserver.port=4040 -Dserver.httpsPort=0 -Dserver.context-path=/subsonic ...
```

No change in behavior if I set `context-path` or `contextPath` (I guess Spring treats them equivalently?).

muff1nman:
To debug proxy issues you can increase the logging on `org.airsonic.player.service.NetworkService`. One way to do this is using an additional java system property `-Dlogging.level.org.airsonic.player.service.NetworkService=trace`.

Assuming you're a developer type, you can also just set a debugger in the NetworkService and see what happens in there.

tari:
[...]

PR #628, opened 2017-11-15
Default server.use-forward-headers=true

tari:
Fix for #594. I'm not certain what this will do if not behind a proxy, but don't think it should be an issue. May be worth testing though.

### DIFF
--- a/contrib/airsonic.service
+++ b/contrib/airsonic.service
@@ -17,6 +17,7 @@ ExecStart=/usr/bin/java \
           -Dairsonic.home=${AIRSONIC_HOME} \
           -Dserver.context-path=${CONTEXT_PATH} \
           -Dserver.port=${PORT} \
+          -Dserver.use-forward-headers=true \
           -jar ${JAVA_JAR} $JAVA_ARGS
 User=airsonic
 Group=airsonic
--- a/install/docker/run.sh
+++ b/install/docker/run.sh
@@ -16,6 +16,7 @@ if [[ $# -lt 1 ]] || [[ ! "$1" == "java"* ]]; then
      -Dserver.host=0.0.0.0 \
      -Dserver.port=$AIRSONIC_PORT \
      -Dserver.contextPath=$CONTEXT_PATH \
+     -Dserver.use-forward-headers=true \
      -Dairsonic.home=$AIRSONIC_DIR/data \
      -Dairsonic.defaultMusicFolder=$AIRSONIC_DIR/music \
      -Dairsonic.defaultPodcastFolder=$AIRSONIC_DIR/podcasts \


muff1nman:
If this is really needed, it should be configured by default in the shipped `application.properties`. I'm still looking to understand why explicit configuration of that setting is needed though.

muff1nman:
This should not be needed. I tested httpd with ssl and it seems to be working just fine. No mixed content errors or anything.

```
Listen 443 https
<VirtualHost *:443>
        SSLEngine on
        SSLCertificateFile "/etc/httpd/conf.d/test.crt"
        SSLCertificateKeyFile "/etc/httpd/conf.d/test.key"
        ProxyRequests off
        ProxyPreserveHost on
        SSLProxyEngine on

        RequestHeader set X-Forwarded-Proto "https"

        ProxyPass "/airsonic" "http://localhost:8080/airsonic"
        ProxyPassReverse "/airsonic" "http://localhost:8080/airsonic"
</VirtualHost>
```

Is your airsonic actually localhost or did you change it when posting on #594

tari:
It is actually localhost; I'm not sure why ProxyPassReverse alone isn't sufficient, but it seems like I'm not the only one with that problem so it's probably not just weird configuration on my end.

I was hesitant to set this in application.properties because it may do the wrong thing in some situations- if not behind a proxy a client may be able to set those headers directly and cause wrong redirects (which may not be a problem if it's not a security issue).

muff1nman:
Can you please create clear reproduce steps for #594? I have yet to see a case where there was an actual issue with the proxies that was not a configuration issue.

jolny:
Without the "RequestHeader set X-Forwarded-Proto "https"" line, I got these mixed content errors.
But you mean the above configuration, without use-forward-headers, should suffice?

synchrone:
@muff1nman I have a guess that maybe you're testing on localhost domain, for which the browser ignores https -> http redirect in an iframe. I'm trying to deploy standalone Airsonic behind an nginx I don't control

@tari normally X-Forwarded headers are only respected when set by a whiitelisted server (like nginx [set_real_ip_from](http://nginx.org/en/docs/http/ngx_http_realip_module.html#set_real_ip_from), and any security implications only make sense for X-Forwarded-For anyway (since that's what apps trust to be the correct origin ip)

People have been fighting over this simple servlet container configuration option since [way back](https://github.com/airsonic/airsonic/commit/4d6ca84ec6b4e4f4503cb21f0c1fc10145a55695#diff-a9a452901374e197ac59af71f41ac4e1R132), often implementing horrible hacks (like this one in HAProxy, while it was only 1 config option to just make Spring+Tomcat/Jetty to properly implement HTTP Proxy protocol.

tari:
@muff1nman I did some more investigation and it's dependent on `ProxyPreserveHost` being enabled. More detail in [my comment there](https://github.com/airsonic/airsonic/issues/594#issuecomment-348645518); not using this setting but still honoring X-Forwarded headers in other contexts seems inconsistent and confusing, supported by the observation that I'm not the only user to experience this problem.

muff1nman:
@synchrone That's helpful to know. Yes I have been testing on a single host. I'm just trying to get this reproduced so I can follow along with the pains that are being experienced. Let's continue the discussion in #594 and leave this pr alone for a minute.

muff1nman:
As per my comment in https://github.com/airsonic/airsonic/issues/594#issuecomment-352199029 this shouldn't be needed anymore as long as we stick with Tomcat.

jooola:
Using this with standalone behind haproxy fixed the few broken links ( in the right pane for the currenlty played songs and the stream url finally has a HTTPS )

PR #643, opened 2017-12-03
Language rework

comradekingu:
Are apostrophes supposed to be escaped with "'" ?

---

"random songs when play queue" is very unclear.
So is "User table Quote"

----

The "wap" strings appears outdated, does it still work?

I seem to remember it is called URL cloaking, not "URL rewriting"

"main.wiki=<table class="detail" is not supposed to be there (?)


Lightsockie:
I'm really glad someone is doing this. As a pedant, the weird phrasing and other oddities were brutal. 

But "email" is objectively more clear and consistent with common parlance than "e-mail", grandpa.

jooola:
> Are apostrophes supposed to be escaped with "'" ?
[Absolutely](https://stackoverflow.com/questions/4015345/how-to-properly-escape-quotes-inside-html-attributes/4015380#4015380)

muff1nman:
Awesome job cleaning this up! My only hesitation is that we are not very good/consistent about escaping these strings where they are used. Sometimes the html is interpreted, sometimes it is escaped with c:out so it makes it hard for the non-developer to make changes to these files. Have you tested these changes to make sure the escaping is correct?

comradekingu:
@Lightsockie "E-mail" makes for fewer errors in translation, seeing as in other languages, the equivalent of "email" is an error, whereas "e-mail" isn't.
@jooola I started, and then i figured it was another issue for another time.
@muff1nman It is all over the place and then some. Bugreport to follow. This is ready to go AFAIK.

airsonic-main/src/main/resources/org/airsonic/player/i18n/ResourceBundle_en.properties
```diff
@@ -61,15 +61,15 @@ top.podcast=Podcasts
...
 top.missing=No media folders found. Please change the settings.
 top.logout=Log out {0}
 # left.jsp
-left.scanning=Scanning media folders...
+left.scanning=Scanning media folders…
 left.statistics={0}&nbsp;artists<br>{1}&nbsp;albums<br>{2}&nbsp;songs<br>{3}<br>{4}&nbsp;hours
 left.shortcut=Shortcuts
 left.playlists=Playlists
-left.radio=Internet TV/radio
+left.radio=Internet TV & radio
```
muff1nman: Needs to be escaped.

airsonic-main/src/main/resources/org/airsonic/player/i18n/ResourceBundle_en.properties
```diff
@@ -288,7 +288,7 @@ settingsheader.general=General
 settingsheader.advanced=Advanced
 settingsheader.personal=Personal
 settingsheader.musicFolder=Media folders
-settingsheader.internetRadio=Internet TV/radio
+settingsheader.internetRadio=Internet TV & radio
```
muff1nman: Please escape this

airsonic-main/src/main/resources/org/airsonic/player/i18n/ResourceBundle_en.properties
```diff
@@ -414,19 +414,19 @@ transcodingsettings.step1=Step 1
...
+transcodingsettings.info=<p class="detail">(%s = The file to be transcoded, %b = Max bitrate of the player, %t = Title, %a = Artist, %l = Album)</p> <p>Transcoding is the process of converting from one media format to another. {1}''s transcoding engine allows streaming media that would not otherwise be streamable. It is performed on-the-fly and doesn''t require any extra storage space.<p/> <p>The transcoding is done by third-party command line program, which must be installed in {0}. You may add your own custom transcoder given it fulfills the following requirements: <ul> <li>Command line interface.</li> <li>Ability to send output to stdout.</li> <li>If used in step 2, ability to read input from stdin.</li> </ul> </p> <p> Note that transcoding is activated on a per-player basis from <b>Settings → Players</b>.</p>
 # internetRadioSettings.jsp
 internetradiosettings.streamurl=Stream URL
 internetradiosettings.homepageurl=Homepage
 internetradiosettings.name=Name
 internetradiosettings.enabled=Enabled
-internetradiosettings.add=Add Internet TV/radio
+internetradiosettings.add=Add Internet TV & radio
```
muff1nman: Same here.

PR #647, opened 2017-12-10
Issue 617 - fixes DLNA RecentAlbum browsing 

akpdw:
Removes an incorrect override in AlbumUpnpProcessor which was preventing
RecentAlbumUpnpProcessor from behaving correctly.

Also adds a correct getAllItemsSize() implementation for RecentAlbumUpnpProcessor.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/upnp/AlbumUpnpProcessor.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/upnp/AlbumUpnpProcessor.java
@@ -26,10 +26,8 @@ import org.airsonic.player.domain.CoverArtScheme;
 import org.airsonic.player.domain.MediaFile;
 import org.airsonic.player.domain.MusicFolder;
 import org.airsonic.player.service.SearchService;
-import org.fourthline.cling.support.model.BrowseResult;
 import org.fourthline.cling.support.model.DIDLContent;
 import org.fourthline.cling.support.model.PersonWithRole;
-import org.fourthline.cling.support.model.SortCriterion;
 import org.fourthline.cling.support.model.container.Container;
 import org.fourthline.cling.support.model.container.MusicAlbum;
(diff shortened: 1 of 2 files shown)


airsonic-main/src/main/java/org/airsonic/player/service/upnp/AlbumUpnpProcessor.java
```diff
@@ -61,20 +59,6 @@ public AlbumUpnpProcessor() {
...
-        DIDLContent didl = new DIDLContent();
-
-        List<MusicFolder> allFolders = getDispatchingContentDirectory().getSettingsService().getAllMusicFolders();
-        List<Album> selectedItems = getAlbumDao().getAlphabeticalAlbums((int) firstResult, (int) maxResults, false, true, allFolders);
-        for (Album item : selectedItems) {
-            addItem(didl, item);
-        }
-
-        return createBrowseResult(didl, (int) didl.getCount(), getAllItemsSize());
-    }
```
muff1nman: Just to make sure I understand, the above is just a code cleanup as it does nothing functionally different than the base class, correct?

akpdw: Oh wait--hold off on this. Now I know what happened; I had thought that for some weird reason I'd left in some old un-refactored code which was breaking the RecentAlbumUpnpProcessor. But now that I'm looking more closely I realize that this was actually an optimization that had been put in...which was breaking the RecentAlbumUpnpProcessor.

This fix just removes the optimization, which works but isn't, erm, optimal.  The right thing to do is to move the optimization so it doesn't break its subclasses.

akpdw:
Ok, this restores the optimization in the AlbumProcessor and just re-overrides it in the RecentAlbums subclass.

muff1nman:
Ah okay I understand now. Thanks for making it clearer.

PR #655, opened 2017-12-16
Ensure login url is sent to the correct url

muff1nman:


jvoisin:
What is this PR doing?

almet:
This PR seems to do a forward to the loginFormUrl using the RequestDispatcher, instead of a 302 redirect.

I don't quite understand why it's needed, though. @muff1nman could you please explain what you had in mind?

muff1nman:
This was something I was testing around the time that #594 was getting debugged. The thought was that it would better handle redirect to the https endpoint when behind a proxy. However, I never got to definitively answer whether this was needed or not.

almet:
Then, we probably want to close this issue? To me, it seems that proxy configuration is a good way to mitigate this.

PR #660, opened 2017-12-21
 Change page encoding from iso-8859-1 to utf-8 

jooola:
After changing page encoding from iso-8859-1 to utf-8, everything worked fine except this : 
image
But already fixed !

I got some logs but I don't think it is related to this change 
[airsonic.log](https://github.com/airsonic/airsonic/files/1580279/airsonic.log)


muff1nman:
Merged #674 in place of this.

jooola:
@muff1nman Not sure if patching this problem using #674 fix is really clean. Means I have to keep the script to convert special chars to Unicode in 18n resources. Maybe we should reconsider this PR ?

Issue #380, opened 2017-05-04
External Player only checks for Flash


issuemover631:
<a href="https://github.com/Lightsockie"><img src="https://avatars0.githubusercontent.com/u/1125399?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [Lightsockie](https://github.com/Lightsockie)**
_Thursday May 04, 2017 at 03:32 GMT_
_Originally opened as https://github.com/Libresonic/libresonic/issues/379_

----

From some discussion in the IRC, it seems the external player created from share links only checks for flash, bypassing jwplayer completely.

As seen [here](https://github.com/Libresonic/libresonic/blob/4d365f4024ca9548470357174c36b1c842d8c091/libresonic-main/src/main/webapp/WEB-INF/jsp/externalPlayer.jsp#L7):
`/src/main/webapp/WEB-INF/jsp/externalPlayer.jsp`
```html
<%--@elvariable id="model" type="java.util.Map"--%>
<%@ page language="java" contentType="text/html; charset=utf-8" pageEncoding="iso-8859-1" %>

<html>
<head>
    <%@ include file="head.jsp" %>
    <script type="text/javascript" src="<c:url value="/script/swfobject.js"/>"></script>
    <script type="text/javascript" src="<c:url value="/script/prototype.js"/>"></script>

```

jooola:
cc @airsonic/maintainers What the state of our current html5 player ? Does the external player also have an html5 player ?

muff1nman:
See #448 for the todo to migrate the external player to MediaElement. Once that is done, this can be resolved.

jvoisin:
This was fixed by #962

PR #696, opened 2018-02-21
Initial effort to replace Flash with HTML5 audio in external player

jpsutton:
This is my attempt to replace Flash in the external player with HTML5 audio (not using mediaelement.js at this time).  It has basic functionality, but currently has the following limitations:

1) no track listing as in the previous flash-based UI.
2) no ability to change tracks manually (tracks will play in order, sequentially).
3) because of limits in the HTML5 audio tag and the way that transcoded audio is delivered, some audio has an incorrect time shown in the player.

This is NOT a complete implementation.  I could use some help getting this complete before it (potentially) goes into a release.

Mentioning #380, as it's related to this effort.

### DIFF
--- a/airsonic-main/src/main/webapp/WEB-INF/jsp/externalPlayer.jsp
+++ b/airsonic-main/src/main/webapp/WEB-INF/jsp/externalPlayer.jsp
@@ -15,51 +15,52 @@
     </c:if>
 
     <script type="text/javascript">
+        var curTrack = 0;
+        var playlist = new Array();
+
         function init() {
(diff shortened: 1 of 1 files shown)


jooola:
Nice this is great we will get rid of the flash player someday !

> This is my attempt to replace Flash in the external player with HTML5 audio (not using mediaelement.js at this time)

Why not using mediaelement.js directly ?

jpsutton:
Just because I was doing a quick fix on my own server, and I'm not familiar with mediaelement.js.

PR #699, opened 2018-02-27
Fix #596

### LINKED ISSUE #596: SubStandard theme includes old Subsonic Logo
Under Settings | General, selecting the "SubStandard" theme reveals the old Subsonic "submarine" logo as the background image.  I'm not sure if use of this image presents any issues (the main reason for me to point it out), but this theme should probably be renamed and changed to use the new logo (or removed entirely) in any case.

jooola:
Replace the old background with new airsonic themed background.


jooola:
image

PR #727, opened 2018-04-04
Catch ClientAbortException exception to avoid huge traceback fix #724

### LINKED ISSUE #724: Huge traceback in the logs from normal Connection reset by peer IO exception
## Problem description

Whenever a connection is broken by a client (which is pretty common), a huge traceback of about 300 lines and 40k bytes is generated in the logs, but is largely useless.

The log message block starts with:
```
org.apache.catalina.connector.ClientAbortException: java.io.IOException: Connection reset by peer
```

It seems like it would be worth suppressing the huge traceback for such a common and expected condition.  It makes the logs huge for no useful reason and makes it harder to see and debug other real issues.

### Steps to reproduce

Stream a file from a client and skip to the next track before all the data from the last one has transferred, or from a client where you can otherwise force a premature close of the connection.

## System information

 * **Airsonic version**: 10.1.1-RELEASE – December 16, 2017
 * **Operating system**: Debian stretch, Apache Tomcat/8.5.14 (Debian), java 1.8.0_162, Linux
 * **Proxy server**: None
 * **Client**: Any

Bonome:
A one line log replace the full backtrace

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/StreamController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/StreamController.java
@@ -29,6 +29,7 @@ import org.airsonic.player.service.sonos.SonosHelper;
 import org.airsonic.player.util.HttpRange;
 import org.airsonic.player.util.StringUtil;
 import org.airsonic.player.util.Util;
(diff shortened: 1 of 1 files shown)


muff1nman:
Thanks for the PR!

Issue #164, opened 2016-12-04
Support MusicBrainz tags


issuemover631:
<a href="https://github.com/fxthomas"><img src="https://avatars0.githubusercontent.com/u/613594?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [fxthomas](https://github.com/fxthomas)**
_Sunday Dec 04, 2016 at 11:25 GMT_
_Originally opened as https://github.com/Libresonic/libresonic/issues/163_

----

[MusicBrainz](http://musicbrainz.org/) is a great open option for auto-tagging music files in a library. Several tagging software make use of their services and write tags to the music files containing unique IDs that precisely identify albums and tracks. Among others, their own [Picard](https://picard.musicbrainz.org/) as well as [Beets](http://beets.io/) work very well.

If it exists, can we add a link to the related MusicBrainz entry on each album (next to the search links)?

tesshucom:
I was worried about where to write, but I'll leave a footprint here.
My clone is already modernized with new tag support.

|tag name |tag id|Subsonic/Airsonic |Jpsonic |Music Center (SONY) |itunes (APPLE)
|:---|:---|:---:|:---:|:---:|:---:|
|title |TIT2 |● |● |● |●
|title sort|TSOT | |● |● |●
|artist|TPE1 |● |● |● |●
|artist sort|TSOP | |● |● |●
|album |TALB |● |● |● |●
|album sort|TSOA | |● |● |●
|album artist|TPE2 |● |● |● |●
|album artist sort|TSO2 | |● |● |●
|genre|TCON |● |● |● |●
|Release year|TYER |● |● |● |●
|composer|TCOM | |● |● |●
|composer sort|TSOC | |● | |●
|track no|TRCK |● |● |● |●
|disk no|TPOS |● |● |● |●

The reason is that the index/sort/search in Japanese is not possible with Airsonic tags alone.
Japanese is a linguistically unique language (comparable to Eskimo), so it is not suitable for machine processing.
However, correct processing is possible by following the specifications of SONY / APPLE as described above.
This is a reasonable anti-pattern that is useful for a wide range of non-English speaking support.

If you extend the supported tags, we'll definitely see performance degradation.
Therefore, careful tag item selection that can convince the user is necessary.
(It's not a light addition)

My opinions about MusicBrainz are on the side of EugeneKay.

> but we need to make sure not to screw up the basics ;-). 

The Subsonic API is originally a file structure based system.
[...]

PR #731, opened 2018-04-18
Show link to MusicBrainz release on album pages

fxthomas:
Hello all, long time no see!

I've been holding onto this for a long time, and finally managed to finish testing it. This PR shows a link to the MusicBrainz release in the album page if at least one track has the corresponding tag :

164-mb-support-album

It does so by reading the `MUSICBRAINZ_RELEASE_ID` tag and storing it inside a new column in the database :

164-mb-support-db

See related issue: #164


muff1nman:
Nice - I'll try to review here in the next week

airsonic-main/src/main/java/org/airsonic/player/controller/MainController.java
```diff
@@ -134,6 +134,7 @@ protected ModelAndView handleRequestInternal(@RequestParam(name = "showAll", req
             map.put("sieblingAlbums", sieblingAlbums);
             map.put("artist", guessArtist(children));
             map.put("album", guessAlbum(children));
+            map.put("mbReleaseId", guessMbReleaseId(children));
```
bfkelsey: Maybe something like musicBrainzId would be better for naming of this element?

muff1nman: Well we have to be careful because there could also be Music Brainz Track ID, Music Brainz Artist ID, Music Brainz Disc ID, etc. So I think we need 'releaseId' at a minimum. `mbReleaseId` or `musicBrainzReleaseId`. Best thing would probably be `musicBrainzReleaseId` in the code base and `mb_release_id` for the database column name.

fxthomas: `musicBrainzReleaseId' sounds good!

fxthomas:
Thanks for the review! I figured using "everything in MediaFile" like the rest of the code would reduce the effort if we change the database structure.

I updated the attribute names, and will add a test case when I can.

airsonic-main/src/main/webapp/WEB-INF/jsp/albumMain.jsp
```diff
@@ -262,6 +262,11 @@
         <span class="header"><a target="_blank" href="${wikipediaUrl}">Wikipedia</a></span> |
         <span class="header"><a target="_blank" href="${allmusicUrl}">allmusic</a></span> |
         <span class="header"><a target="_blank" href="${lastFmUrl}">Last.fm</a></span> |
+        <c:if test="${not empty model.mbReleaseId}">
+          <sub:url value="https://musicbrainz.org/release/${model.mbReleaseId}" var="musicBrainzUrl" encoding="UTF-8">
```
muff1nman: model.mbReleaseId needs to be updated to the new variable as well, right?

fxthomas: Yup, that should have been the case :)

fxthomas:
I just finished fixing my last commit, and added tests for reading MusicBrainz tags. This should be ready for merging!

PR #734, opened 2018-04-28
Extract controllers and ui to site module

bfkelsey:
- This will modularize the application more and make it easier to add
other web apis without affecting existing apis.

Signed-off-by: Ben Kelsey <bfkelsey@gmail.com>


jooola:
@bfkelsey Should `airsonic/airsonic-main/src/main/resources/org/airsonic/player` also move to `airsonic-site` ?

jooola:
@bfkelsey Did you run the war with the latest commit ? Just wondering if you haven't deleted too much ? Or are those files duplicate already ?

bfkelsey:
Those files are in the new site module too. I didn't run this latest commit
but the previous ones worked without any breaks. I can try it later.

muff1nman:
@bfkelsey Splitting the webui out into the site module makes a ton of sense, but what do you envision as the end goal here? My concern is that the airsonic-main module is now kinda useless by itself as this PR is written. I'm wondering if it makes more sense to try to split things up such that there are two war files, one for the API and one for the existing webui. Then with that, I image there would be an airsonic-core module with anything shared between the two such as the domain classes. 

Additionally I'm a bit concerned about some of the non-trivial changes that are mixed in here that should be separate PRs. E.g. the upgrade of spring boot, other dependency changes, the change of packaging type to jar from war (which I would need some convincing on).

Finally is there something you have in mind that is dependent on this large change? I am just trying to understand the benefits to compare with the risk of a change like this especially considering we are just barely stabilizing after the last major change (rebrand).

bfkelsey:
@muff1nman If you feel this is too much of a change I can just work off a fork for a while or merge into a dev or future branch until new ui/api is complete.

muff1nman:
> Changing the main module to jar was so that it can be used as a library and the site module uses war packaging so it would be required in this pr.

Ah okay makes sense now.

> My main goal of this is to provide a completely separate api that can be used by the new ui that is being built. 

We can easily add new apis (see the start of one effort at https://github.com/airsonic/airsonic/pull/553). That shouldn't be a problem as long as they don't have the same uri structure.

> The benefit of this is that someone could attach any front-end onto the application. It will also allow separating out the different existing front-ends eg web-ui or subsonic-api.

I think separating out the subsonic-api would confuse users more than anything else. In my mind, it would be nice for users to always be able to deploy a war and know that any clients dependent on the subsonic api will still work. In the event that they're using an airsonic unique client, that client should be able to use a new api that is deployed alongside of the existing subsonic api.

bfkelsey:
I think it is better to separate out the different interfaces into different packages to increase maintainability and reduce coupling. As for users, right now building it or deploying will not change. They will still build a single war and install it. In the future it could be possible to make a choice where they could build in only certain user interfaces.

muff1nman:
> could be possible to make a choice where they could build in only certain **user interfaces**

I agree that at some point we will want to split out the current web ui and in general increase maintainability and reduce coupling. But again, I think we should keep the subsonic **API** in whatever "base" package we envision. 

I think we both want something similar but to more concretely layout how I would like to see the line drawn is:

- Headless WAR (provides subsonic API and any new airsonic API, contains Application.java, REST controllers, services, dao, etc)
- Legacy UI WAR (contains JSPs, controllers, services and other java specific to the legacy frontend)
- New UI WAR (basically just compiled JS files?)

I dont see a need for a jar - maybe I could see a minimal one containing just the domain classes in the case someone wanted to build something in Java and wanted to deserialize the REST json..

jooola:
> - Headless WAR (provides subsonic API and any new airsonic API, contains Application.java, REST controllers, services, dao, etc)
> - Legacy UI WAR (contains JSPs, controllers, services and other java specific to the legacy frontend)
> - New UI WAR (basically just compiled JS files?)

I would also see that in the long run !

IMO we need the Airsonic API as soon as possible since I personally don't want to build the new UI on the subsonic API. And this PR seems to dig a nice whole to plant our new tree (airsonic API).
What I mean by that I would prioritize the airsonic API and kinda plug the subsonic API in it, even if we know that they will be 2 parallel services.

bfkelsey:
@muff1nman I think we are on the same page on what has to be done, maybe not how though. I think the backend and all of the interfaces (subsonic api, legacy web ui, new api) should be in different modules. You can then have another module that will pull everything together and output a war.

muff1nman:
> I think the backend and all of the interfaces (subsonic api, legacy web ui, new api) should be in different modules

Maven modules? Yes I could see that being a good end goal. But I think we have a lot of cleanup work that needs to be done first before actually separating things. I.e. first clean up the package structure and creating the proper logical separations via java package changes. 

Honestly though, I'd don't see this as an important item right now especially because there isn't anything that is blocked/required by a reorg like this, it doesn't provide any immediate benefits, has quite a potential for regressions, and hampers other ongoing work. So can we table doing this for now?

bfkelsey:
Okay we can put it on hold.

PR #747, opened 2018-05-22
Handle player id as an Integer instead of String

snw35:
Change player id to an Integer. This matches the internal type to the database, fixes type-casting issues when Postgres is used, and eliminates the need to have `stringtype=unspecified` on your JDBC connection URL.

Tested on:
MariaDB 10.2
Postgres 9.5
Default sqlite

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/dao/TranscodingDao.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/dao/TranscodingDao.java
@@ -59,7 +59,7 @@ public class TranscodingDao extends AbstractDao {
      * @param playerId The player ID.
      * @return All active transcodings for the player.
      */
-    public List<Transcoding> getTranscodingsForPlayer(String playerId) {
+    public List<Transcoding> getTranscodingsForPlayer(Integer playerId) {
         String sql = "select " + QUERY_COLUMNS + " from transcoding2, player_transcoding2 " +
                      "where player_transcoding2.player_id = ? " +
                      "and   player_transcoding2.transcoding_id = transcoding2.id";
@@ -72,7 +72,7 @@ public class TranscodingDao extends AbstractDao {
      * @param playerId       The player ID.
      * @param transcodingIds ID's of the active transcodings.
      */
-    public void setTranscodingsForPlayer(String playerId, int[] transcodingIds) {
+    public void setTranscodingsForPlayer(Integer playerId, int[] transcodingIds) {
         update("delete from player_transcoding2 where player_id = ?", playerId);
         String sql = "insert into player_transcoding2(player_id, transcoding_id) values (?, ?)";
         for (int transcodingId : transcodingIds) {
--- a/airsonic-main/src/main/java/org/airsonic/player/service/PlayerService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/PlayerService.java
@@ -30,6 +30,7 @@ import org.apache.commons.lang.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.DependsOn;
 import org.springframework.stereotype.Service;
+import org.springframework.web.bind.ServletRequestUtils;
 
 import javax.annotation.PostConstruct;
 import javax.servlet.http.Cookie;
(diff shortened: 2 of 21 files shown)


airsonic-main/src/main/java/org/airsonic/player/dao/TranscodingDao.java
```diff
@@ -93,6 +97,9 @@ public void createTranscoding(Transcoding transcoding) {
         }
         transcoding.setId(existingMax + 1);
         String sql = "insert into transcoding2 (" + QUERY_COLUMNS + ") values (" + questionMarks(QUERY_COLUMNS) + ")";
+
+
+
         update(sql, transcoding.getId(), transcoding.getName(), transcoding.getSourceFormats(),
```
muff1nman: Can you kill this extra white space?

airsonic-main/src/main/java/org/airsonic/player/dao/TranscodingDao.java
```diff
@@ -72,11 +72,15 @@
...
      * @param transcodingIds ID's of the active transcodings.
      */
-    public void setTranscodingsForPlayer(String playerId, int[] transcodingIds) {
-        update("delete from player_transcoding2 where player_id = ?", playerId);
-        String sql = "insert into player_transcoding2(player_id, transcoding_id) values (?, ?)";
-        for (int transcodingId : transcodingIds) {
-            update(sql, playerId, transcodingId);
+    public void setTranscodingsForPlayer(Integer playerId, int[] transcodingIds) {
+        if (playerId == null) {
+            return;
```
muff1nman: Hm, this seems like we should ensure we aren't calling this method with a null playerId rather than adding a check here.

airsonic-main/src/main/java/org/airsonic/player/service/PlayerService.java
```diff
@@ -88,11 +90,16 @@ public synchronized Player getPlayer(HttpServletRequest request, HttpServletResp
                                          boolean remoteControlEnabled, boolean isStreamRequest) {
 
         // Find by 'player' request parameter.
-        Player player = getPlayerById(request.getParameter("player"));
+        Player player;
+        try {
+            player = getPlayerById(ServletRequestUtils.getIntParameter(request, "player"));
+        } catch (ServletRequestBindingException e) {
+            player = null;
```
muff1nman: This makes me a bit uneasy of silently throwing away this error here. Seems like the we should fix the exception getting thrown.

airsonic-main/src/main/java/org/airsonic/player/service/PlayerService.java
```diff
@@ -118,9 +125,9 @@ public synchronized Player getPlayer(HttpServletRequest request, HttpServletResp
         if (player == null) {
             player = new Player();
             createPlayer(player);
-//            LOG.debug("Created player " + player.getId() + " (remoteControlEnabled: " + remoteControlEnabled +
-//                      ", isStreamRequest: " + isStreamRequest + ", username: " + username +
-//                      ", ip: " + request.getRemoteAddr() + ").");
+//          LOG.debug("Created player " + player.getId() + " (remoteControlEnabled: " + remoteControlEnabled +
+//              ", isStreamRequest: " + isStreamRequest + ", username: " + username +
+//              ", ip: " + request.getRemoteAddr() + ").");
         }
```
muff1nman: Please discard this change.

muff1nman:
I'll have to dig into that travis error a bit more.

airsonic-main/src/main/java/org/airsonic/player/dao/TranscodingDao.java
snw35: No prob.

airsonic-main/src/main/java/org/airsonic/player/dao/TranscodingDao.java
snw35: Agreed - will search where it is called from and see if I can add checks there.

airsonic-main/src/main/java/org/airsonic/player/service/PlayerService.java
snw35: Yes, I wasn't sure about that, I'll add 'throws Exception' after 'getPlayer' and remove the inline check if that works.

airsonic-main/src/main/java/org/airsonic/player/service/PlayerService.java
snw35: Forgot to reset that after I'd used it for debug - will do.

muff1nman:
This looks pretty clean now. Can you please squash? I'm going to do some testing as well on this.

snw35:
Thanks - rebased and squashed. Let me know if anything comes up in testing.

Issue #723, opened 2018-03-22
Web player cuts off tracks based on expected rather than actual data size


eharris:
## Problem description

In my testing of a workaround for doing automatic replaygain via transcoding (see issue #525), I discovered that the web app seems to be cutting off tracks early despite there being more data to play.  This was tested and confirmed both on Firefox (Debian jessie 52.7.2esr-1~deb9u1) and Chrome (Version 64.0.3282.186 (Official Build) (64-bit) on OS X).

Testing with an actual Subsonic player that uses the API does not exhibit the problem.  Only the web player seems to be affected.

### Steps to reproduce

1. Change your user to have a bitrate limit of 32kbps in the user settings.
2. Change the downsample command on the transcoding page to replace the variable `%bk` with a fixed higher bitrate, like `128k`.  If you were previously using the stock ffmpeg transcode, the new command line for downsample would be `ffmpeg -i %s -map 0:0 -b:a 128k -v 0 -f mp3 -`
3. Use the web interface to play an mp3 file that has a bitrate higher than the limit configured in step 1 (so that it will get transcoded).
4. Notice that with the given values, the song will only play for 1/4 of the time (32k/128k), and the duration shown in the web player is also 1/4 of the actual song duration.  If you use 32k and 64k, the song will be cutoff 1/2 way in.

## System information

 * **Airsonic version**: 10.1.1-RELEASE – December 16, 2017
 * **Operating system**: Debian Linux (jessie)
 * **Proxy server**: None
 * **Client**: Firefox (Debian jessie 52.7.2esr-1~deb9u1) and Chrome (Version 64.0.3282.186 (Official Build) (64-bit) on OS X).

## Additional notes

I was using this config (an unusually low bitrate limit) in order to force airsonic to always transcode the files but only for particular users so that I could use the transcoding callout to apply replay gain to the audio.  Again, see #525 for more details on why.  I don't know of another way to force certain specific users playback to always go through a transcode.  Is there another way?

ae-g-i-s:
I'm not completely sure, but this is probably related, unless somebody fixed this issue in Airsonic: https://github.com/tsquillario/Jamstash/issues/131#issuecomment-349507059

The server will sometimes send an incorrect Content-Range header and simply refuse to deliver the remaining bytes (because they don't exist). DSub for Android doesn't care, and neither does my hacky bash client, because they just grab the entire file and don't request a specific range. Browsers do care because they try to be smart about chunked transfers.

eharris:
[...]

Issue #548, opened 2017-08-31
opus streaming/transcoding to HTML5 player not working


postags:
There seems to be an issue with streaming an opus transcode to the HTML5 player. Starting a track & initiating a transcode throws airsonic for this loop:

```
2017-08-31 15:48:58.160  INFO --- o.a.p.io.InputStreamReaderThread         : (/var/airsonic/transcode/opusenc) Skipping chunk of type "LIST", length 150
2017-08-31 15:49:24.169  INFO --- o.a.p.c.StreamController                 : Got HTTP range: 2523136-
2017-08-31 15:49:24.170  INFO --- o.a.p.io.PlayQueueInputStream            : greg listening to "Earth Grid (2011)/01 Openings.mp3"
2017-08-31 15:49:24.172  INFO --- o.a.p.io.TranscodeInputStream            : Starting transcoder: [/var/airsonic/transcode/ffmpeg] [-i] [/media/Warehouse/music/current/Zomes/Earth Grid (2011)/01 Openings.mp3] [-ab] [128k] [-v] [0] [-f] [wav] [-] 
2017-08-31 15:49:24.174  INFO --- o.a.p.io.TranscodeInputStream            : Starting transcoder: [/var/airsonic/transcode/opusenc] [--quiet] [--bitrate] [128] [--downmix-stereo] [--title] [Openings] [--album] [Earth Grid] [--artist] [Zomes] [-] [-] 
2017-08-31 15:49:24.322  INFO --- o.a.p.io.InputStreamReaderThread         : (/var/airsonic/transcode/opusenc) Skipping chunk of type "LIST", length 150
2017-08-31 15:49:50.433  INFO --- o.a.p.c.StreamController                 : Got HTTP range: 2523136-
2017-08-31 15:49:50.434  INFO --- o.a.p.io.PlayQueueInputStream            : greg listening to "Earth Grid (2011)/01 Openings.mp3"
2017-08-31 15:49:50.438  INFO --- o.a.p.io.TranscodeInputStream            : Starting transcoder: [/var/airsonic/transcode/ffmpeg] [-i] [/media/Warehouse/music/current/Zomes/Earth Grid (2011)/01 Openings.mp3] [-ab] [128k] [-v] [0] [-f] [wav] [-] 
[...]

PR #759, opened 2018-06-20
Add option to disable seeking on transcodes. (Mitigates #548 & #723)

WillyPillow:
As per #548, #723, and tsquillario/Jamstash#131, the current method of
estimating `Content-Length` creates various problems.

However, if headers such as `Accept-Ranges` is omitted, clients will only
use the first connection, which is `Transfer-Encoding: chunked`, and no
`Content-Length` is necessary.

Doing this has the side effect that (at least on the web player) seeking
to a specific time is no longer possible, thus this was made an opt-in
option.

Signed-off-by: WillyPillow <wp@nerde.pw>

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/CoverArtController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/CoverArtController.java
@@ -309,7 +309,7 @@ public class CoverArtController implements LastModified {
         VideoTranscodingSettings videoSettings = new VideoTranscodingSettings(width, height, offset, 0, false);
         TranscodingService.Parameters parameters = new TranscodingService.Parameters(mediaFile, videoSettings);
         String command = settingsService.getVideoImageCommand();
-        parameters.setTranscoding(new Transcoding(null, null, null, null, command, null, null, false));
+        parameters.setTranscoding(new Transcoding(null, null, null, null, command, null, null, false, true));
         return transcodingService.getTranscodedInputStream(parameters);
     }
 
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/StreamController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/StreamController.java
@@ -152,19 +152,20 @@ public class StreamController  {
                 playQueue.addFiles(true, file);
                 player.setPlayQueue(playQueue);
 
-                if (!file.isVideo()) {
+                TranscodingService.Parameters parameters = transcodingService.getParameters(file, player, maxBitRate, preferredTargetFormat, null);
+
+                if (parameters.isEnableSeek() && !file.isVideo()) {
                     response.setIntHeader("ETag", file.getId());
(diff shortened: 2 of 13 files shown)


WillyPillow:
Observed effects on different clients:
- Web player: Shows "Live Broadcast", unable to seek
- Chrome (external player): Progress bar does not move, unable to seek
- Firefox (external player): Can seek in buffered region
- Mpv: Can seek in buffered region
- Ultrasonic: Same behavior as before, can only seek after whole file is downloaded

It might be possible to modify the web player's source code to allow mpv-like behavior, but judging from Chrome's behavior above, I suspect that it wouldn't be trivial.

tari:
I'm working on a more general fix to streaming making bad assumptions about transcoded sizes (https://github.com/airsonic/airsonic/issues/685#issuecomment-409778446), so that may be more useful if I can make it happen in a shortish amount of time.

mlim15:
Thank you for this fix - my use case is exclusively streaming at very low bitrates (32Kbps), and this change allows me to stream VBR opus transcodes through the browser and external players once I use the new option to disable seeking.

I have tested with the following players and can confirm the following behaviour (as above):
- DSub on Android - Functionality unchanged, works and only seeks after entire file is streamed
- Airsonic web on Firefox - Works and allows seeking what has streamed in so far
- Airsonic web on Chrome - Works and says "Live Broadcast" instead of offering seeking
- Jamstash on Firefox - Works and allows seeking what has streamed in so far
- Jamstash on Chrome - Works but does not allow seeking, progress bar does not move

Seeking is not important to me so all of the above outcomes are fine with me - just happy I can finally stream more than 3 seconds of an opus transcode in the browser.

tari:
I think so too; this approach is the most pessimistic way to correctly handle VBR output, leaving room for improvements as I've suggested on #685.

muff1nman:
Looking through this again, I'm not sure it makes sense to have this as an option for each transcoding. I imagine it will be that you either want this enabled across the board or not at all. In addition, I would hope that this option could be removed at some point when this gets fixed permanently so adding a database field for the transcoding table doesn't make sense.

WillyPillow:
Because I run a server for multiple friends, my original intent was to have fine-tuned controls to fit different use cases. (But in the end they usually just stream lossless anyway, so it's not like it's used :/ )

When I wrote the patch, I thought we could just do an additional dropColumn in Liquibase if this were to be removed, though admittedly there would still be an redundant XML file laying around.

If you guys want the simpler version I can probably cook up a patch in the next few days though.

muff1nman:
@WillyPillow yes a dropColumn is easy in theory, but in practice I think if we can avoid modifying the database it is preferable. I'm always a proponent for simplicity :)

Let me know if I can help with the changes if you get busy or something.

WillyPillow:
The new patch is done. I thought it would be easier to start from scratch, so I'll close this in favor of #798 .

PR #765, opened 2018-07-25
Issue #105 Add support for multiple genres in id3 tags

### LINKED ISSUE #105: Add support for multiple genres
<a href="https://github.com/tiagoad"></img></a> **Issue by [tiagoad](https://github.com/tiagoad)**
_Sunday Jun 12, 2016 at 10:56 GMT_
_Originally opened as https://github.com/Libresonic/libresonic/issues/105_

----

It would be great if subsonic would support music with multiple genre tags.
I believe there's a fork with that feature implemented.

For reference (never answered by the subsonic team): http://forum.subsonic.org/forum/viewtopic.php?f=3&t=15455

devonbryant:
Signed-off-by: Devon Bryant <devonbryant@gmail.com>

This is a PR for supporting multiple genres in id3 tags.  I left `genre` as a string field on media files and albums (same behavior as before) so it shouldn't affect any existing clients.  The primary change is two new join tables: album to genre and media_file to genre.  Searches and lookups go through the tie tables instead of the string field.  The splitting of the genre string happens during scans/ingestion and the delimiter is configurable via properties.

Fixes #105


muff1nman:
Thanks for the PR! I will review soon

peioe:
Will this work with multiple genres tags in FLAC/vorbis too ? That would be awesome.

adylise:
Multi-genre support is pretty essential for me, so getting this to work with vorbis comments and id3 tags would be really nice. :)

biconou:
@devonbryant 
I suggest that you give more informations on what is done in this PR (bullet points list in the PR description).
Is this PR save for the Rest API ? What do you think ?

kaysond:
I'm working on testing this (rebased to the latest). So far everything seems to work fine. Media gets imported properly with the genres separated according to my specified delimiter (I used ";" which is different from the default).

I also tried it out with UltraSonic on Android, and it appears to work seamlessly in that the genres are correctly displayed, but there's a weird issue with albums not showing up correctly. Unsure if that's an UltraSonic thing or Airsonic API thing, but the behavior is the same as the current release.

There are a couple of issues I think need to be addressed:

1. The delimiter option should be added to the UI
2. If you remove your libraries and re-add them to get the new parsing, you get this horribleness: https://imgur.com/V6LmjhE where all the old, concatenated genres are still there but empty. I think this is related to #796 but we should make sure there's an easy way to remove old empty genres.

I'll keep using it for a while to see if anything comes up. Please let me know if you'd like to see anything else specific tested.

airsonic-main/src/main/resources/liquibase/6.5/add-multiple-genres-join.xml
```diff
@@ -0,0 +1,88 @@
+<databaseChangeLog
```
biconou: I think that the liquibase/6.5 is not correct. I'm not sure but it should be named after the target airsonic version. 
@muff1nman

airsonic-main/src/main/java/org/airsonic/player/dao/GenreDao.java
```diff
@@ -0,0 +1,102 @@
+/*
```
biconou: Because we are adding here a new table/DAO/domain object, we should start to write unit and integration tests also.

airsonic-main/src/main/resources/liquibase/6.5/add-multiple-genres-join.xml
```diff
@@ -0,0 +1,88 @@
...
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+    <!-- Drop the previous genre table and create a new one w/ an id primary index -->
+    <changeSet id="recreate-genre-table" author="devonbryant">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <columnExists tableName="genre" columnName="id"/>
+            </not>
+        </preConditions>
+        <dropTable tableName="genre"/>
```
biconou: There is a drop table here. What about existing datas ?

kaysond: I think you have to drop the genre table otherwise you'd end up with this problem which is also a problem if you re-add a file library: https://imgur.com/V6LmjhE

adylise: Very excited to see progress being made here! :)

nicolahinssen:
Very nice work! 
Will it also be possible to split the multiple genres using a Null character? As the official ID3v2.4 standards state:

4.2. Text information frames The text information frames are often the most important frames, containing information like artist, album and more. There may only be one text information frame of its kind in an tag. All text information frames supports multiple strings, stored as a null separated list, where null is reperesented by the termination code for the charater encoding.

airsonic-main/src/main/resources/liquibase/6.5/add-multiple-genres-join.xml
adylise: > 
> 
> @adylise What do you mean ?

Mult-genre support is something I've been really looking forward to eventually having, and I'm very happy to see some people working on it and testing stuff out for it. :)

airsonic-main/src/main/java/org/airsonic/player/dao/GenreDao.java
```diff
@@ -0,0 +1,102 @@
...
+        }
+    }
+
+    public void addGenre(Album album, Genre genre) {
+        if (addGenre("album", "album_count", album.getId(), genre.getId())) {
+            genre.incrementAlbumCount();
+        }
+    }
+
+    public void clearGenres() {
```
biconou: This method is only used in integration tests. So it should be moved as a utility in test sources.

airsonic-main/src/test/java/org/airsonic/player/service/MediaScannerServiceTestCase.java
```diff
@@ -111,6 +109,7 @@ public Statement apply(Statement base, Description description) {
     public void testScanLibrary() {
         MusicFolderTestData.getTestMusicFolders().forEach(musicFolderDao::createMusicFolder);
         settingsService.clearMusicFolderCache();
+        genreDao.clearGenres();
```
biconou: Add assertions here to verify that there are no more genres in database.

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -271,20 +281,39 @@ private void scanFile(MediaFile file, MusicFolder musicFolder, Date lastScanned,
...
     }
 
-    private void updateGenres(MediaFile file, Genres genres) {
-        String genre = file.getGenre();
-        if (genre == null) {
-            return;
-        }
-        if (file.isAlbum()) {
-            genres.incrementAlbumCount(genre);
+    private List<Genre> getGenres(MediaFile file, Map<String, Genre> cache) {
```
biconou: This method extracts the collection of genres from a media file parsing the genre ID3 tag. Ok but why this isn't done directly while the MediaFile object is instanciated ? This would simplify the library scan process. 

@devonbryant

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -228,38 +232,44 @@ private void doScanLibrary() {
...
-        searchService.index(file);
-
         // Update the root folder if it has changed.
         if (!musicFolder.getPath().getPath().equals(file.getFolder())) {
             file.setFolder(musicFolder.getPath().getPath());
             mediaFileDao.createOrUpdateMediaFile(file);
         }
 
         if (file.isDirectory()) {
+            searchService.index(file, Collections.emptyList());
```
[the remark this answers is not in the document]
biconou: I see, this is because it has been removed previously. But you could simplify that by not having to treat differently the isDirectory case. See my other comment bellow.

biconou:
@devonbryant Can we discuss on some refactoring on the code you proposed ? I opened several discussions on code changes.

kaysond:
@devonbryant @biconou Any updates on this?

tesshucom:
Database modifications are clearly obvious and good for anyone familiar with the database.
However, the range of influence seems to be very large in the deep part.

By clarifying where the "genre" input path is, it may be possible to solve it in a simpler, less risky way.
(Or unless it is clear, the risk range is unknown )

If there is a multi-genre request only on the UI and the "Genre" input route is via SearchService, it is possible to consider an approach to lucene index instead of changing to DB.
Another approach is to start with limited multi-genre support for SearchService only.
In this cases, the risk of change is much less and performance is likely to be a much better way.


(However, for future "genre" specifications, it is necessary to consider future prospects and design considerations, etc.)

eharris:
Since @devonbryant hasn't responded in almost a year, it seems like he has abandoned this PR, so I'm planning to pick this up.  I'll be submitting a new WIP PR soon (based on this one) with the aim to get this feature cleaned up and finished.

I am not intending to make the separator a configurable option, because doing so assumes that only one style of multiple genre tags will be present in an entire library.  What I am planning to do is try to reasonably accommodate all of: id3v2.3 parens `(17)(92)Folk-Rock`, id3v2.4 null-separated, and the semi-accepted "standard" of semicolon separators.

peioe:
@eharris What about multiple tags in vorbis comments ? MP3Tag (for example) allows to add multiple tags to FLAC files and this seems like the only "real" standard. See on https://en.wikipedia.org/wiki/Vorbis_comment it says:

> Any tag name is allowed, and there is no format that the data values must be in. This is in contrast to the ID3 format used for MP3s, which is highly structured. Field names are also permitted to be used more than once. It is encouraged to use this feature to support multiple values, for example two ARTIST=... fields to list both artists of a single composition. 

It would be great to support this :)

eharris:
@peioe Please put comments and discussion about the feature request in the ticket for the feature (#105) not in this PR.  My comment here was intended only to let people who were interested in this PR know of the intent to work on it.

tesshucom:
I think we should focus on things other than storing the input in the database.

 - What kind of usage do you expect when you support multiple genres?
 - Is the CRUD update / delete design robust?

If what is ultimately required of Genere is a functional design that assumes a search, it is more consistent if done with Lucene instead of a database.

 - It is not desirable to scatter word division and filter design specifications in the system.
 - Such a design has already been codified in Lucene's FW. Much safer and faster than proprietary designs.

It is possible to work with multiple values ​​in Lucene, if you rethink that the values ​​currently stored in the database are already composite keys.

In my clone server, it has been practiced, and I personally think that the design change of DB is unnecessary.
The database serves as storage for storing file information at the time of scanning.
Designs that divide and store in complex tables via specific business logic have many disadvantages.

eharris:
@tesshucom I'm open to doing this through Lucene if it can do it adequately.  But please put your comments in the parent ticket since this PR will be closed soon.

almet:
Hi, I'm just doing some triaging here.

@eharris, if I understand the history here, you were the last one working on this. Were you blocked somehow? Is there any way we can help you here? Or maybe you don't want / have the time to work on this anymore ? In which case, that would be super helpful to know where you left things and what's left to do !

PR #767, opened 2018-07-30
Use HTTPS and java.net.URI for scrobbling

### LINKED ISSUE #588: AudioScrobblerService should escape usernames in URLs
See https://github.com/Libresonic/libresonic/issues/483, where a trailing space in the username generated an invalid URL.

tari:
HTTPS will help prevent eavesdropping on the auth token, and using URI
will ensure unusual characters (like spaces, accidental or otherwise)
are escaped correctly.

Fixes #588


tari:
Test failure is an outdated dependency triggering the OWASP dependency checker- I think it's just a build dependency that's triggering it.

PR #807, opened 2018-09-30
archetype code for rest api integration tests

biconou:
Hi all, 

Here is an example code that should guide to write more integration tests against the airsonic rest api.
I'm working on that because I want to continue working on the jukebox feature. But, in order to do some safe refactoring on the api implementation I first decided to construct a solid test case bunch. 

**What you should notice about the test case I propose :** 
- database is embeded (due to test airsonic.properties)
- each test class uses a dedicated temporary dir as airsonic.home (due to `@ClassRule` static member)
- Spring configuration can be overriden using the `@TestConfiguration` inner class (not used in this first example)

@muff1nman

### DIFF
--- a/airsonic-main/pom.xml
+++ b/airsonic-main/pom.xml
@@ -87,6 +87,16 @@
             <artifactId>spring-boot-starter-security</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.security</groupId>
+            <artifactId>spring-security-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.springframework.security</groupId>
             <artifactId>spring-security-ldap</artifactId>
         </dependency>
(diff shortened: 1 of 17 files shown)


muff1nman:
@biconou This is definitely a good idea. In fact I was working on getting something similar while trying to test the streaming issues. See #808. Can you take a look at that to see if that is usable for your use case? Sorry for not posting it sooner.

The reason why I prefer it over doing spring test directly is:

1. It provides a portable way to test transcoding with ffmpeg
2. It is more parallelizeable than spring test. I have built quite a few code bases with spring test and it always seems to hit a point where it starts taking forever to get all of the tests run.
3. Cucumber is pretty powerful and encourages code reuse compared to spring test where its hard to utilize things like junit parameterized tests.

With that said, if you feel that you'd like to continue using spring test, I'm totally down for bringing this in. I'd much prefer having duplicate methods of running integration tests than having less integration tests because folks don't know how to use a particular bdd framework (i.e. cucumber).

airsonic-main/pom.xml
```diff
@@ -86,6 +86,15 @@
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-security</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
```
muff1nman: Can you mark this as test as well?

biconou: Done

muff1nman:
Looks like the jukebox tests are failing?

biconou:
@muff1nman Fixed

PR #834, opened 2018-11-25
Add extended favicons

heyarne:
I extended the favicons to span a couple extra devices. Also I hand-pixeled a 16x16 pixel version using @jooola's icon as the template. This avoids the Moiré-effect of the current 16x16 pixel version. I also tried hand-pixeling a 32x32-version but tbh I wasn't masochistic enough to pull it off on a Thinkpad touchpad / trackpoint. ;) I hope this is appreciated!


muff1nman:
Nice! @jooola this look good to you?

jooola:
Yup, sure. I used Real favicon myself few times, it's a nice tool.

heyarne:
I updated the head.jsp now, too. Also I added a pixelated 32x32 icon. I'm not quite sure how the files in `/webapp` get served, but I put one `favicon.ico` and one the `browserconfig.xml` into the root. Do they get served? Both are requested automatically by some browsers by convention. Anything I could configure is in `/icons`.

PR #846, opened 2018-12-09
Update plugins

randomnicode:
Update some maven plugins to pull them out of antiquity


muff1nman:
Looks like a CVE failure going on - maybe related to the bump in that module version?

PR #855, opened 2019-01-03
Fix #778: Defragment embedded HSQLDB database more frequently

### LINKED ISSUE #778: db/airsonic.data is very large
## Problem description

I'm running out of room on my `/` partition, and traced high disc usage to `/var/airsonic/db/airsonic.data`, which is 4.9 G. Is that normal? I'm wondering why this file is so large.

### Steps to reproduce

1. Use Airsonic
2. Check disc usage

## System information

 * **Airsonic version**: 10.1.1-RELEASE – December 16, 2017 (I haven't bothered upgrading to the latest version because I don't use LDAP.
 * **Operating system**: Linux, Raspbian (Debian)
 * **Proxy server**: None
 * **Client**: N/A.

## Additional notes

I have 11,119 music files in my library.

fxthomas:
Hey folks, happy new year!

I finally got around to fixing #778 for those of us using the embedded database. A few things to note before reviewing the code :

* I set the checkpoint size limit to 64MB (from a default of 200MB). This means that the DB will do a checkpoint each time the log ("pending changes") grows above that limit. This seems sane enough for Airsonic, and should not impact performance given the amount of data larger instances handle.
* The defrag limit is set to 32MB of wasted space, checked on each checkpoint. If the wasted space is below 32MB, nothing happens (just a regular checkpoint). This _will_ impact performance while the defrag is happening, because this is an expensive operation (usually below 10s for me, with ~15k songs).
* This is not compatible with HSQLDB 2 (it uses a different syntax), but I'm not sure we need to support it.

This sounds reasonable to me, and works on my instance. Comments are of course welcome!

### DIFF
--- a/airsonic-main/src/main/resources/liquibase/10.2/changelog.xml
+++ b/airsonic-main/src/main/resources/liquibase/10.2/changelog.xml
@@ -1,0 +1,6 @@
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
+    <include file="setup-hsqldb-checkpoint-defrag.xml" relativeToChangelogFile="true"/>
+</databaseChangeLog>
--- a/airsonic-main/src/main/resources/liquibase/10.2/setup-hsqldb-checkpoint-defrag.xml
+++ b/airsonic-main/src/main/resources/liquibase/10.2/setup-hsqldb-checkpoint-defrag.xml
@@ -9,6 +9,10 @@
         </preConditions>
 
         <sql>
+          CHECKPOINT DEFRAG;
+          <comment>Defragment the database before enabling auto defrag, so that the biggest part of the work is done during migration.</comment>
+        </sql>
+        <sql>
           SET LOGSIZE 64;
           <comment>Automatically run a CHECKPOINT when the log is above 64MB.</comment>
         </sql>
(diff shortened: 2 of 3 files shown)


muff1nman:
Great job finding those settings. My only concern is for those where their database hasnt been defragged in a long time (ever?). I imagine this first defrag will take longer than 10s? Is this something we just need to communicate clearly on the release notes?

fxthomas:
Yes, it will take a little longer. When I did a big defrag with my old multi-GB `db` folder I remember it taking about a minute (a large part of it being loading the DB in memory, I suspect).

How does issuing a first `CHECKPOINT DEFRAG` during the migration sound? Airsonic will clearly wait during an explicit migration operation instead of randomly freezing during operation ; this is also mentioned in the logs if we need to debug, and we can mention it in the release notes just to be cautious.

muff1nman:
Okay so not too bad, but I think your idea about running a defrag for the first time that migration runs would be smart. Can you add that?

fxthomas:
Yup, done. Just for reference, this lasts about 2s on my collection (already defragmented recently):

```
2019-01-09 22:44:15.223  INFO --- liquibase  : Successfully acquired change log lock
2019-01-09 22:44:15.251  INFO --- liquibase  : Reading from DATABASECHANGELOG
2019-01-09 22:44:17.738  INFO --- liquibase  : classpath:liquibase/db-changelog.xml: classpath:liquibase/10.2/setup-hsqldb-checkpoint-defrag.xml::setup-hsqldb-checkpoint-defrag::fxthomas: Custom SQL executed
2019-01-09 22:44:17.740  INFO --- liquibase  : classpath:liquibase/db-changelog.xml: classpath:liquibase/10.2/setup-hsqldb-checkpoint-defrag.xml::setup-hsqldb-checkpoint-defrag::fxthomas: Custom SQL executed
2019-01-09 22:44:17.741  INFO --- liquibase  : classpath:liquibase/db-changelog.xml: classpath:liquibase/10.2/setup-hsqldb-checkpoint-defrag.xml::setup-hsqldb-checkpoint-defrag::fxthomas: Custom SQL executed
2019-01-09 22:44:17.742  INFO --- liquibase  : classpath:liquibase/db-changelog.xml: classpath:liquibase/10.2/setup-hsqldb-checkpoint-defrag.xml::setup-hsqldb-checkpoint-defrag::fxthomas: ChangeSet classpath:liquibase/10.2/setup-hsqldb-checkpoint-defrag.xml::setup-hsqldb-checkpoint-defrag::fxthomas ran successfully in 2306ms
2019-01-09 22:44:17.754  INFO --- liquibase  : Successfully released change log lock
```

Issue #495, opened 2017-07-17
Review libresonic #476 (CUE sheet support for single-track albums)


muff1nman:
PR https://github.com/Libresonic/libresonic/pull/476 looks promising and would resolve #303. Only thing that is really bothersome is the jitpack dependency. Does anyone have CUE type media they could test with?

jooola:
Why not simply ask the PR owner to post it here also ? (on airsonic) 🤣

korden32:
I would be happy to test & use CUE support, but as I already commented at the original PR, it doesn't work for me - CUE parsing itsef is working and I see tracks in the database, but I can't play such tracks - it throws an error

```
java.lang.SecurityException: Access denied to file Intro
        at org.airsonic.player.service.MediaFileService.getMediaFile(MediaFileService.java:141) ~[classes!/:10.0.0-RELEASE]
        at org.airsonic.player.controller.CoverArtController.createMediaFileCoverArtRequest(CoverArtController.java:180) ~[classes!/:10.0.0-RELEASE]
        at org.airsonic.player.controller.CoverArtController.createCoverArtRequest(CoverArtController.java:150) ~[classes!/:10.0.0-RELEASE]
...
```
It probably requires some fixes in getFile or getSingleFileAlbumSongWholePath 

I can provide example flac+cue for testing if needed

tesshucom:
At Airsonic, #856 looks like successor work. I'll close it. If there is a problem, please reopen it.

Issue #303, opened 2017-03-18
cue sheet support for one flac per disc files


issuemover631:
<a href="https://github.com/svolki"><img src="https://avatars2.githubusercontent.com/u/26509675?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [svolki](https://github.com/svolki)**
_Saturday Mar 18, 2017 at 18:56 GMT_
_Originally opened as https://github.com/Libresonic/libresonic/issues/301_

----

It would be great, if libresonic could handle cue sheets embedded into flac files, see https://en.wikipedia.org/wiki/Cue_sheet_(computing).

Using this a whole CD image can be stored in one flac file and be used as playback source as well as backup of the original disc.

jooola:
A PR is present on the libresonic repo, #495 is following this PR.

Closing this ! ( can be reopened if the PR isn't merged in Airsonic )

muff1nman:
Let's not get too hasty closing issues. The policy should be the issue gets closed after the PR is merged.

tesshucom:
At Airsonic, #856 looks like successor work. Too many multiple posts. I'll close it. If there is a problem, please reopen it.

PR #856, opened 2019-01-04
CUE sheet support for music and audiobook types

Yetangitu:
This PR is based on https://github.com/Libresonic/libresonic/pull/476 and relates to #495  in that it implements .cue support for _music_ and _audiobook_ media types. It matches _.cue_ files to identically-named sound files (e.g `sound_file.cue` will match `sound_file.mp3` or `sound_file.flac`). Tracks defined in _.cue_ files are added to the database and can be individually played, this is achieved by (ab)using the `downsample` facility to serve the relevant part of the audio source file. No actual downsampling is enforced, the resulting track will have the same bitrate as the source file unless that exceeds the specified maximum bitrate.

Tested with a collection of ~500 mp3-files containing about 11.000 tracks through both the web interface as well as the Subsonic API.

This is a rough first draft of the implementation, let me know if this is of interest. I made it for my own instance which hosts a number of indexed public radio recordings.

Re. Libresonic, this PR could also be of interest to your project as it fixes some problems with the PR I based it on and adds new functionality.

To actually get this PR to work you'll want to change the `downsample` command (in `Settings -> Transcoding`) to something resembling the following:

```
ffmpeg -ss %ss -to %to -i %s -map 0:0 -b:a %bk -v 0 -f mp3 -
```

If this PR is merged this should be added as a default configuration.

### DIFF
--- a/airsonic-main/pom.xml
+++ b/airsonic-main/pom.xml
@@ -538,10 +538,6 @@
             <version>64.2</version>
         </dependency>
         <dependency>
-            <groupId>com.jayway.jsonpath</groupId>
-            <artifactId>json-path</artifactId>
-        </dependency>
-        <dependency>
             <groupId>com.sun.mail</groupId>
             <artifactId>jakarta.mail</artifactId>
             <scope>runtime</scope>


Yetangitu:
BTW, the way this is currently implemented the base files (the sound files which are indexed through _.cue_ sheets) tend to get hidden in the avalanche of indexed files, making it harder to play them 'all at once'. While it is possible to play these files track by track this is less efficient and - at least in the web version - not gapless.

Another way to realise this feature would be to add the _.cue_ file itself as album - i.e. separate from the base file it refers to - which can be opened to reveal the containing tracks. This would make it possible to easily refer to both the base file (to play it as a whole) or to individual tracks.

Current scheme:

```
directory - track_1_from_cue_1
            track_2_from_cue_1
            track_3_from_cue_1
            ...
            track_n_from_cue_1
            track_1_from_cue_2
            track_2_from_cue_2
            track_3_from_cue_2
            ...
            track_n_from_cue_2
            base_file_1.mp3 
            base_file_2.mp3
```

Alternative scheme:

```
directory - virtual_album_1 - track_1_from_cue_1
          .                   track_2_from_cue_1
          .                   track_3_from_cue_1
          .                   ...      
          .                   track_n_from_cue_1
          - virtual_album_2 - track_1_from_cue_2   
          .                   track_2_from_cue_2
          .                   track_3_from_cue_2
          .                   ...
          .                   track_n_from_cue_2
          - base_file_1.mp3
          - base_file_2.mp3 
```

This could also be turned into a configurable preference ('virtual albums' or something like that).

RuralHunter:
+1

RuralHunter:
one question, should cue be added into music file types in setting?

Yetangitu:
No, you don't need to change anything in the current implementation, it is hard-coded to look for files with a '.cue' extension. In the current implementation these files need to be named identically to to sound files which they index, I experimented with using the FILE attribute from the _.cue_ file to find the related sound file. This code is not yet in this PR so for now name those files well.

RuralHunter:
ok, I will try and thanks a lot for the PR. this will be a great feature for me as I have many cue+ape music files if it works as expected.

Yetangitu:
I did not test with _ape_ but the code should be format-agnostic. You might run into some of the effects of #685 (I did) in which case you could try to disable 'Enable seek on transcoded files' (in the _Transcoding_ section).

RuralHunter:
I installed the PR. But it seems the scheduled rescan did not pick up the cue files in my existing library.
```
2019-01-07 03:00:00.262  INFO --- o.a.p.s.MediaScannerService              : Starting to scan media library.
2019-01-07 03:00:04.027  INFO --- o.a.p.s.MediaScannerService              : Scanned media library with 250 entries.
2019-01-07 03:00:06.843  INFO --- o.a.p.s.MediaScannerService              : Scanned media library with 500 entries.
2019-01-07 03:00:09.556  INFO --- o.a.p.s.MediaScannerService              : Scanned media library with 750 entries.
2019-01-07 03:00:10.623  INFO --- o.a.p.s.MediaScannerService              : Scanned media library with 855 entries.
2019-01-07 03:00:10.625  INFO --- o.a.p.s.MediaScannerService              : Marking non-present files.
2019-01-07 03:00:10.632  INFO --- o.a.p.s.MediaScannerService              : Marking non-present artists.
2019-01-07 03:00:10.636  INFO --- o.a.p.s.MediaScannerService              : Marking non-present albums.
2019-01-07 03:00:10.676  INFO --- o.a.p.s.MediaScannerService              : Completed media library scan.
2019-01-07 03:00:10.932  INFO --- o.a.p.service.PlaylistService            : Starting playlist import.
2019-01-07 03:00:10.933  INFO --- o.a.p.service.PlaylistService            : Completed playlist import.
```
How to trigger it to pick up the once ignored cue files?

RuralHunter:
I tried on a new folder and there is problem on cue+ape files:
```
org.jaudiotagger.audio.exceptions.CannotReadException: No Reader associated with this extension:ape
	at org.jaudiotagger.audio.AudioFileIO.readFile(AudioFileIO.java:353) ~[jaudiotagger-2.2.5.jar!/:na]
	at org.jaudiotagger.audio.AudioFileIO.read(AudioFileIO.java:198) ~[jaudiotagger-2.2.5.jar!/:na]
	at org.airsonic.player.service.metadata.JaudiotaggerParser.getRawMetaData(JaudiotaggerParser.java:86) ~[classes!/:10.2.0-SNAPSHOT]
	at org.airsonic.player.service.metadata.MetaDataParser.getMetaData(MetaDataParser.java:47) [classes!/:10.2.0-SNAPSHOT]
	at org.airsonic.player.service.MediaFileService.createSingleFileAlbumChildren(MediaFileService.java:445) [classes!/:10.2.0-SNAPSHOT]
	at org.airsonic.player.service.MediaFileService.updateChildren(MediaFileService.java:406) [classes!/:10.2.0-SNAPSHOT]
	at org.airsonic.player.service.MediaFileService.getChildrenOf(MediaFileService.java:194) [classes!/:10.2.0-SNAPSHOT]
	at org.airsonic.player.service.MediaScannerService.scanFile(MediaScannerService.java:248) [classes!/:10.2.0-SNAPSHOT]
	at org.airsonic.player.service.MediaScannerService.scanFile(MediaScannerService.java:252) [classes!/:10.2.0-SNAPSHOT]
	at org.airsonic.player.service.MediaScannerService.scanFile(MediaScannerService.java:252) [classes!/:10.2.0-SNAPSHOT]
	at org.airsonic.player.service.MediaScannerService.doScanLibrary(MediaScannerService.java:188) [classes!/:10.2.0-SNAPSHOT]
	at org.airsonic.player.service.MediaScannerService.access$000(MediaScannerService.java:45) [classes!/:10.2.0-SNAPSHOT]
	at org.airsonic.player.service.MediaScannerService$2.run(MediaScannerService.java:157) [classes!/:10.2.0-SNAPSHOT]
```

RuralHunter:
This is strange. Since I can add APE files into library without this patch. Does the old process not get the meta data of APE files?

RuralHunter:
One more thing: in my case, the -ss and -to parameters should be added in transcode command too and they should be after the -i parameter, like this:
`ffmpeg -i %s -ss %ss -to %to -map 0:0 -b:a %bk -v 0 -f mp3 -`
Otherwise I got the error with ffmpeg:
`ffmpeg ape Option to (record or transcode stop time) cannot be applied to input url xxxxx -- you are trying to apply an input option to an output file or vice versa. Move this option before the file it belongs to.`

RuralHunter:
for the jaudiotagger parsing error, I created an issue #865

Yetangitu:
Madsonic has done some work on metadata collection for less common formats which could be adapted to _ape_.

https://gitlab.com/madsonic/madsonic-main/blob/master/src/main/java/org/madsonic/service/MediaFileService.java#L941

```
				// ### fix duration/bitrate on special formats
				if (Arrays.asList("aac", "aiff", "dff", "dsf", "opus", "ogg", "flac", "mod", "mid", "it", "s3m", "xm").contains(StringUtils.lowerCase(mediaFile.getFormat()))) {
					FFmpegParser fparser = new FFmpegParser();
					fparser.setTranscodingService(transcodingService);
					if (fparser != null) {
						MetaData fmetaData = fparser.getRawMetaData(file);
						mediaFile.setDurationSeconds(fmetaData.getDurationSeconds());
						mediaFile.setBitRate(fmetaData.getBitRate());
						if (!"flac".contains(StringUtils.lowerCase(mediaFile.getFormat()))) {
							if (fmetaData.getTitle() != null) mediaFile.setTitle(fmetaData.getTitle());
							if (fmetaData.getArtist() != null) mediaFile.setArtist(fmetaData.getArtist());
							if (fmetaData.getAlbumName() != null) mediaFile.setAlbumName(fmetaData.getAlbumName());
							if (fmetaData.getYear() != null) mediaFile.setYear(fmetaData.getYear());
						}
					}
				}			
```

Yetangitu:
There seem to be some edge cases with picking up changed/new files which I'll tackle later this day when I have some time.

Yetangitu:
> I installed the PR. But it seems the scheduled rescan did not pick up the cue files in my existing library.
> 
> ```
> 2019-01-07 03:00:00.262  INFO --- o.a.p.s.MediaScannerService              : Starting to scan media library.
> 2019-01-07 03:00:04.027  INFO --- o.a.p.s.MediaScannerService              : Scanned media library with 250 entries.
> 2019-01-07 03:00:06.843  INFO --- o.a.p.s.MediaScannerService              : Scanned media library with 500 entries.
> 2019-01-07 03:00:09.556  INFO --- o.a.p.s.MediaScannerService              : Scanned media library with 750 entries.
> 2019-01-07 03:00:10.623  INFO --- o.a.p.s.MediaScannerService              : Scanned media library with 855 entries.
> 2019-01-07 03:00:10.625  INFO --- o.a.p.s.MediaScannerService              : Marking non-present files.
> 2019-01-07 03:00:10.632  INFO --- o.a.p.s.MediaScannerService              : Marking non-present artists.
> 2019-01-07 03:00:10.636  INFO --- o.a.p.s.MediaScannerService              : Marking non-present albums.
> 2019-01-07 03:00:10.676  INFO --- o.a.p.s.MediaScannerService              : Completed media library scan.
> 2019-01-07 03:00:10.932  INFO --- o.a.p.service.PlaylistService            : Starting playlist import.
> 2019-01-07 03:00:10.933  INFO --- o.a.p.service.PlaylistService            : Completed playlist import.
> ```
> How to trigger it to pick up the once ignored cue files?

Use `touch` to change the time stamp on the directories containing those files.

Yetangitu:
By the way, a question of policy: where in 'airsonic-main/src/main/resources/liquibase/...` should changes w.r.t. the default transcoding and downsampling commands be added? These commands are defined in various iterations in `../legacy/schemaXX.xml`  and `../6.3/mediaelement.xml`, they're also in `airsonic-main/src/main/java/org/airsonic/player/service/SettingsService.java` in the form of `private static final String DEFAULT_<SOMETHING>_COMMAND`. I'd say both need to be changed, in the latter case it is clear where but in the former this is more a matter of policy - adding a new file to the versioned directories in `../liquibase/` seems the cleanest option.

muff1nman:
https://github.com/airsonic/airsonic/tree/master/airsonic-main/src/main/resources/liquibase/10.2 is the folder for the current batch of database schema changes.

RuralHunter:
FYI, to make the transcode work better with APE, we should use this kind of transcode command:
`ffmpeg -ss %ss -i %s -to %to -copyts -map 0:0 ...`

Yetangitu:
> FYI, to make the transcode work better with APE, we should use this kind of transcode command:
> `ffmpeg -ss %ss -i %s -to %to -copyts -map 0:0 ...`

I assume the essential part of this command to be the `-map 0:0` section? That is already part of the _downsample_ command which is used by this patch. The `-copyts` section would lead to strange timestamps with tracks starting with non-zero timestamps so I don't think that a good idea. The `-ss %ss -to %to` is a prerequisite for cutting the base file in track-sized sections so that will have to be part of all *audio* transcode commands with the transcoder taking care of filling in the correct values to play tracks from base files. For video files the transcoder should not mess with any start/end times as those are probably the result of the user wishing to play a different section of the video file.  The current incarnation of the PR does not discriminate between audio or video, I'll code up something to handle this case this weekend.

RuralHunter:
No, it has nothing to do with "-map". The point is "-to" before the APE input parameter is not support at least for my ffmpeg. So I have to put it after the "-i" input parameter. We can use "-to" with "-copyts" with track end time, or use "-t" with track duration.

Yetangitu:
OK, I'll go find me an APE track - none of those around here - and see what works with the various incarnations of `ffmpeg` (3.2.5, 4.0.0 and 4.0.2) currently in use on some of my systems. The ffmpeg manual has the following to say on the placement of `-ss` and `-to`:

> -ss position (input/output)
           When used as an input option (before "-i"), seeks in this input file to position. Note that in most formats it is not possible to seek exactly, so ffmpeg will seek to the closest seek point before position.  When transcoding and -accurate_seek is enabled (the default), this extra segment between the seek point and position will be decoded and discarded. When doing stream copy or when -noaccurate_seek is used, it will be preserved. When used as an output option (before an output url), decodes but discards input until the timestamps reach position.
position must be a time duration specification, see the Time duration section in the ffmpeg-utils(1) manual.

RuralHunter:
You can get an APE example here: https://example-files.online-convert.com/audio/ape/example.ape

RuralHunter:
Yes, it's not practical to put "-ss" after "-i" as this will decode all the content before the start position which is meaningless and too slow. The main problem is the "-to" parameter. You can also find more info there: https://trac.ffmpeg.org/wiki/Seeking

CompulsiveConfig:
Received the following error for a TTA single file album [with accompanying CUE](https://files.catbox.moe/44mmee.cue): 
```
java.lang.NullPointerException
	at org.airsonic.player.service.MediaFileService.createSingleFileAlbumChildren(MediaFileService.java:450)
	at org.airsonic.player.service.MediaFileService.updateChildren(MediaFileService.java:406)
	at org.airsonic.player.service.MediaFileService.getChildrenOf(MediaFileService.java:194)
	at org.airsonic.player.service.MediaFileService.getChildrenOf(MediaFileService.java:175)
	at org.airsonic.player.controller.MainController.handleRequestInternal(MainController.java:91)
```

Files without CUE run as expected, including TTA.
This is on FreeBSD 11.2-STABLE with java 1.8.0_202

Yetangitu:
Does this happen with all .TTA files (a format which I do not have access to and as such have not tested) or only with this one?

CompulsiveConfig:
> Does this happen with all .TTA files (a format which I do not have access to and as such have not tested) or only with this one?

This happened with all TTA files. 

After some investigation, I was able to diagnose the issue and [start working on a fix](https://github.com/CompulsiveConfig/airsonic/tree/cue-support-tta). This is my first practice with Java for some years, so please improve where you see fit. 

The core of the issue is that when the TTA file is sorted through `MetaDataParserFactory`, it is only applicable to the base `MetaDataParser` class. This class does not provide the Mediafile with sufficient metadata for the `MediaFileService.createSingleFileAlbumChildren()`, and the null value in `metaData.getDurationSeconds` causes the exception above. 

In a similar manner to the APE issue, `JaudiotaggerParser` does not contain support for TTA files. Forcing tta acceptance in `JaudiotaggerParser.isApplicable()` only leads to a similarly insufficient Mediafile. 

Therefore, I instead enabled `FFmpegParser` applicability by setting `settingsService.getMusicFileTypesAsArray()` as a public method, _and for the purpose of a temporary test_, replaced `settingsService.getVideoFileTypesAsArray()` with `settingsService.getMusicFileTypesAsArray()` . These arrays will need to reconsidered in the future. 

However, this fixed most of the issues with TTA files. I can now play them, and my entire collection of TTAs + Cue is listed as individual songs, with most timings as expected. The only two issues I still experience are: 

1. In the web client, weird increases in overall play time when transcoding to MP3, of about 2-3 minutes. The audio stops at the predicted stop time (correctly displayed in the playlist), but the seek bar displays several more minutes of unusable audio. This is also interrupts the transition from one song to the next. This symptom was not present in Ultrasonic on mobile. 

2. Lack of support for proper FLAC transcoding, which has normally been functional as an output stream to client players. Perhaps this may have something to do with the downsampling function.

If you are still having issues with APE, I would recommend a similar treatment for it. ffmpeg generally takes whatever you throw at it quite well, while Jaudiotagger seems to have a limited format support. 

Thanks for all your hard work!

airsonic-main/src/main/java/org/airsonic/player/service/MediaFileService.java
```diff
@@ -416,6 +429,84 @@ private void updateChildren(MediaFile parent) {
...
+                if (soundFile != null) {
+                    MetaDataParser parser = metaDataParserFactory.getParser(soundFile);
+                    MetaData metaData = null;
+                    if (parser != null) {
+                        metaData = parser.getMetaData(soundFile);
+                    }
+                    long wholeFileSize = FileUtil.length(soundFile);
+                    int wholeFileLength = 0; //todo: find sound length without metadata
+                    if (metaData != null) {
+                        wholeFileLength = metaData.getDurationSeconds();
```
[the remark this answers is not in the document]
jvoisin: @Yetangitu care to add a check here?

pom.xml
```diff
@@ -30,6 +30,10 @@
             <id>jaudiotagger-repository</id>
             <url>https://dl.bintray.com/ijabz/maven</url>
         </repository>
+        <repository>
+            <id>jitpack.io</id>
+            <url>https://jitpack.io</url>
```
jvoisin: Can't the cue lib be found in the official maven repo?

airsonic-main/src/main/java/org/airsonic/player/domain/MediaFile.java
```diff
@@ -471,12 +484,70 @@ public Integer apply(MediaFile from) {
         };
     }
 
+    public void setPathForSingleFileMedia(String mediaFilePath, int songStart, int songEnd) {
```
jvoisin: This is a bit gross, wouldn't be possible to have this stored in a field somewhere instead?

pom.xml
Yetangitu: It couldn't when I made the changes a long time ago, maybe it can now, will check. For now I'm just bringing the changes over to the latest version of airsonic, will look at these issues later.

airsonic-main/src/main/java/org/airsonic/player/domain/MediaFile.java
Yetangitu: Maybe, will look later, no time now...

RuralHunter:
for cuelib, I think you can use:
```
<dependency>
            <groupId>org.digitalmediaserver</groupId>
            <artifactId>cuelib-core</artifactId>
            <version>2.0.0</version>
        </dependency>
```
It's in the official maven repo

Yetangitu:
> > Does this happen with all .TTA files (a format which I do not have access to and as such have not tested) or only with this one?
> 
> This happened with all TTA files.

Can you provide me with a TTA + cue file for testing purposes? I don't care what is in the file, as long as it is indexed through a cue file.

Yetangitu:
> for cuelib, I think you can use:
> 
> ```
> <dependency>
>             <groupId>org.digitalmediaserver</groupId>
>             <artifactId>cuelib-core</artifactId>
>             <version>2.0.0</version>
> </dependency>
> ```
> 
> It's in the official maven repo

Thanks, I happened upon that one s bit earlier and did some testing, it seems to work.

Just like my question to @CompulsiveConfig wrt TTA files it might help if you can provideo me with an APE + cue file for testing purposes. I'd like to get this thing finished and working for as many file types as possible - and preferably merged so I can ditch my own fork.

RuralHunter:
oh, which problem do you mean ? I think I don't have more standing issue with this PR.

Yetangitu:
> oh, which problem do you mean ? I think I don't have more standing issue with this PR.

I don't have any problems but others - @CompulsiveConfig being one of them - seem to have some. Those problems seem to relate to the use of different file formats (TTA and APE, maybe FLAC as well?) so I'd like to iron those out before this gets merged. I'll push the changes I made so far.

pom.xml
Yetangitu: Now using `cuelib-core` from _org.digitalmediaserver_

airsonic-main/src/main/java/org/airsonic/player/domain/MediaFile.java
Yetangitu: While this might be 'gross' it does have the advantage of being the least invasive way to add this functionality. I can change it to something more 'pure' but that'd necessitate changes to a fair bit of infrastructure as well as a schema update.

Yetangitu:
This change makes FFmpegParser applicable for metadata extraction on all playable file types instead of just video files. According to https://github.com/airsonic/airsonic/pull/856#issuecomment-490087257 this enables extraction of metadata from TTA - and with that I assume also other 'exotic' - file types. Since I do not have much in the way of 'exotic' file types I'd suggest those of you who do to test this change.

Yetangitu:
First Travis run: failed integration test in JDK11, rest worked fine. Second run: failed integration test in JDK9, rest works fine.

Conclusion: Travis has problems. One more try by reverting the empty commit from above...

eharris:
Keep in mind that the separate "Downsampling" setting will be going away hopefully very soon.  See #1124, specifically the second commit.  Please also see #1138, and add comments for any concerns that might relate to this support.

Yetangitu:
> Keep in mind that the separate "Downsampling" setting will be going away hopefully very soon. See #1124, specifically the second commit. Please also see #1138, and add comments for any concerns that might relate to this support.

OK, that means I better look for a more robust way to index and serve partial tracks. Back to the drawing board but for now it works.

Yetangitu:
See https://github.com/airsonic/airsonic/issues/1138#issuecomment-529201631 for the path I intend to take wrt. the disappearance of the _downsampling_ setting. This will also take care of the [gross](https://github.com/airsonic/airsonic/pull/856#discussion_r319751901) way cue-indexed tracks are currently represented in the database by introducing a new `startPosition` field. No time now to finish the implementation though, first have to finish the ground for a new barn I'm building IRL.

eharris:
I have some concerns with a few of the implementation details of this (although overall I like the idea of adding support for indexed media):

* I'm a little worried that this might be adding support that isn't general enough (that is, it is too specific to `ffmpeg`).  Perhaps we should consider if we should make the "sectioning" be done by a separate command or "helper" script rather than within the transcoding commands?
* How might indexing into files work if the index is by byte offset rather than time?
* The current implementation breaks the convention we've had up until now of using only single character substitution keys in the transcoding commands.  If we decide we are willing to break this convention, we should probably first (in a separate PR) add support for some form of substitution quoting, e.g. `%{ss}`.  Otherwise the substitutions should be made single character, perhaps `%S` and `%T`?
* What happens when there isn't a seek position (or some other parameter substitution) defined for a file?
* Further extensions (or generalization) to the transcoding substitution support might also 
be warranted.  For instance, making parameters optional (conditionally included only if they contain a value). Perhaps something like adding %? for a conditional, so the transcoding command line segment might be: `%?S{-ss %S} %?T{-to %T}`

If we are going to add support for indexed (segmented) media files, it should be done in a general fashion.  As an example, this same "sectioning" capability would seem like it could be applied to playing chapters of a video file as well?

eharris:
I just (re-)realized that the transcoding commands already support `%o` and `%d` for starting offset and duration (of video files), so rather than create new `%ss` and `%to` substitutions, you should probably just adapt those so they are usable for any type of file (not just video).

This would obviate the concerns I expressed above with adding new substitutions.

Yetangitu:
I'm in the process of changing the way tracks are split without using the (doomed) _downsampling_ option, instead of which I'm introducing a (hopefully not doomed) _split_ command which takes care of this. As to the double-letter substitution keys, I've changed the way things work where it now uses `%d` (for _duration_) and `%o` (for _offset_) instead of `%to` and `%ss`. As to the question with regard to _byte_ instead of _time_ offsets this could be handled by introducing a few new options like `%bo` (for _byte offset_) or `%O` (less clear but not against the previously mentioned convention of only using single-character substitution keys (why?)).

The split command is configurable and can be set to anything that works. By default it uses `ffmpeg` but this can be changed if needed as long as the command can dump its output to `stdout`.

I have not tried video files yet but in principle things should work, the _split_ command does nothing more than split a given section out of a media file without re-encoding its contents. I'm currently using the following command:

```
private static final String DEFAULT_SPLIT_COMMAND = "ffmpeg -ss %o -t %d -i %s -vcodec copy -acodec copy -f %f -";
```

`%f`: input file format

This will split both audio as well as video files - it is format/container-agnostic as long as `ffmpeg` supports the format).

Yetangitu:
RFC: the database scheme currently defines `media_file.path` as being unique even though it does not serve as primary key. If indexed files are to become 1st-class entities this constraint has to be changed as a single `path` no longer identifies a track. The solution used thus far - which was described as being 'gross' (https://github.com/airsonic/airsonic/pull/856#discussion_r319751901) - was to create unique path names by adding the start and finish times to the actual path, thereby enabling the data to be recorded in the database. These start and finish times were also used to split the tracks out of the file for playback.

In the current version of the patch these start and finish times are no longer used, instead there is a new 'start_position' field in the database which defaults to `0`. Tracks are fully identified by the combination of `path` and `start_position`. In theory you could add `duration_seconds`to the mix to enable tracks with different end positions but that doesn't make much sense for this application.

Here's the question: either the `unique` constraint on `path`needs to be changed into a compound index on `path` and `start_position` and those parts of the code which use `path` to identify tracks have to be changed accordingly _or_ the 'gross' constructed path names need to stay in some form or other. Choose...

Yetangitu:
Another RFC: in the current incarnation of this patch (which is not yet in the PR) the `split` command is injected into the transcoding, moving any existing transcoding steps down the line. This means that the number of available transcoding steps for cue-index tracks are reduced from 3 to 2. This could be solved by adding an extra (invisible, not user-configurable) step but the question is whether there are 3-step transcoders being used _in the wild_. Anyone?

muff1nman:
Let me read/think over this in detail.

Yetangitu:
I will, but meanwhile I'll push the latest version to give some ideas...

eharris:
@Yetangitu I don't think that I like the idea of having a separate "cutter/split" command, for one main reason: Some formats don't really allow for "cutting" without re-encoding anyway.  For lossless formats like WAV and FLAC, I agree that it doesn't really matter.  But for mp3 or other lossy formats, I would think that ffmpeg will end up re-encoding the file anyway, so why not just have the "segmenting" portion be a standard part of the transcoding framework?  And I'd rather not to have yet another special case for lossless vs lossy.

Yetangitu:
@eharris ffmpeg does not re-encode lossy formats, it cuts them at frame boundaries. A single (non-transient) mp3 frame generally contains 1152 samples, at 44.1 kHz a single frame has a duration of ( (1152 / 44100) * 1000 =) ~26ms. There is a marked difference in speed and quality between this method of splitting and a re-encoding version.

Yetangitu:
Here's the salient bits of the airsonic log when splitting a file using this method:

```
2019-09-13 21:57:49.418  INFO --- o.a.p.io.TranscodeInputStream            : Starting transcoder: 
Starting transcoder: [/srv/airsonic_test/transcode/ffmpeg] [-ss] [3243] [-t] [250] [-i] [/export/audio/music/Top_2000/2018/17-1600_1800_top_2000_20181226.mp3] [-vcodec] [copy] [-acodec] [copy] [-f] [mp3] [-]
...
(/srv/airsonic_test/transcode/ffmpeg) [mp3 @ 0x564b748fe540] Estimating duration from bitrate, this may be inaccurate  
(/srv/airsonic_test/transcode/ffmpeg) Input #0, mp3, from '/export/audio/music/Top_2000/2018/17-1600_1800_top_2000_20181226.mp3':
(/srv/airsonic_test/transcode/ffmpeg)   Duration: 01:49:11.04, start: 0.000000, bitrate: 192 kb/s
(/srv/airsonic_test/transcode/ffmpeg)     Stream #0:0: Audio: mp3, 44100 Hz, stereo, fltp, 192 kb/s
(/srv/airsonic_test/transcode/ffmpeg) Output #0, mp3, to 'pipe:':
(/srv/airsonic_test/transcode/ffmpeg)   Metadata:
(/srv/airsonic_test/transcode/ffmpeg)     TSSE            : Lavf58.20.100
(/srv/airsonic_test/transcode/ffmpeg)     Stream #0:0: Audio: mp3, 44100 Hz, stereo, fltp, 192 kb/s
(/srv/airsonic_test/transcode/ffmpeg) Stream mapping:
(/srv/airsonic_test/transcode/ffmpeg)   Stream #0:0 -> #0:0 (copy)
(/srv/airsonic_test/transcode/ffmpeg) Press [q] to stop, [?] for help
(/srv/airsonic_test/transcode/ffmpeg) size=    1812kB time=00:01:17.29 bitrate= 192.1kbits/s speed= 140x
(/srv/airsonic_test/transcode/ffmpeg) size=    2957kB time=00:02:06.11 bitrate= 192.0kbits/s speed= 119x    
(/srv/airsonic_test/transcode/ffmpeg) size=    4484kB time=00:03:11.29 bitrate= 192.0kbits/s speed= 122x
(/srv/airsonic_test/transcode/ffmpeg) size=    5860kB time=00:04:09.98 bitrate= 192.0kbits/s speed= 126x    
(/srv/airsonic_test/transcode/ffmpeg) video:0kB audio:5860kB subtitle:0kB other streams:0kB global headers:0kB muxing overhead: 0.000750%
```

Notice the `speed=140x` (or somewhere around that value) parts. This is on a Pentium E2220, not a powerhouse by modern standards.

Yetangitu:
> @Yetangitu ... why not just have the "segmenting" portion be a standard part of the transcoding framework?

That is what this patch does, it adds splitting into the transcoding framework by injecting the split command into transcoders or creating a splitting transcoder where no transcoding would be done otherwise.

airsonic-main/src/main/java/org/airsonic/player/domain/MediaFile.java
almet: Hey @jvoisin and @Yetangitu, from what I'm understanding the main reason this is not merged is this discussion. Is that correct?

Yetangitu: This function looks different in the latest version of the patch, it is only used to make it possible to store tracks in the database while the actual start time has been moved to a new field and the end time is calculated by adding the duration to the start time.

I don't know why it has not been merged, I more or less dropped the issue and went on with my life and other projects. I have been running a patched version for a long time now so I know the thing works as intended so that can not be the reason for this patch to remain in limbo.

I just noticed that the patch is suffering from bit rot, not that strange given the fact that I haven't touched it. If the maintainers are not interested in merging it I'll just keep my own patched version up to date. @jvoisin, @muff1nman, @eharris  - what about it? I requested a review several months ago but nothing has happened. If there is no interest in this feature or in my implementation I'd say just close this PR, otherwise let me know so I can update the patch. In that case I'd like for it to be merged within a reasonable time frame so it doesn't accrue more bit rot - I'm a bit short on time so I probably won't be able to keep it up to date all the time. Your call, I'm fine with keeping a patched/forked version running for my own needs.

tesshucom:
Be sure to write web documentation if you intend to add.
Please explain its function and operation specifications.

___

JaudiotaggerParser does not support APE files.
Indeed, it may be.
But, what percentage of users are troubled for it?
Is that a problem that can only be solved with an Airsonic implementation?

A balance may be considered for the boundaries between benefits and disadvantages for public interest.
And that is possible before we start the technology discussion.

The impression that it is extensive and too verbose as a fix to support a single format.
It is not advisable to add complexity for critical features that already have problems.
Unnecessarily deteriorates maintainability.
Even more limited features that are not useful to everyone.

___

Documenting the specification is a compromise.
It adds more functionality to deep functions that are already problematic.
Please prepare the documentation for future issues and user confusion.

tesshucom:
JaudiotaggerParser does not support APE files.
Indeed, it may be.
But, what percentage of users are troubled for it?
Is that a problem that can only be solved with an Airsonic implementation?

A balance may be considered for the boundaries between benefits and disadvantages for public interest.
And that is possible before we start the technology discussion.

___

The impression that it is too extensive as a fix to support a single format.
It is not advisable to add unnecessarily complexity for critical features that already have problems.
If add, I think a web document is needed to prepare for future issues and user confusion.

Airsonic is not only doing simple additions, there are also functions that are omit from Subsonic.
There is a certain rationale for that decision.
For example, video conversion and subtitle mechanisms were complex and deadweight compared to the public interest.
In fact, no one complains when you remove it.

Paradoxically, if the function and purpose are clear, the scope is extremely narrow, and recursive verification is possible, it is easy to add challenging functions.
If there is a risk, everyone will know that deleting or reverting is easy.

tesshucom:
Since the confusion caused by the duplicated content is hindering, we are organizing it.
#495 and #303 don't have a valid PR, but this "issues" seems to follow it.
So I closed them as a "replacement".
If you have different problems, plese reopen them.

airsonic-main/src/main/java/org/airsonic/player/domain/MediaFile.java
Yetangitu: Let's resolve this for now. As long as media_file.path is designated as unique (see https://github.com/airsonic/airsonic/pull/856#issuecomment-531203448) the _gross_ mangling of this field is needed to allow indexed tracks to be added to the database. Either change the database scheme so this field is no longer unique and tracks are identified by media_file.path + media_file.start_position or accept that this _gross_ scheme is in fact needed.

Yetangitu:
It looks like Airsonic development - at least here - has ground to a halt. Is anyone still working on this project?

PR #859, opened 2019-01-05
Fix keyboard shortcuts in play queue

fxthomas:
Using the "space" key to resume playback, as well as +/- to increase/decrease volume, were not working with MediaElement.js. This should make them work again.

### DIFF
--- a/airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
+++ b/airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
@@ -167,15 +167,17 @@
     }
 
     /**
-     * Start playing from the current playlist
+     * Start/resume playing from the current playlist
      */
     function onStart() {
         if (CastPlayer.castSession) {
             CastPlayer.playCast();
         } else if ($('#audioPlayer').get(0)) {
-            var audioPlayer = $('#audioPlayer');
-            if(audioPlayer.paused) {
-                skip(0, audioPlayer.currentTime);
+            if ($('#audioPlayer').get(0).src) {
+                $('#audioPlayer').get(0).play();  // Resume playing if the player was paused
+            }
+            else {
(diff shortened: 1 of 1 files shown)


airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
```diff
@@ -167,16 +167,16 @@
...
             CastPlayer.playCast();
         } else if ($('#audioPlayer').get(0)) {
-            var audioPlayer = $('#audioPlayer');
-            if(audioPlayer.paused) {
-                skip(0, audioPlayer.currentTime);
-            }
+            // Resume playing if the player was paused
+            if ($('#audioPlayer').get(0).src) $('#audioPlayer').get(0).play();
+            // Start the first track if the player was not yet loaded
+            else skip(0);
```
muff1nman: Small nitpick, but can we retain the braces with the if/else branching? I'd like to keep with the existing style.

fxthomas: Yes, done!

airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
```diff
@@ -167,15 +167,17 @@
...
+     * Start/resume playing from the current playlist
      */
     function onStart() {
         if (CastPlayer.castSession) {
             CastPlayer.playCast();
         } else if ($('#audioPlayer').get(0)) {
-            var audioPlayer = $('#audioPlayer');
-            if(audioPlayer.paused) {
-                skip(0, audioPlayer.currentTime);
+            if ($('#audioPlayer').get(0).src) '
```
muff1nman: I think this is missing a bracket here and instead has a stray quote.

PR #863, opened 2019-01-06
Fix the registration of Sonos service and link with security.

Nacrylic:
Main is fix the registration of Sonos service and link with security.

I start with the dev I do in Feature/#386.

1) Remove the CSRF for callback sonos service (ws/sonos/**). The security is implements in SonosLinkkInterceptor except for two Soap action for set the link. 
2) Update the sonos.wsdl, they are some change. But in the wsdl the version didn't change but the content is not the same.

For implementation of link by "Application link" and "Ananymous" I do:

3) Create a new table sonoslink, for save the information about user link to sonos and security verification in SonosLinkkInterceptor
4) Use a new ehCache (7 min) for link code on link process
5) Remove the old method for get user in Sonos soap action implementation and all pass by SonosLinkkInterceptor
6) Add two new jsp page for Application link process
7) Anonymous link fallback on fixed user 'sonos', if the user doen't exist the Anonymous is Unauthorized. So the anonymous link can be see just the 'sonos' user content.
8) Add the possibility in sonos setting for set the link method
9) Add the choice on root Airsonic host address in sonos setting,
10) Add a little console output when save in sonos setting for see some error or success on save in sonos setting
11) Set the getMediaUri and covertArt to "ext/**" request style with JWT token in parameters.

They are some tranlation, I do the english and french. They are some text to translate in the string.xml, I do english and french.

For the Airsonic host setting, they are some problem in the home network when we use the name of machine, they work sometime, sometime not! With the ip they work all time. So Airsonic must be on the ip fixed machine! Or the address mus be set by real DNS. I didn't try, I still not have my own DNS :-)

For the token between Sonos and Airsonic I use a non-expiring authentication token for the moment. Maybe will be change to the expiring authentication token with automatic refresh for more secure.

Nacrylic.


muff1nman:
Thank you for your PR! Due to the size/amount of changes, please allow me time to review fully. Thanks

t3chn0m4g3:
I would really like to see this in the next release. Streaming to Sonos is something I really miss since I switched from Subsonic.

airsonic-main/src/main/java/org/airsonic/player/controller/SonosSettingsController.java
```diff
@@ -50,46 +51,42 @@
...
         settingsService.setSonosEnabled(sonosEnabled);
         settingsService.setSonosServiceName(sonosServiceName);
+        settingsService.setSonosCallbackHostAddress(request.getParameter("callbackHostAddress"));
         settingsService.save();
 
-        sonosService.setMusicServiceEnabled(false, NetworkService.getBaseUrl(request));
-        sonosService.setMusicServiceEnabled(sonosEnabled, NetworkService.getBaseUrl(request));
-    }
+        List<String> returnCodes = sonosService.setMusicServiceEnabled(sonosEnabled, settingsService.getSonosCallbackHostAddress());
+
```
jvoisin: No need to add two blank lines here

airsonic-main/src/main/java/org/airsonic/player/controller/SonosSettingsController.java
```diff
@@ -50,46 +51,42 @@
...
 
-    public void setSettingsService(SettingsService settingsService) {
-        this.settingsService = settingsService;
+        return new ModelAndView("sonosSettings", "model", map);
     }
 
-    public void setSonosService(SonosService sonosService) {
-        this.sonosService = sonosService;
+    private  Map<String, Object> getModel(HttpServletRequest request){
+        Map<String, Object> map = new HashMap<String, Object>();
```
jvoisin: No need to specify the type in the HashMap, since it'll be automatically inferred.

airsonic-main/src/main/java/org/airsonic/player/service/SecurityService.java
```diff
@@ -55,11 +66,27 @@
...
     @Autowired
     private Ehcache userCache;
 
+    @Autowired
+    private Ehcache sonosLinkcodeCache;
+
+    @Autowired
+    private JWTSecurityService jwtSecurityService;
+
+    // This is initialize in GlobalSecurityConfig, is autowired.
```
jvoisin: Maybe something like `// This is initialized in GlobalSecurityConfig, which is autowired`.

airsonic-main/src/main/java/org/airsonic/player/service/SecurityService.java
```diff
@@ -55,11 +66,27 @@
...
+    // Something wrong here, some circular ref, maybe rebuild responsibilities...
```
jvoisin: Please prefix this comment with `TODO ` if it's something that must get fixed.

airsonic-main/src/main/java/org/airsonic/player/service/SecurityService.java
```diff
@@ -336,15 +374,114 @@ protected boolean isFileInFolder(String file, String folder) {
...
+
+    private void setUser(String username) throws SonosSoapFault.LoginUnauthorized {
+        User user = getUserByName(USERNAME_SONOS, true);
+        Authentication authentication = authenticate(user.getUsername(), user.getPassword());
+        SecurityContextHolder.getContext().setAuthentication(authentication);
+    }
+
+        // must be strict 32 character limit on length
+    private String createLinkCode() {
+        return UUID.randomUUID().toString().replace("-", "");
```
jvoisin: If this is security-related, please use [SecureRandom]( https://docs.oracle.com/javase/8/docs/api/java/security/SecureRandom.html ) instead.

airsonic-main/src/main/java/org/airsonic/player/service/SonosService.java
```diff
@@ -116,26 +116,48 @@
     @Resource
     private WebServiceContext context;
 
-    public void setMusicServiceEnabled(boolean enabled, String baseUrl) {
+    /**
+     * Try to enable/disbale the Sonos link.
```
jvoisin: `disbale` → `disable`

airsonic-main/src/main/java/org/airsonic/player/service/SonosService.java
```diff
@@ -116,26 +116,48 @@
...
+     *
+     * @param enabled or not, is not the sonoslink table is clear
```
jvoisin: `@param enabled or not, is not the sonoslink table is clear` → `@param Enable/disable the music service. If disabled, the sonoslink table is cleared`

airsonic-main/src/main/java/org/airsonic/player/service/SonosService.java
```diff
@@ -116,26 +116,48 @@
...
 
         for (String sonosController : sonosControllers) {
             try {
-                new SonosServiceRegistration().setEnabled(baseUrl, sonosController, enabled,
-                                                          sonosServiceName, sonosServiceId);
-                break;
+                if( registration.setEnabled(baseUrl, sonosController, enabled, sonosServiceName, sonosServiceId, authenticationType)){
+
+                    messagesCodes.add("sonossettings.sonoslink.success");
+                    // Clean old links.
```
jvoisin: `Clean` → `Clear`

airsonic-main/src/main/java/org/airsonic/player/service/sonos/SonosLinkkInterceptor.java
```diff
@@ -0,0 +1,121 @@
...
+            }
+        } catch (JAXBException e) {
+           throw new SonosSoapFault.LoginUnauthorized();
+        }
+
+    }
+
+
+    private String getToken(SoapMessage message) throws JAXBException {
+        QName creadentialQName = new QName("http://www.sonos.com/Services/1.1", "credentials");
```
jvoisin: Can't this use `https` instead of `http`?

airsonic-main/src/main/java/org/airsonic/player/service/sonos/SonosServiceRegistration.java
```diff
@@ -31,75 +30,159 @@
...
 import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
 
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 
 /**
+ * <p>Registration with Sonos controller. They are 2 types of registration they are still supported by Sonos.
+ * The ANONYMOUS and APPLICATION_LINK. The third USER_ID, must still work but will be not supported.</p>
```
jvoisin: `must still work ` → `might still work`

airsonic-main/src/main/java/org/airsonic/player/service/sonos/SonosServiceRegistration.java
```diff
@@ -31,75 +30,159 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
 
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 
 /**
+ * <p>Registration with Sonos controller. They are 2 types of registration they are still supported by Sonos.
```
jvoisin: `they are still` → `that are`

airsonic-main/src/main/java/org/airsonic/player/service/sonos/SonosServiceRegistration.java
```diff
@@ -31,75 +30,159 @@
...
  */
+@Component
 public class SonosServiceRegistration {
-
     private static final Logger LOG = LoggerFactory.getLogger(SonosServiceRegistration.class);
 
-    public void setEnabled(String airsonicBaseUrl, String sonosControllerIp, boolean enabled, String sonosServiceName, int sonosServiceId) throws IOException {
+    /**
+     * The type of Authentication fo Sonos, the old want USER_ID, is will be not supported. We must use the
+     * Anonymous or AppLink. The USER_ID
```
jvoisin: What about something like this instead:

```
The type of Authentication for Sonos: we must either use the Anonymous of Applink method, since the old USER_ID one isn't really supported anymore.
```

airsonic-main/src/main/webapp/WEB-INF/jsp/sonoslink.jsp
```diff
@@ -0,0 +1,41 @@
...
+<html><head>
+    <%@ include file="head.jsp" %>
+    <script type="text/javascript">
+        if (window != window.top) {
+            top.location.href = location.href;
+        }
+    </script>
+
+</head>
+<body style="min-width:550px" class="mainframe bgcolor1" onload="document.getElementById('j_username').focus()">
```
jvoisin: Please use the [autofocus]( https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#autofocus )  property directly on the input field instead of using javascript.

airsonic-main/src/main/webapp/WEB-INF/jsp/sonoslinkClose.jsp
```diff
@@ -0,0 +1,23 @@
...
+<html><head>
+    <%@ include file="head.jsp" %>
+    <script type="text/javascript">
+        if (window != window.top) {
+            top.location.href = location.href;
+        }
+    </script>
+
+</head>
+<body style="min-width:550px" class="mainframe bgcolor1" onload="document.getElementById('j_username').focus()">
```
jvoisin: Please use the [autofocus]( https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#autofocus )  property directly on the input field instead of using javascript.

airsonic-main/src/test/java/org/airsonic/player/service/sonos/sonos_manuel_test_.http
```diff
@@ -0,0 +1,43 @@
+#======================================================================================================
+# For manuel test
```
jvoisin: `manuel` → `manual`

airsonic-main/src/main/java/org/airsonic/player/service/SecurityService.java
Nacrylic: So I re-evaluate that code, this service must not be get by component scan but declared in global security. And it will be move to the security package. 

All initialisation about security it's set there and the service just use the global component.

airsonic-main/src/main/java/org/airsonic/player/service/SecurityService.java
Nacrylic: Right.

airsonic-main/src/main/webapp/WEB-INF/jsp/sonoslinkClose.jsp
Nacrylic: I copied/pasted the code from the login page and left it as such so as not to mix styles too much.

airsonic-main/src/main/webapp/WEB-INF/jsp/sonoslink.jsp
Nacrylic: I copied/pasted the code from the login page and left it as such so as not to mix styles too much.

jwater7:
Someone needs to merge this in before it gets too out of date.  Most of the code review comments are just nitpicks as far as I can tell.

jwater7:
In favor of something else?  Can you point us to any work that is happening on this topic?

jvoisin:
Check #1177 :)

PR #864, opened 2019-01-07
Fix #860 (external database performance) by using connection pooling …

### LINKED ISSUE #860: External (Postgresql) database much slower than hsql
I've been running Airsonic with an external (_Postgresql_) database for about a year now. For developing CUE support (https://github.com/airsonic/airsonic/pull/856) I added a testing instance using the 'legacy' (hsql) database and pointed it at the same library of about 200.000 tracks. This made clear just how big the performance discrepancy between the two is. That _hsql_ would be faster was to be expected given that it is mostly in-memory but I do wonder whether there is something else in play: running a full scan over the entire library takes about 50 minutes when using _hsql_ versus  ~5 hours when using _Postgresql_. In both cases the database is kept on the same SSD and runs on the same server as Airsonic runs on. That server is nothing special, a rather memory-starved (2GB) Intel SS4200 with a dual-core Pentium E2220. I did some experiments with:

- different jdbc drivers; pgjdbc-ng (http://impossibl.github.io/pgjdbc-ng/) ended up being twice as slow as the default driver
- connecting through unix sockets (using https://github.com/fiken/junixsocket - no noticeable performance difference)
- connecting through localhost instead of the public address (as expected this made no difference)

While scanning using _hsql_ the java process sits at ~60% with peaks up to ~170% (dual-core) when committing the database log (it does this several times during a scan), process size goes up to ~700MB (I got GC/heap errors before, had to raise the maximum heap size from 200MB to 600MB (as said this is a memory-starved server).  When using _Postgresql_ java hovers around 15% and postgres around 9%, java process size lies around 500MB and the system has plenty of idle time. Average disk IO lies at 100KB/s for the postgres wal writer process (which writes the journal) with occasional peaks around 1MB for postgres writer. All this is clearly within system capacity and the slowdown does not seem to be caused by IO or CPU starvation. Something is slowing the combination Airsonic/Postgresql down but I have not tracked down the culprit yet.

Yetangitu:
Fixes #860 by using connection pooling (using commons-dbcp2). I only tested this with _Postgresql_ where it turned a 5 hour scan job (200.000 track library) into a 15 minute scan job. 

This change only applies to databases using the embedded jdbc connector (using 'embed'), using dbcp2 with the 'legacy' _hsql_ database does not work (it all ends with `Invocation of init method failed; nested exception is java.lang.AbstractMethodError: org.hsqldb.jdbc.jdbcConnection.isValid(I)Z`). While _hsql_ perfoms well enough without connection pooling it does make about 10 connections per library item on a library scan (tested with a 200.000 track library, a single scan used 2.1 million connections) so it might make sense to get connection pooling to work here as well.


muff1nman:
Excellent suggestion. Seems that the Travis build didn't find the latest commit - I restarted it.

Yetangitu:
Travis is still on strike it seems, I notice one of the errors is (line 871, also elsewhere) `nested exception is java.lang.AbstractMethodError: org.hsqldb.jdbc.jdbcConnection.isValid(I)Z`. That is the error which is generated when trying to use dbcp2 with _hsql_, something which this PR does not enable (it only changes the `embed` profile). Something seems messed up.

Yetangitu:
Travis still fails on_hsql_, strangely enough. This change should not touch any code which involves _hsql_.

The problem seems to lie in the test cycle which uses _hsql_ with the `embed` profile, something which I already found problematic. Either the tests need to be changed - the best option as it currently tests a scenario which is known not to work - or the _hsql_ case needs to be made dbcp2-compatible.

Yetangitu:
Here's why it fails: _hsql_ 1.8 does not support [Connection.isValidI() (since Java 1.6)](https://docs.oracle.com/javase/8/docs/api/java/sql/Connection.html#isValid-int-) which is used in dbcp2.

The solution would be to upgrade _hsql_ (but v2.0 is a different beast from v1.8) or change the tests so that they don't use the `embed` profile.

Yetangitu:
The last commit changes the tests to use the `legacy` profile, this avoids the incompatibility between _hsql_ v1.8 and dbcp2. Was there any specific reason to use the `embed` profile instead of `legacy`? If not, this should solve the issue and with that the PR can be merged.

Yetangitu:
...sigh... it is a sad state of affairs when it takes more time to placate the testing subsystem to behave than it does to write the software in the first place. Now it complains about the dbcp2 dependency not being used, which it isn't as that would break the test build as that runs with _hsql_ which is incompatible with such.. etcetera.

muff1nman:
Ah dang sorry for roller coaster rabbit hole you've gone through. Give me a day or two and I can help get this the rest of the way there.

Yetangitu:
Thanks, this being my first adventure with this sort of maven-auto-magic (I tend to live closer to the compiler) and my somewhat underwhelming build system taking even longer than Travis to run those integration tests it does get somewhat laborious to tame all those labour-saving software robots.

muff1nman:
This got me most of the way:

```
diff --git a/airsonic-main/pom.xml b/airsonic-main/pom.xml
index d3d3d7e..b4eda4b 100755
--- a/airsonic-main/pom.xml
+++ b/airsonic-main/pom.xml
@@ -148,6 +148,7 @@
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-dbcp2</artifactId>
             <version>2.5.0</version>
+            <scope>runtime</scope>
             <!-- this wants to pull in commons-logging v1.2 which conflicts with other dependencies -->
             <exclusions>
                 <exclusion>
@@ -605,23 +606,6 @@
                 </executions>
             </plugin>
             <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-dependency-plugin</artifactId>
-                <executions>
-                    <execution>
-                        <id>analyze</id>
-                        <goals>
-                            <goal>analyze-only</goal>
-                        </goals>
-                        <configuration>
-                            <ignoredUnusedDeclaredDependencies>
-                                <ignoredUnusedDeclaredDependency>org.apache.commons:commons-dbcp2</ignoredUnusedDeclaredDependency>
-                            </ignoredUnusedDeclaredDependencies>
-                        </configuration>
-                    </execution>
-                </executions>
-            </plugin>
-            <plugin>
                 <groupId>org.owasp</groupId>
                 <artifactId>dependency-check-maven</artifactId>
             </plugin>
```

Basically, we don't actually need to exclude the dbcp2 dependency if it is correctly marked as a runtime dependency (it isn't needed for compile).

There looks to be a new CVE which is now failing the build - but that is unrelated to this PR.

Yetangitu:
OK, I'll give it a go somewhat later today, all the while thinking of how easy things _seemed_ to be before they were complicated by Rube Goldstein-like build systems.

PR #866, opened 2019-01-08
fix #865 JaudiotaggerParser not support APE format

### LINKED ISSUE #865: JaudiotaggerParser does not support APE files
## Problem description

When scan APE files, there is error in log:
```
org.jaudiotagger.audio.exceptions.CannotReadException: No Reader associated with this extension:ape
	at org.jaudiotagger.audio.AudioFileIO.readFile(AudioFileIO.java:353) ~[jaudiotagger-2.2.5.jar!/:na]
	at org.jaudiotagger.audio.AudioFileIO.read(AudioFileIO.java:198) ~[jaudiotagger-2.2.5.jar!/:na]
	at org.airsonic.player.service.metadata.JaudiotaggerParser.getRawMetaData(JaudiotaggerParser.java:86) ~[classes!/:10.2.0-SNAPSHOT]
	at org.airsonic.player.service.metadata.MetaDataParser.getMetaData(MetaDataParser.java:47) [classes!/:10.2.0-SNAPSHOT]
	at org.airsonic.player.service.MediaFileService.createSingleFileAlbumChildren(MediaFileService.java:445) [classes!/:10.2.0-SNAPSHOT]
	at org.airsonic.player.service.MediaFileService.updateChildren(MediaFileService.java:406) [classes!/:10.2.0-SNAPSHOT]
	at org.airsonic.player.service.MediaFileService.getChildrenOf(MediaFileService.java:194) [classes!/:10.2.0-SNAPSHOT]
	at org.airsonic.player.service.MediaScannerService.scanFile(MediaScannerService.java:248) [classes!/:10.2.0-SNAPSHOT]
	at org.airsonic.player.service.MediaScannerService.scanFile(MediaScannerService.java:252) [classes!/:10.2.0-SNAPSHOT]
	at org.airsonic.player.service.MediaScannerService.scanFile(MediaScannerService.java:252) [classes!/:10.2.0-SNAPSHOT]
	at org.airsonic.player.service.MediaScannerService.doScanLibrary(MediaScannerService.java:188) [classes!/:10.2.0-SNAPSHOT]
	at org.airsonic.player.service.MediaScannerService.access$000(MediaScannerService.java:45) [classes!/:10.2.0-SNAPSHOT]
	at org.airsonic.player.service.MediaScannerService$2.run(MediaScannerService.java:157) [classes!/:10.2.0-SNAPSHOT]
```
It seems Jaudiotagger does not support APE format and I do not see it supports APE either on its website. However, JaudiotaggerParser claims it [supports APE](https://github.com/airsonic/airsonic/blob/77ca475fbe066b453c75e8614fd738a4667ff252/airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java#L306) which should not.
I tried FFmpegParser instead and it is successful with APE files. But FFmpegParser currently only claims to support video files. I think it should claims to support APE files.

RuralHunter:
See the detail in #865. We can use FFmpegParser  to parse APE format

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/metadata/FFmpegParser.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/metadata/FFmpegParser.java
@@ -144,7 +144,7 @@ public class FFmpegParser extends MetaDataParser {
                 return true;
             }
         }
-        return false;
+        return "ape".equals(format);
     }
 
     public void setTranscodingService(TranscodingService transcodingService) {
--- a/airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
@@ -303,7 +303,6 @@ public class JaudiotaggerParser extends MetaDataParser {
                 format.equals("wav") ||
                 format.equals("mpc") ||
                 format.equals("mp+") ||
-                format.equals("ape") ||
                 format.equals("wma");
     }
 


airsonic-main/src/main/java/org/airsonic/player/service/metadata/FFmpegParser.java
```diff
@@ -144,7 +144,7 @@ public boolean isApplicable(File file) {
                 return true;
             }
         }
-        return false;
+        return "ape".equals(format);
```
jvoisin: I'm not super-duper comfortable with this kind of hardcoding. Could you please make it a bit more generic, so that we can do the same for more file formats?

RuralHunter: hmm...I don't quite understand. This is just handle the format in the similar way of [JaudiotaggerParser.java](https://github.com/airsonic/airsonic/blob/ac0a722a105612b652df311600c94f2ea93cc8b2/airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java#L298)

eharris:
I think this PR is now obsolete with #1237.  Closing in favor of that one.

RuralHunter:
I think PR #1237 removes the support of APE from JaudiotaggerParser which is only part of the work. Another part is to add support of APE in FFmpegParser, which I think is still required?

PR #869, opened 2019-01-25
Pull translations from transifex

jooola:
French has been translated at 100% by a contributor on Transifex !

Thanks Doryan !
Still need some reviewing.


jooola:
Some single quote escaping are not respected, I think we had trouble with this last time we tried to change this. Need to dig in old tickets again.

PR #878, opened 2019-02-10
Update jackson version

bfkelsey:


bfkelsey:
@muff1nman it looks like there is a security issue in the chameleon library that you maintain.

PR #880, opened 2019-02-14
With jetty

tesshucom:
If tomcat is a prerequisite product, there is no need to deal immediately.
Although it is for developers, operation with jetty is possible if a few points are corrected.

Due to the complexity of the problem, we described the situation individually in the following issues.

https://github.com/tesshucom/airsonic/issues/5


tesshucom:
Uh.
It seems that a timeout error has occurred with pr check.
In my travis environment this did not happen ...

>No output has been received in the last 10m0s.

tesshucom:
Comment fixed.

muff1nman:
Merged in https://github.com/airsonic/airsonic/commit/131713aaf423cfe19962c0e2dd88453a7b523463.

PR #887, opened 2019-02-21
Fix #885, fix translations

### LINKED ISSUE #885: No "Save" button in"Advanced" Parameters. Impossible to save SMTP server definition + Updated file for French translation
Issue with Advanced Configuration. See below

jooola:
This is supposed to fix #885 
- pulled latest translations from Transifex
- added a script the helps replace any illegal char like éèà... to its Unicode representation (Building a bash script might be simpler)
- Ran the script against the translations files


jooola:
Haven't tested this yet.

jooola:
I will change to go script into python since it is way more commonly used. So please do not merge yet

muff1nman:
go or python is fine. doesn't matter to me one way or the other.

jooola:
Ok, though It could be easier for future contributors. This is good to go then.

airsonic-main/src/main/resources/org/airsonic/player/i18n/ResourceBundle_nl.properties
```diff
@@ -600,135 +600,136 @@ edittags.year=Jaar
...
+helppopup.smtpPort.title=SMTP-poort
+helppopup.smtpPort.text=<p>De serverpoort waarmee moet worden verbonden voor uitgaand e-mailverkeer.</p>
+helppopup.smtpCredentials.title=SMTP-inloggegevens
+helppopup.smtpCredentials.text=<p>De inloggegevens voor het verbinden met de SMTP-server. Laat leeg om te verbinden zonder authenticatie.</p>
+helppopup.smtpEncryption.title=SMTP-versleuteling
+helppopup.smtpEncryption.text=<p>De te gebruiken versleutelingsmethode voor verbindingen met de SMTP-server. Kies "Geen" voor geen versleuteling.</p>
+helppopup.smtpFrom.title='Van'-adres
+helppopup.smtpFrom.text=<p>Het e-mailadres van de afzender voor e-mails afkomstig van de Airsonic-server. Moet een geldig e-mailadres zijn.</p>
 helppopup.captcha.title=CAPTCHA
+<<<<<<< HEAD
```
muff1nman: Looks like a leftover merge..

muff1nman:
Can we avoid committing the compiled go binary? The source file should be good enough.

jooola:
Yup sure, but do we still need to convert some chars encoding or has it been fixed after switching to UTF-8 compilation and JSP pages ?

muff1nman:
hmm, we might be okay now with things in utf8.

jooola:
Ok this should be good to go

PR #876, opened 2019-02-08
Upgrade HSQLDB to 2.x

harawata:
The primary goal is to fix #872 and its side effects like #531 .

PR #895, opened 2019-03-08
Make embedded HSQLDB more reliable

fxthomas:
This is a precursor PR for #876 that should make HSQLDB more reliable in a few aspects before we can switch to 2.x (thanks @harawata for the help in testing!):

* cff97ea made sure the log file cannot grow uncontrollably large ; big operations (scan/cleanup) now automatically run a checkpoint as well.
* The Liquibase migration introduced in cff97ea now works for HSQLDB 1.x and 2.x.
* I also went ahead and implemented proper shutdown for the HSQLDB database.

Note that the migration docs for HSQLDB heavily recommend running a proper `SHUTDOWN` (or at least a `CHECKPOINT`) when migrating from 1.8 to 2.x.

Thoughts?


airsonic-main/src/main/java/org/airsonic/player/Application.java
```diff
@@ -56,6 +57,11 @@ public ServletRegistrationBean dwrServletRegistrationBean() {
         return servlet;
     }
 
+    @Bean
+    public TerminateBean terminateBean() {
+        return new TerminateBean();
+    }
+
```
muff1nman: I would actually just register this in the xml file for the legacy profile: https://github.com/airsonic/airsonic/blob/master/airsonic-main/src/main/resources/applicationContext-db-legacy.xml.

airsonic-main/src/main/java/org/airsonic/player/dao/GenericDaoHelper.java
```diff
@@ -35,4 +43,35 @@ public NamedParameterJdbcTemplate getNamedParameterJdbcTemplate() {
...
+
+        // Since this method is a best-effort operation, we don't want to show
+        // a message if the checkpoint failed ; just assume the operation is
+        // unsupported.
+        catch(java.sql.SQLException e) {
+            LOG.debug("An exception occurred during database checkpoint: " + e.toString());
+        }
+
+        return false;
+    }
```
muff1nman: I would actually extract this into its own bean as well as this logic is not "Generic" but very specific to hsql. In other words, I would create a new DAO interface specific for checkpointing. The default implementation should be a noop. For Hsql, you would have this logic. Does that make sense?

airsonic-main/src/main/java/org/airsonic/player/spring/TerminateBean.java
```diff
@@ -0,0 +1,55 @@
...
+        } finally {
+            try {
+                if(conn != null)
+                    conn.close();
+            } catch(Exception ex) {
+                ex.printStackTrace();
+            }
+        }
+    }
+}
```
muff1nman: I would name this bean to be more specific to hsql.

airsonic-main/src/main/java/org/airsonic/player/Application.java
fxthomas: Sounds reasonable, will do that!

airsonic-main/src/main/java/org/airsonic/player/dao/GenericDaoHelper.java
fxthomas: It does, though I'm a bit uncertain on how to do that. If I overwrite the `daoHelper` bean in the legacy application configuration, does that work? Or is there a better way?

airsonic-main/src/main/java/org/airsonic/player/spring/TerminateBean.java
fxthomas: Good point, will do that!

airsonic-main/src/main/java/org/airsonic/player/dao/GenericDaoHelper.java
fxthomas: Overwriting was pretty easy, it should work now. As a bonus I've moved all the legacy-specific logic to the `LegacyHsqlDaoHelper` class so that everything is in the same place!

airsonic-main/src/main/java/org/airsonic/player/dao/DaoHelper.java
```diff
@@ -46,4 +46,14 @@
...
+
+    /**
+     * Tries to perform a checkpoint against the database, if supported
+     *
+     * Database checkpoints will make sure that the database is written on the disk
+     * and optimize on-disk storage.
+     *
+     * @return true if the checkpoint succeeded, false otherwise
+     */
+    public default boolean checkpoint() { return false; }
```
muff1nman: It dones't look like the return value is ever checked. I'd just leave the return value `void` - probably better to throw an exception if it fails anyways.

airsonic-main/src/main/java/org/airsonic/player/controller/MusicFolderSettingsController.java
```diff
@@ -93,9 +97,15 @@ protected void formBackingObject(@RequestParam(value = "scanNow",required = fals
...
     private void expunge() {
+        LOG.info("Cleaning database...");
+        LOG.info("Deleting non-present artists...");
         artistDao.expunge();
+        LOG.info("Deleting non-present albums...");
         albumDao.expunge();
+        LOG.info("Deleting non-present media files...");
         mediaFileDao.expunge();
+        LOG.info("Database cleanup complete.");
+        mediaFileDao.checkpoint();
```
muff1nman: These logs should probably be debug or trace.

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
```diff
@@ -0,0 +1,60 @@
...
+        super(dataSource);
+    }
+
+    @Override
+    public boolean checkpoint() {
+        // HSQLDB (at least version 1) does not handle automatic checkpoints very well by default.
+        // This makes sure the temporary log is actually written to more persistent storage.
+        LOG.info("Database checkpoint in progress...");
+        getJdbcTemplate().execute("CHECKPOINT DEFRAG");
+        LOG.info("Database checkpoint complete.");
```
muff1nman: these log lines should probably be trace.

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
```diff
@@ -0,0 +1,60 @@
...
+        Connection conn = null;
+        try {
+            // Properly shutdown the embedded HSQLDB database.
+            LOG.info("Database shutdown in progress...");
+            JdbcTemplate jdbcTemplate = getJdbcTemplate();
+            conn = DataSourceUtils.getConnection(jdbcTemplate.getDataSource());
+            conn.setAutoCommit(true);
+            jdbcTemplate.execute("SHUTDOWN");
+            LOG.info("Database shutdown complete.");
+
```
muff1nman: I'd also change these logs to trace.

muff1nman:
Couple more minor things, but overall looks well structured.

fxthomas:
Alright, PR updated with the last couple of cleanups. This should be ready to merge!

Issue #896, opened 2019-03-09
m4b files are not recoginzed


SenorSmartyPants:
m4b files are not recognized. Renaming file extension to m4a causes files to be found properly.

Please add m4b to list of extensions that airsonic recoginizes.

eharris:
Resolved with #897.

PR #897, opened 2019-03-10
Support m4b files

SenorSmartyPants:
anywhere m4a is referenced in java code add support for m4b as well.

This is in reference to #896 

Just adding m4b to the music file types in the UI did not make m4b files recoginized by the scanner.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/SettingsService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/SettingsService.java
@@ -138,7 +138,7 @@ public class SettingsService {
     private static final String DEFAULT_IGNORED_ARTICLES = "The El La Los Las Le Les";
     private static final String DEFAULT_SHORTCUTS = "New Incoming Podcast";
     private static final String DEFAULT_PLAYLIST_FOLDER = Util.getDefaultPlaylistFolder();
-    private static final String DEFAULT_MUSIC_FILE_TYPES = "mp3 ogg oga aac m4a flac wav wma aif aiff ape mpc shn mka opus";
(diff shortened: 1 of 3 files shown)


SenorSmartyPants:
Travis build failed, but none of my code changes should have broken anything. Not sure what to do about that.

SenorSmartyPants:
I think this is related to JaudiotaggerParser.java. #866 fails build as well and it touches the same file.

SenorSmartyPants:
Force pushed to fix my screwed up rebase.

muff1nman:
Doesn't this need transcoding like m4a files?

SenorSmartyPants:
They are practically identical to m4a files, except the audio in them is audiobooks. 

It would depend on the client software if they need transcoding. But for apps like Booksonic they would not.

But either way, airsonic currently supports adding m4a files to the library with no problem. This should merely do the same thing if the extension is m4b.

muff1nman:
Sure, but if you want the same behavior as m4a, we need an entry in the transcoding table similar [similar to m4a](https://github.com/airsonic/airsonic/blob/master/airsonic-main/src/main/resources/liquibase/legacy/schema28.xml#L161)

SenorSmartyPants:
Ok, I'm very new to this code base. I thought the transcoding was covered with the information on the transcoding tab in settings.

Is there a tool for making database changes and saving the script for that to include in a PR?

jooola:
Ref #896

muff1nman:
We use liquibase for managing database changes. I'd use a good text editor that can properly format XML files, reference the liquibase [documentation](http://www.liquibase.org/documentation/) and reference the [previous changelogs we already have](https://github.com/airsonic/airsonic/blob/master/airsonic-main/src/main/resources/liquibase/legacy/schema28.xml#L161) to create a new changelog.

PR #903, opened 2019-03-18
Add a bunch of hardening to the systemd unit

jvoisin:
I tested this locally, it's working on my machine™
Nothing fancy, no new privileges, private stuff (/tmp, …),
no exotic devices/access/…, …

- [ ] Add a more relaxed profile for people that are allowing their user to have r/w access on the media

### DIFF
--- a/contrib/airsonic.service
+++ b/contrib/airsonic.service
@@ -21,6 +21,8 @@ ExecStart=/usr/bin/java \
 User=airsonic
 Group=airsonic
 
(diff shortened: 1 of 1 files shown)


muff1nman:
I think we discussed over IRC, but can you comment out some of the more aggressive isolations with comments explaining that users can uncomment if X? I.e. ReadWrite - uncomment if you do not want airsonic to have write access to your music folder. This may break some functionality.

muff1nman:
Merged

PR #905, opened 2019-03-20
Add some checks to avoid stackstrace

jvoisin:
Blindly casting to integer without a try-catch
can lead to a stacktrace for some values.

Signed-off-by: jvoisin <julien.voisin@dustri.org>

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/CoverArtController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/CoverArtController.java
@@ -150,7 +150,12 @@ public class CoverArtController implements LastModified {
         if (id.startsWith(PODCAST_COVERART_PREFIX)) {
             return createPodcastCoverArtRequest(Integer.valueOf(id.replace(PODCAST_COVERART_PREFIX, "")), request);
         }
-        return createMediaFileCoverArtRequest(Integer.valueOf(id), request);
+        try {
+            return createMediaFileCoverArtRequest(Integer.valueOf(id), request);
+        } catch (java.lang.NumberFormatException x) {
+            LOG.warn("Invalid cover id: " + id, x.toString());
+            return null;
+        }
     }
 
     private CoverArtRequest createAlbumCoverArtRequest(int id) {
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/PlayerSettingsController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/PlayerSettingsController.java
@@ -26,6 +26,8 @@ import org.airsonic.player.service.PlayerService;
 import org.airsonic.player.service.SecurityService;
 import org.airsonic.player.service.TranscodingService;
 import org.apache.commons.lang.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Controller;
 import org.springframework.ui.Model;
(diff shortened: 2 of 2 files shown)


airsonic-main/src/main/java/org/airsonic/player/controller/CoverArtController.java
```diff
@@ -150,7 +150,12 @@ private CoverArtRequest createCoverArtRequest(HttpServletRequest request) {
         if (id.startsWith(PODCAST_COVERART_PREFIX)) {
             return createPodcastCoverArtRequest(Integer.valueOf(id.replace(PODCAST_COVERART_PREFIX, "")), request);
         }
-        return createMediaFileCoverArtRequest(Integer.valueOf(id), request);
+        try {
+          return createMediaFileCoverArtRequest(Integer.valueOf(id), request);
+        } catch (java.lang.NumberFormatException x) {
+          LOG.warn("Illegal cover id: " + id , x);
```
eharris: Invalid, not illegal

eharris: Invalid, not illegal

airsonic-main/src/main/java/org/airsonic/player/controller/PlayerSettingsController.java
```diff
@@ -151,9 +155,21 @@ protected String doSubmitAction(@ModelAttribute("command") PlayerSettingsCommand
 
     private void handleRequestParameters(HttpServletRequest request) throws Exception {
         if (request.getParameter("delete") != null) {
-            playerService.removePlayerById(ServletRequestUtils.getIntParameter(request, "delete"));
+            int id = ServletRequestUtils.getIntParameter(request, "delete");
+            try {
+                playerService.removePlayerById(id);
+            } catch (java.lang.NumberFormatException x) {
+                LOG.warn("Illegal played id: " + id , x);
```
eharris: Should be `player` not `played` and invalid, not illegal

airsonic-main/src/main/java/org/airsonic/player/controller/PlayerSettingsController.java
```diff
@@ -151,9 +155,21 @@ protected String doSubmitAction(@ModelAttribute("command") PlayerSettingsCommand
...
+                return;
+            }
         } else if (request.getParameter("clone") != null) {
-            playerService.clonePlayer(ServletRequestUtils.getIntParameter(request, "clone"));
+            int id = ServletRequestUtils.getIntParameter(request, "clone");
+            try {
+                playerService.clonePlayer(id);
+            } catch (java.lang.NumberFormatException x) {
+                LOG.warn("Illegal played id: " + id , x);
```
eharris: Should be `player` not `played`, and invalid rather than illegal

eharris:
You need to rebase on master, as you have checkstyle issues to fix, and also want to avoid further Travis failures.

airsonic-main/src/main/java/org/airsonic/player/controller/CoverArtController.java
```diff
@@ -150,7 +150,12 @@ private CoverArtRequest createCoverArtRequest(HttpServletRequest request) {
         if (id.startsWith(PODCAST_COVERART_PREFIX)) {
             return createPodcastCoverArtRequest(Integer.valueOf(id.replace(PODCAST_COVERART_PREFIX, "")), request);
         }
-        return createMediaFileCoverArtRequest(Integer.valueOf(id), request);
+        try {
+            return createMediaFileCoverArtRequest(Integer.valueOf(id), request);
+        } catch (java.lang.NumberFormatException x) {
+            LOG.warn("Invalid cover id: " + id, x);
```
eharris: I would not want a full stack trace for these caught `NumberFormatException`s.  Maybe use `x.toString()` instead?

muff1nman:
Why are we so concerned with burrying stack traces. This is not a good practice.

PR #907, opened 2019-03-20
Remove momentjs

jvoisin:
MomentJS was only used in a single javascript function,
to do Seconds → (Minutes, Seconds). This commit
replaces this with a hand-rolled version, and removes
MomentJS.

Signed-off-by: jvoisin <julien.voisin@dustri.org>


airsonic-main/src/main/webapp/script/playQueue/javaJukeboxPlayerControlBar.js
```diff
@@ -68,13 +68,10 @@ function updateJavaJukeboxPlayerControlBar(song){
...
 function songTimeAsString(timeInSeconds) {
-    var m = moment.duration(timeInSeconds, 'seconds');
-    var seconds = m.seconds();
-    var secondsAsString = seconds;
-    if (seconds < 10) {
-        secondsAsString = "0" + seconds;
-    }
-    return m.minutes() + ":" + secondsAsString;
+    var minutes = Math.floor(timeInSeconds / 60);
+    var seconds = seconds - minutes * 60;
```
muff1nman: Shouldn't this be `var seconds = timeInSeconds - minutes * 60;`?

PR #917, opened 2019-03-24
Remove the jwplayer script

jvoisin:
Since it was replaced with mediaelement in
airsonic 10, it can safely be removed.


muff1nman:
I think we might need to finish getting the external player updated before we can remove it.

PR #920, opened 2019-03-25
Simplify the travis file

jvoisin:
Simplify the travis file
    
- Remove `sudo` since it's deprecated, along with the useless `apt update` call
- Remove the manual maven installation since it's already present on the travis image by default.
    
This should reduce a bit the time taken for a whole CI run, and improve a bit it's maintainability.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 language: java
-sudo: required
 jdk:
   - oraclejdk8
 services:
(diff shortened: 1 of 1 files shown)


muff1nman:
We can try this - I believe at some point we needed the newer version of maven, but chances are that the default travis image has a newer version of maven.

Issue #853, opened 2018-12-30
Remove UI dependency to externally hosted static assets


zrav:
## Problem description

Currently Airsonics web-UI makes use of Google hosted font files. This causes problems in offline / airgapped networks.

### Steps to reproduce

Disconnect network from internet, access and use the Airsonic web-UI. Notice long delays, as the browser waits for the asset requests to time out.

## System information

 * **Airsonic version**: v10.1.2

## Additional notes

There's no reason these files couldn't be provided by Airsonic itself.

jooola:
Haha, this one has a big history ^^.

Duplicate of #174 #399

jooola:
Could be cherry picked from https://github.com/Libresonic/libresonic/pull/173

Issue #712, opened 2018-03-05
Feature: Add option to disable any external links/lookups/connections


eharris:
I would like to install an instance of airsonic on a private network that for security reasons does not allow outside internet access.  Any attempt to connect to outside services is considered a security issue.  For this reason, I'd need to be able to completely disable the server from ever attempting to connect to any other outside host or service, it would only be for local private network access.

This would include the need to disable the ability to check for updates, look up cover art, connect to last.fm, google, wikipedia, allmusic, etc.  Even dns lookups might be a problem as they could just hang with no response.

I know that the user administration allows for users to be restricted from being able to change the cover art or share files, but I don't see any way to disable the links to outside search services.  And there should be a way to ensure that nothing else the server does under the covers will ever result in a connection attempt or a link that might be followed to an off-net service.

jooola:
Yes I think we already had some similar ticket asking to disable last.fm artist info fetch and other.

As work around I suggest you to add those domains into your hosts file and block them (redirect to 127.0.0.1)

Still I don't know what are all the external services that airsonic uses so you can try those and test. You could also try to monitor yo out trafic and see if they are any other services.

From what I know : 
- Last FM Scrobble `post.audioscrobbler.com`
- Last FM `last.fm`
- Chart Lyrics `api.chartlyrics.com`

Unsure :
- Podcasts `itunes.com`
- Sonos `sonos.com`

I would be surprised if those are used at all but adding them here anyway :
- System `www.w3.org`
- System `www.springframework.org`
- System `www.liquibase.org`
- System `www.macromedia.com`

Same for those : 
```
<%@ page session="false"%>
<%@ taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core" %>
<%@ taglib prefix="fmt" uri="http://java.sun.com/jsp/jstl/fmt" %>
<%@ taglib prefix="fn" uri="http://java.sun.com/jsp/jstl/functions" %>
<%@ taglib prefix="spring" uri="http://www.springframework.org/tags" %>
<%@ taglib prefix="form" uri="http://www.springframework.org/tags/form" %>
<%@ taglib prefix="sec" uri="http://www.springframework.org/security/tags" %>
<%@ taglib prefix="sub" uri="http://airsonic.org/taglib/sub" %>
<%@ taglib prefix="str" uri="http://jakarta.apache.org/taglibs/string-1.1" %>
```
Maybe they are only fetched on client side ?
Like Google fonts.

eharris:
[...]

Issue #439, opened 2017-05-26
Remove Ubuntu fonts and store Roboto locally


issuemover631:
<a href="https://github.com/jooola"><img src="https://avatars2.githubusercontent.com/u/19195485?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [jooola](https://github.com/jooola)**
_Friday May 26, 2017 at 17:23 GMT_
_Originally opened as https://github.com/Libresonic/libresonic/pull/438_

----

Ubuntu fonts are only used in the `<h1>` element. I suggest to remove this font to speed up the loading time !

fix #398 

----

_**[jooola](https://github.com/jooola)** included the following code: https://github.com/Libresonic/libresonic/pull/438/commits_

Issue #399, opened 2017-05-12
Find new fonts and provide them locally


issuemover631:
<a href="https://github.com/jooola"><img src="https://avatars2.githubusercontent.com/u/19195485?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [jooola](https://github.com/jooola)**
_Friday May 12, 2017 at 01:58 GMT_
_Originally opened as https://github.com/Libresonic/libresonic/issues/398_

----

I think we should provide fonts locally instead of loading them from the google api

jooola:
Moved to #853

PR #922, opened 2019-03-26
Remove the external fonts usage

jvoisin:
This is a radical (and simple) solution to #853, #399, #439, #174 and partially #712: deleting the external fonts.

Feel free to close this PR if this is too radical.


jvoisin:
This is how my instance is looking without external fonts:

.

jooola:
Good idea, I'm fine with this.

jooola:
I still would use the Web safe fonts list:
```css
font-family: Verdana, Tahoma, Geneva, sans-serif;
```
https://www.w3schools.com/cssref/css_websafe_fonts.asp

PR #925, opened 2019-03-26
Fix various minor issues found by LGTM

jvoisin:
- Unnecessary boxing
- Integer overflow
- Path traversal via zip
- Dangerous synchronisation pattern


jvoisin:
The path traversal could lead to an RCE if the attacker is able to write a file in `/var/airsonic/transcode/`.

PR #926, opened 2019-03-26
Fix two NULL-deref

jvoisin:


jvoisin:
How would you gracefully handle this instead?

PR #929, opened 2019-03-27
Update mediaelementjs

jvoisin:
Update mediaelement from 4.1.1 to 4.2.9
    
I've tested this on mp3, ogg and flac, everything is working.
   
Apart from fixing a decent amount of bugs (https://github.com/mediaelement/mediaelement/releases), this has the nice side effect of removing a copy of jquery 1.9.1. It will also break the compatibility with IE9 and IE10, but I would argue that this is a feature not a bug.


jvoisin:
I seems that there is an issue travis-ci side, unrelated to this PR :/

PR #931, opened 2019-03-27
Fix a couple of NULL-deref found by coverity

jvoisin:
See https://scan.coverity.com/projects/airsonic for details

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/HomeController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/HomeController.java
@@ -142,7 +142,7 @@ public class HomeController  {
         List<Album> result = new ArrayList<>();
         for (MediaFile mediaFile : ratingService.getHighestRatedAlbums(offset, count, musicFolders)) {
             Album album = createAlbum(mediaFile);
-            Double averateRating = ratingService.getAverageRating(mediaFile);
+            Double averageRating = ratingService.getAverageRating(mediaFile);
             if (averageRating == null) {
               continue;
             }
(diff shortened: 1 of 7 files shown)


airsonic-main/src/main/java/org/airsonic/player/controller/HomeController.java
```diff
@@ -142,7 +142,11 @@ protected ModelAndView handleRequestInternal(HttpServletRequest request) throws
         List<Album> result = new ArrayList<>();
         for (MediaFile mediaFile : ratingService.getHighestRatedAlbums(offset, count, musicFolders)) {
             Album album = createAlbum(mediaFile);
-            album.setRating((int) Math.round(ratingService.getAverageRating(mediaFile) * 10.0D));
+            Double averageRating = ratingService.getAverageRating(mediaFile);
+            if (averageRating == null) {
```
fxthomas: Isn't it better to assume `averageRating = 0` in case it doesn't exist?

jvoisin: The `continue` ensures that the album's rating isn't modified. You prefer to set the rating to zero?

airsonic-main/src/main/java/org/airsonic/player/service/TranscodingService.java
```diff
@@ -328,10 +328,10 @@ private TranscodeInputStream createTranscodeInputStream(String command, Integer
             title = "Unknown Song";
         }
         if (album == null) {
-            title = "Unknown Album";
+            album = "Unknown Album";
```
eharris: I noticed these myself, good catch!

PR #934, opened 2019-03-28
Generate new passwords in a secure way

jvoisin:
Previously, lost passwords were generated via
org.apache.commons.lang.RandomStringUtils,
which is using java.util.Random internally.

This PRNG is has a 48-bit seed, that can easily be bruteforced
if an attacker is able to get the PRNG's output, for example
but resetting their own account multiple times,
leading to trivial privileges escalation attacks.

This commit makes use of java.security.SecureRandom
instead.


jvoisin:
This PR fixes `CVE-2019-10908`.

PR #936, opened 2019-03-28
Fix StringIndexOutOfBounds errors thrown by DWR/YUI

fxthomas:
This fixes the following errors in DWR caused by updating YUI from 2.3.6 to 2.4.8 in 86e58cea3a2f069e1e1fe1d1860bdc535d41d212:

```
2019-03-28 22:03:10.272  WARN --- o.d.servlet.JavaScriptHandler            : Compression system (YahooJSCompressor) failed to compress script
java.lang.StringIndexOutOfBoundsException: begin 919, end 1000, length 920
        at java.base/java.lang.String.checkBoundsBeginEnd(String.java:3107) ~[na:na]
        at java.base/java.lang.String.substring(String.java:1873) ~[na:na]
        ...
```

See more info in https://github.com/yui/yuicompressor/issues/161.

Tested on both self-contained JAR and Tomcat 8.

### DIFF
--- a/airsonic-main/pom.xml
+++ b/airsonic-main/pom.xml
@@ -232,6 +232,15 @@
             <artifactId>yuicompressor</artifactId>
             <version>2.4.8</version>
             <scope>runtime</scope>
+            <!-- Necessary to avoid StringIndexOutOfBoundsException when running as a JAR -->
+            <!-- See: https://github.com/yui/yuicompressor/issues/161 -->
+            <!-- See also: https://stackoverflow.com/questions/8429095 -->
+            <exclusions>
+              <exclusion>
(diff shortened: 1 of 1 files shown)


jvoisin:
This might sound naive, but do we really need yui in airsonic?

- I don't think that javascript downloads are the main bottleneck in a pile of java streaming audio files.
- Most of our javascript files are already compressed
- YUI [isn't maintained]( https://github.com/yui/yuicompressor/releases )

fxthomas:
It's there as a dependency for DWR, which is, unless I missed something, not replaceable easily.

jvoisin:
Damn :/

PR #940, opened 2019-03-28
Fix yet another batch of coverity-found issues

jvoisin:


airsonic-main/src/main/java/org/airsonic/player/controller/CoverArtController.java
```diff
@@ -684,6 +684,10 @@ public String toString() {
             this.height = height;
 
             int hash = key.hashCode();
+            // http://findbugs.sourceforge.net/bugDescriptions.html#RV_ABSOLUTE_VALUE_OF_HASHCODE
+            if (hash == Integer.MIN_VALUE) {
+              hash++;
+            }
             int rgb = COLORS[Math.abs(hash) % COLORS.length];
```
eharris: This could probably just be `int rgb = COLORS[(key.hashCode() & Integer.MAX_VALUE) % COLORS.length];`

airsonic-main/src/main/java/org/airsonic/player/domain/CacheElement.java
```diff
@@ -41,7 +41,12 @@ public CacheElement(int type, String key, Object value, long created) {
...
 
     public static long createId(int type, String key) {
-        return ((long) type << 32) | Math.abs(key.hashCode());
+        int hashcode = key.hashCode();
+        // http://findbugs.sourceforge.net/bugDescriptions.html#RV_ABSOLUTE_VALUE_OF_HASHCODE
+        if (hashcode == Integer.MIN_VALUE) {
+          hashcode++;
+        }
+        return ((long) type << 32) | Math.abs(hashcode);
     }
```
eharris: This could be `return ((long) type << 32) | (key.hashCode() & Integer.MAX_VALUE);`

PR #951, opened 2019-04-01
Use a random key to "encrypt" the remember-me cookie's value

jvoisin:
Since Spring's default remember-me technique is
terrible security-wise (`user:timstamp:md5(use:timestamp:password:key)`),
we should at least use a random key, instead of a fixed one,
otherwise, and attacker able to capture the cookies
might be able to trivially bruteforce offline
the password of the associated user.

My Java-fu is pretty weak, so feel free to tell me if there is a more idiomatic way of doing this.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/security/GlobalSecurityConfig.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/security/GlobalSecurityConfig.java
@@ -33,14 +33,6 @@ public class GlobalSecurityConfig extends GlobalAuthenticationConfigurerAdapter 
 
     static final String FAILURE_URL = "/login?error=1";
 
-    private static final String key;
-
-    static {
-      byte[] array = new byte[32];
-      new SecureRandom().nextBytes(array);
-      key = new String(array);
-    }
-
     @Autowired
     private SecurityService securityService;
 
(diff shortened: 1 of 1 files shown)


airsonic-main/src/main/java/org/airsonic/player/security/GlobalSecurityConfig.java
```diff
@@ -31,6 +33,14 @@
...
     static final String FAILURE_URL = "/login?error=1";
 
+    private static final String key;
+
+    static {
+      byte[] array = new byte[32];
+      new SecureRandom().nextBytes(array);
+      key = new String(array);
+    }
+
```
muff1nman: This isn't bad, but maybe a cleaner way to do this is to create a `public static String generateRememberMeKey() {...}` function that you then call once inline like `.and().rememberMe().key(generateRememberMeKey())`. This would allow us to move this logic around easier if needed.

jvoisin: The issue is that `generateRememberMeKey()` would return a different value each time, but I guess that since the `configure` methods are only called once, it should be ok.

fxthomas:
Rather annoyingly (esp. when running a development server) this requires to login each time you reload the server, even when "Remember me" is checked. Is there a better way, maybe store the random key for a given amount of time / pass it as a property on the command-line?

jvoisin:
I don't think that storing it on the filesystem is a good idea, I prefer to have an ephemeral one, invalidating all remember-me cookies on each restart.

What would make sense, is to have a `dev` mode in airsonic, to disable *annoying* security measures: using a fixed keys for the remember-me cookie, disabling CSP, …

muff1nman:
I think some users are going to be okay with the tradeoff of storing it on the filesytem. What about making this a property that can be optionally set? If its absent, generate it

jvoisin:
I don't really see a use-case for regular users to be honest.
But it wouldn't really harm to have a property anyway, so if anyone wants to implement it, it's fine by me.

PR #961, opened 2019-04-02
Remove WAP-related stuff

### LINKED ISSUE #960: Remove WAP
## Problem description

It was suggested by @jooola that we remove the WAP feature; @jvoisin agreed to take a look at it. The relevant code can be found [here](https://github.com/airsonic/airsonic/tree/f04ec61d0f5b41fb90517374ca8b1779bc0d8cb2/airsonic-main/src/main/webapp/WEB-INF/jsp/wap).

WAP was a standard that was used mainly pre-iPhone-1 to build web applications for mobile phones. The standard isn't used much anymore today<sup>[citation needed]</sup> and removing it would be in line with the practice of removing other features that were not used by many (for example the [chat](https://github.com/airsonic/airsonic/issues/195)).

jvoisin:
WAP (https://en.wikipedia.org/wiki/Wireless_Application_Protocol)
is a thing from the past that nobody is using anymore.

This should close #960


muff1nman:
Can you cleanup the resource bundles for messages as well?

Issue #868, opened 2019-01-14
Shares are weird (and broken)


heyarne:
## Problem description

I just realized that shares are broken in multiple ways:

1) The share page still needs flash.
2) ~~Shares need a login.~~
3) There is no API endpoint to receive a single share. There is one listing all shares of the current user,  but that obviously needs authentication.
4) You can only create Shares in the API with non-idv3-ids

(I stated earlier that shares need a login, you can see the edits. That's not true as you can login with the JWT that gets generated when creating a share with the current interface.)

I was planning to work on https://github.com/heyarne/airsonic-ui/issues/38 because one of my main motivations to start using airsonic was not only having music access myself but to be able to share it with my friends. Refreshing it with the new UI seemed like ticking off two things at once. As the API migration seems to take its time, is there any way this could be sorted out?

jooola:
Might be related to some similar older issues.

heyarne:
After building from current master (eea9416), which has @jvoisin's pull request #962 merged, I have made the following observations:

- In Chromium I get the complete playlist / album / playback queue that I shared; I can select different tracks in it and they play just fine
- In Firefox I get a single track (the first one)
- In both browsers I can't skip to a specific position in a track; the track is paused instead when I click the bar

Here's a screenshot in Chromium:
![2019-04-09-100305_1342x711_scrot](https://user-images.githubusercontent.com/525780/55783523-e4be5900-5aae-11e9-92a5-18d882d8873b.png)

Here's a screenshot in Firefox:
![image](https://user-images.githubusercontent.com/525780/55783311-5e097c00-5aae-11e9-9ae3-7edeb1eb17a6.png)

Additionally, shares created in DSub do not contain the JWT, which means they lack authentication and can not be passed on easily. This is expected though because the pull request did not touch any back end code.

As a further note: I tried debugging the JWT on https://jwt.io/#debugger and it was valid without any kind of secret, which makes it easily forgeable. Is there a way to define a secret for the JWT signature?

heyarne:
The rendering in Firefox is addressed in #982.

jvoisin:
I can freely seek in the shared music player.
Feel free to re-open this issue if it's not the case on your side :)

PR #962, opened 2019-04-03
Replace the flash player with medialement.js for the shared media

jvoisin:


jvoisin:
This is related to #868

jvoisin:
It's currently looking like this:

x

airsonic-main/src/main/webapp/WEB-INF/jsp/externalPlayer.jsp
```diff
@@ -4,68 +4,17 @@
 <html>
 <head>
     <%@ include file="head.jsp" %>
-    <script type="text/javascript" src="<c:url value="/script/swfobject.js"/>"></script>
     <script type="text/javascript" src="<c:url value="/script/prototype.js"/>"></script>
-
-    <meta name="og:type" content="album"/>
```
heyarne: Any reason for removing this particular meta tag?

jvoisin: Nope, I'll add it back.

jooola:
We forgot to clean the flash libraries.

jvoisin:
Be my guest ♥

almet:
Thanks @jvoisin :-)

PR #969, opened 2019-04-03
HTML5 player, UI tweaks

jasonjayr:
This makes a minor change to jvoisin's patch to make the mediaelement player's html display taller, and always open for playlist views.


jasonjayr:
I should also note, I'm using & testing with FF 67.0b3 (64-bit) on Linux

jvoisin:
I think that we should make the player wider, because with all the added buttons, it's hard to see/use the progress bar.

muff1nman:
Please rewrite after #982 or close if no longer needed.

Issue #428, opened 2017-05-21
Move every libresonic WebUI frame to HTML5


issuemover631:
<a href="https://github.com/jooola"><img src="https://avatars2.githubusercontent.com/u/19195485?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [jooola](https://github.com/jooola)**
_Sunday May 21, 2017 at 10:57 GMT_
_Originally opened as https://github.com/Libresonic/libresonic/issues/427_

----

Now that we have a PR with a HML5 player, I think it would be good if we move every HTML frames to HTML5. At least set the right DOCTYPE !

![image](https://cloud.githubusercontent.com/assets/19195485/26283256/98af50ee-3e24-11e7-9ed2-02ac058e5b83.png)

```
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN"
   "http://www.w3.org/TR/html4/loose.dtd">
```

I think this is a step to refactor libresonic to html5

jvoisin:
This was done in  #972.

PR #972, opened 2019-04-04
Declare the usage of HTML5 in the doctype

jvoisin:
This should marginally increase the performances of jquery (no more quirks mode), as well
as making web browsers happier.

A nice side-effect is that it'll make various html linters able to check if our pages are HTML5 compliant.


jvoisin:
This should close #428

PR #930, opened 2019-03-27
Bump jQuery and jQuery-ui versions

jvoisin:
I bumped it first to 2.X with jQuery migrate, played around but didn't manage
to trigger any warning.  So I bumped it again to the latest available version,
jQuery 3.4.0, which isn't triggering useful warnings either.

PR #973, opened 2019-04-04
[WIP] Bump the version of dependency-check-maven

jvoisin:
Updating dependency-check-maven has the nice side-effect of improving its detection capacities, yielding a few vulnerabilities that we should fix before merging this PR:

- [ ] Jquery and JqueryUI (#930)
- [ ] Liquibase-core
- [ ] spring-*
- [ ] posgresql
- [ ] apache-jsp

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -300,7 +300,7 @@
                 <plugin>
                     <groupId>org.owasp</groupId>
                     <artifactId>dependency-check-maven</artifactId>
-                    <version>4.0.0</version>
+                    <version>5.0.0-M2</version>
                     <inherited>true</inherited>
                     <configuration>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -300,7 +300,7 @@
                 <plugin>
                     <groupId>org.owasp</groupId>
                     <artifactId>dependency-check-maven</artifactId>
-                    <version>4.0.0</version>
+                    <version>5.0.0-M2</version>
```
jooola: v5 is out

jvoisin: I don't understand your comment: the point of this PR is precisely to update `dependency-check-maven` to version 5

jooola: Never mind I misunderstood something. I meant that m2 was not up to date.

Issue #955, opened 2019-04-01
Use SVG files for the icons


jvoisin:
It would be nice to use svg for icons/logos, instead of PNG. The [feather]( https://feathericons.com/ ) project provides such icons under MIT license.

I'll likely try to tackle this during a rainy Sunday.

PR #977, opened 2019-04-04
Use svg for the icons

jvoisin:
## Before:
rast

## After:
vect png

This has several advantages:
1. The images are way more lightweight. But since we're in 2019, it doesn't matter than much anymore, especially given the fact that airsonic is about streaming media files
2. Even when zoomed, the images are still looking cool.
3. All the .svg icons are from [feather]( https://github.com/feathericons/feather ), released under MIT license. This ensures that they all go together pretty well.

If there is interest in this, I'll be happy to convert the rest of the icons.
Also, should we want to, I *think* that it might be possible to change the color of the icons via css.


jvoisin:
an

jvoisin:
aaa

jooola:
Something seems weird using 3 times the same icon. Menu, playlists, list view, do you think this could be arranged ? Maybe if we cotribute to feather to add some new required icons ?

Also player is better floating on the left, instead of middle.

Also is are add next and the append to playlist replaced ? Doesn't seems so ? Same for go up button on the left pane ?

Logout icon would definitly be better, might need some tweaking with the user's avatar, if present.

jvoisin:
meh

I'm not a big fan of having the sidebar icon on the same line than the menu, it looks a bit odd.

jvoisin:
icon

Now with a logout icon!

jvoisin:
This should close #955

airsonic-main/src/main/resources/org/airsonic/player/theme/default_light.properties
```diff
@@ -1,57 +1,61 @@
...
+downloadImage = icons/default_light/download.svg
+forwardImage = icons/default_light/forward.svg
+helpImage = icons/default_light/help.svg
+helpPopupImage = icons/default_light/help.svg
+homeImage = icons/default_light/home.svg
 html5Image = icons/default_light/html5.png
 keyboardImage = icons/default_light/keyboard.png
-logImage = icons/default_light/log.png
+logImage = icons/default_light/log.svg
 logoImage = icons/default_light/logo.png
```
jooola: You can find the logo in svg if you like, check the contrib folder.

airsonic-main/src/main/resources/org/airsonic/player/theme/default_light.properties
```diff
@@ -1,57 +1,61 @@
...
+currentImage = icons/default_light/current.svg
+donateImage = icons/default_light/donate.svg
+donateSmallImage = icons/default_light/donate.svg
+downImage = icons/default_light/down.svg
+downloadImage = icons/default_light/download.svg
+forwardImage = icons/default_light/forward.svg
+helpImage = icons/default_light/help.svg
+helpPopupImage = icons/default_light/help.svg
+homeImage = icons/default_light/home.svg
 html5Image = icons/default_light/html5.png
```
jooola: Do we really display a html5 logo somewhere ? Couldn't we remove it ? I think any logo such as html5 css3 or else shouldn't be in the software, and if we really want to make some fancy "used technologies" page, lets do it on the website.

jooola: I would even remove the ugly apps image from the more page. This could also be moved to the website (already is btw). We discussed this in the past I think, but haven't been changed.

jooola:
Ok here are the cosmetics coming !

- In the albums view (home), the shuffle icon is too big
image

- Everywhere, the more action, start radio action, and maybe more, have a CSS class that prepend the `forward.png` image. I would change this image to be consistent with the rest of the icons, and make it the same size as the shuffle icon (as said above).

- In the player, I don't know if this will be changed in the "rework the control bar" PR, but it would be nice to update the play and volume icon with the feather ones, again to have more consistency. Also the volume icon is ugly.

- The ratingOn icon is still using PNG, I would use the ratingOff icon mixed with a css class that changes the color of the icon. Should be doable using svg files and the fill property. Or we could copy the ratingOff svg icon and change its color by hand.

- In the playqueue, the current icon is black, but it should be the same color as the other icons (grey in our case). Maybe we should add a css class on every svg icon and allow to change the filling color from the theme css file.

- some icon attached to some textes (share for example) are not verticaly aligned with the text. Could be fixed with some Css.

airsonic-main/src/main/resources/org/airsonic/player/theme/default_light.properties
jvoisin: I wanted to simply replace the icons with svg ones. I'll issues a cleanup PR once this is done, instead of piling things up and complicate the review :)

jvoisin:
- I fixed the size of the shuffle icon
- I changed the forward to a svg one, properly aligned (wasn't fun)
- The control bar is closely tied to mediaelements.js, I don't want to touch this for now
- I added a svg for the ratingOn icon
- I fixed the colour of the "current" marker
- I tried to fix some alignment

fxthomas:
Can you also convert icons for the dark themes (I'm using `groove`, and it currently shows an error)?

jvoisin:
I would like to have this merged before adding even more things to it.
What kind of errors are you seeing?

fxthomas:
The top bar and play queue are not rendered at all ; the 500'd when I tried, with a message about a key not being found. I assume that's because only `default_light.properties` was changed in your PR?

jvoisin:
You're right. It's now fixed :)

fxthomas:
I'm still getting `Theme 'groove': No message found under code 'repeatOn' for locale 'en'.` in `playQueue.view`...

jvoisin:
I only tested with the default_clear midnight themes, meh.

This time, it should™ be fixed!

fxthomas:
Okay, this works just like before, aside from a few changes ;)

The only thing I would remove is the background image for `.forward` (it's defined for `.forward::before` in your PR).

jvoisin:
Why do you want to get it removed?
This is how some stuff like `more` are getting their prefixing `»`

fxthomas:
It's defined twice (once in the default CSS and once more in `groove.css` (and in a lot of other themes).

Screenshot from 2019-06-12 22-42-48

jvoisin:
Sigh. I guess I'll revert this one back to png, so that we can get this PR merged.

fxthomas:
...or we can remove the `.forward` for all themes since it's there in the
default CSS?

jvoisin:
oh.
the default css is included.

sigh, I need some sleep…

jvoisin:
Actually, we can't since some themes are using different forward/backward buttons, that might cause alignment issues, and I don't want to check all 31 themes one by one :/

jvoisin:
Like this?

eharris:
I really like the look of the new svg icons, with the exception of the Playing/headphones one, it's just too minimal to be easily identifiable.  Can you tweak it to be more recognizable?  For other options I also looked at the stock speaker.svg and don't like it either.  If we want a stock one, maybe music.svg would be an acceptable option?

fxthomas:
Yup, looks more or less okay now (once #1112 is fixed this should work well), and haven't had issues with it for a few days now:

Screenshot from 2019-06-19 23-53-17

airsonic-main/src/main/resources/org/airsonic/player/theme/default_light.properties
jvoisin: I had some alignment issues, I'll send a separate PR for this.

Issue #976, opened 2019-04-04
Airsonic won't start as a systemd service (insufficient memory error)


Millichrome:
## Problem description
I can't get Airsonic to run as a system service as per the documentation. If I run systemctl status after starting the service I get a "failed" notice and an error about insufficient memory for Java. I am running a Google Cloud small instance.

I believe I can get airsonic running if I use the java command. The application starts but I haven't been able to access the admin interface (this could be an issue with Google Cloud's firewall or some other setting that I don't know about).   

### Steps to reproduce
1. Follow instruction for standalone install on https://airsonic.github.io/docs/install/war-standalone/
2. Modify docs for Ubuntu (/etc/default) and uncomment /etc/sysconfig/airsonic
3. Run `systemctl start airsonic.service`
4. Run `systemctl status airsonic.service`
5. This what I get:
~~~
OpenJDK 64-Bit Server VM warning: INFO: os::commit_memory(0x00007f266d000000, 2555904, 1) failed; error='Operation not permitted' (errno=1)
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (mmap) failed to map 2555904 bytes for committing reserved memory
# An error report file with more information is saved as:
# /tmp/hs_err_pid1498.log
airsonic.service: Main process exited, code=exited, status=1/FAILURE
airsonic.service: Failed with result 'exit-code'.
~~~

I cannot access that log file. It doesn't seem to exist.

## System information
* Airsonic version: v10.2.1
* OS: Ubuntu 18.04 LTS
* No Proxy server (for now)
* Client: Latest Firefox 
* RAM: 1.7 GB (Google Cloud small instance)
* Nothing else installed on the system

I am not a Linux pro. I am just trying to get a self-hosted music server running in the cloud because GPM/Spotify/iTunes doesn't work for me.

jvoisin:
I think that the JVM is lying, because the error is `Permission denied`. Can you take a look at the kernel's logs (likely via `dmesg`) to see if there is something permissions-related?

Millichrome:
Below is the output from dmesg after running systemctl start/status:

https://pastebin.com/PkD6TKpD

jvoisin:
I'm not seeing anything suspicious in it. Nothing in `/var/log/syslog`
either I guess?

Millichrome:
Wow, that was a long log. The only references to airsonic/JVM contain the same info I posted in the OP.

[...]

PR #980, opened 2019-04-05
Fix the systemd unit file for OpenJDK

jvoisin:
Apparently, the OpenJDK isn't able to detect that W^X is enforced on the system, and will miserably fail instead of falling back to non-JIT (let alone implementing a W^X-compatible JIT).


jvoisin:
This is related to #976.

PR #982, opened 2019-04-06
Fix formatting on external player for firefox

muff1nman:


muff1nman:
@jvoisin I think this is a better approach than the min-height attribute for firefox.

jvoisin:
Flexbox, I like this!

heyarne:
FWIW, I'm getting funky syntax highlighting on all the non-escaped URLs. It might be a good idea escaping them as well, even though they probably won't contain any characters that break the link.

PR #983, opened 2019-04-07
Replace latin encoding with utf-8

jvoisin:


jooola:
Curious how the translation are now handled.

PR #987, opened 2019-04-08
Fix a NULL-deref in CoverArtController.java

jvoisin:
ImageIO.read() can (and will) return null in certain cases.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/CoverArtController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/CoverArtController.java
@@ -381,7 +381,10 @@ public class CoverArtController implements LastModified {
                 InputStream in = null;
                 try {
                     in = getImageInputStream(coverArt);
-                    return scale(ImageIO.read(in), size, size);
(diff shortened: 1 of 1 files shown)


airsonic-main/src/main/java/org/airsonic/player/controller/CoverArtController.java
```diff
@@ -381,7 +381,10 @@ public BufferedImage createImage(int size) {
                 InputStream in = null;
                 try {
                     in = getImageInputStream(coverArt);
-                    return scale(ImageIO.read(in), size, size);
+                    BufferedImage bimg = ImageIO.read(in);
+                    if (bimg != null) {
+                      return scale(ImageIO.read(in), size, size);
```
fxthomas: We should probably reuse `bimg` there, right?

eharris:
Turns out this doesn't always fix the whole issue.  `ImageIO.read()` will also throw an NPE if it is passed a null, which I just saw in my build that includes this change.

I'm wondering if it wouldn't be better to back this out and instead just catch NPE's for the whole block?

eharris:
I just added #1241 to improve this.

PR #993, opened 2019-04-08
[WIP] Add a preference to disable all external services

jvoisin:
This preference disables:

- [x] ChromeCast integration
- [ ] Lastfm cover fetching
- [ ] Lyrics via chartlyrics.com

Anything else?

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/PlayQueueController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/PlayQueueController.java
@@ -68,6 +68,8 @@ public class PlayQueueController {
         map.put("partyMode", userSettings.isPartyModeEnabled());
         map.put("notify", userSettings.isSongNotificationEnabled());
         map.put("autoHide", userSettings.isAutoHidePlayQueue());
+        map.put("autoHide", userSettings.isAutoHidePlayQueue());
+        map.put("externalServicesEnabled", settingsService.isExternalServicesEnabled());
         return new ModelAndView("playQueue","model",map);
     }
 }
--- a/airsonic-main/src/main/java/org/airsonic/player/service/SettingsService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/SettingsService.java
@@ -95,6 +95,7 @@ public class SettingsService {
     private static final String KEY_LDAP_SEARCH_FILTER = "LdapSearchFilter";
     private static final String KEY_LDAP_AUTO_SHADOWING = "LdapAutoShadowing";
     private static final String KEY_GETTING_STARTED_ENABLED = "GettingStartedEnabled";
+    private static final String KEY_EXTERNAL_SERVICES_ENABLED = "ExternalServicesEnabled";
     private static final String KEY_SETTINGS_CHANGED = "SettingsChanged";
     private static final String KEY_LAST_SCANNED = "LastScanned";
     private static final String KEY_ORGANIZE_BY_FOLDER_STRUCTURE = "OrganizeByFolderStructure";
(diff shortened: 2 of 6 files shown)


jvoisin:
This is related to #712

airsonic-main/src/main/java/org/airsonic/player/controller/PlayQueueController.java
```diff
@@ -68,6 +68,8 @@ protected ModelAndView handleRequestInternal(HttpServletRequest request, HttpSer
         map.put("partyMode", userSettings.isPartyModeEnabled());
         map.put("notify", userSettings.isSongNotificationEnabled());
         map.put("autoHide", userSettings.isAutoHidePlayQueue());
+        map.put("autoHide", userSettings.isAutoHidePlayQueue());
```
jooola: Duplicate line here no ?

jooola:
What about fine grained service disabling ? One checkbox par service ?
Are you only disabling front-end services or back-end too ? Because it will only mitigate #712 , since top tracks and radios are using last.fm in the back-end.

jvoisin:
I plan to disable front-end and back-end.

eharris:
The airsonic version check is also an external lookup.  A global disable (which also hides that option from users) would be nice.

eharris:
@jvoisin I'd be happy to help with this PR if you'd like.

jvoisin:
Be my guest :)

eharris:
So in looking at this more, why wouldn't you just add these new options to the existing `Advanced` Settings page rather than creating a new page?  Since the Captcha is already there, seems logical for those to be there too.

eharris:
I rebased this on current master, fixed the merge conflict, fixed the checkstyle violations, finished the cleanup backing out the new settings page, and squashed all the commits down to one so it's easier to review.  So fair warning, I've rewritten history!

airsonic-main/src/main/java/org/airsonic/player/service/SettingsService.java
```diff
@@ -180,6 +185,9 @@
     private static final String DEFAULT_LDAP_SEARCH_FILTER = "(sAMAccountName={0})";
     private static final boolean DEFAULT_LDAP_AUTO_SHADOWING = false;
     private static final boolean DEFAULT_GETTING_STARTED_ENABLED = true;
+    private static final boolean DEFAULT_EXTERNAL_SERVICES_CHROMECAST_ENABLED = false;
+    private static final boolean DEFAULT_EXTERNAL_SERVICES_LASTFM_ENABLED = false;
+    private static final boolean DEFAULT_EXTERNAL_SERVICES_CHARTLYRICS_ENABLED = false;
```
muff1nman: Let's not get too ambitious and keep the default behavior the same for now.

muff1nman:
Maybe @eharris  and @jvoisin can collapse this and #1281 into one?

airsonic-main/src/main/java/org/airsonic/player/service/SettingsService.java
```diff
@@ -723,6 +731,30 @@ public void setGettingStartedEnabled(boolean isGettingStartedEnabled) {
         setBoolean(KEY_GETTING_STARTED_ENABLED, isGettingStartedEnabled);
     }
 
+    public boolean isExternalServicesChromecastEnabled() {
```
almet: I'm quite new to Java, so maybe it's idiomatic, but couldn't we add only one method to check for these settings, passing the name of the service as the first argument? That would allow for less code duplication, no?

muff1nman: Yes that would allow for less code duplication, however in this case the duplication isn't terrible as it keeps everything consistent across the whole class. By keeping things consistent it makes it easier to address the code duplication at a larger scale in the future (i.e the whole class).

almet: I understand! (Is there an already open issue to fix this code duplication at some point, or should we open one?)

About this issue, should we move ahead and merge this? If not, what's missing?

muff1nman: I don't think there is an open issue for it, but its pretty low priority in my mind. So low I wouldn't even bother opening an issue for it.

This PR is still marked as WIP which means the author doesn't think its ready to get merged yet.

PR #996, opened 2019-04-09
Compatible with ID3 v2.4

tesshucom:

So, if it is v2.4 form, it works well in the form read additionally.
___

Sorry.
Previous PR did not take into account the relevance of artist, songArtist, albumArtist.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -9,10 +9,13 @@ daysUntilClose: 180
 
 # Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
 exemptLabels:
-  - enhancement
-  - accepted
-  - critical
-  - neverstale
+  - 'enhancement'
+  - 'type: enhancement'
+  - 'accepted'
+  - 'status: accepted'
+  - 'critical'
+  - 'priority: critical'
+  - 'neverstale'
 
 # Set to true to ignore issues in a project (defaults to false)
 exemptProjects: false


jvoisin:
> It can be avoided by running this PR code only if fields can not be read
I think that this is a good idea.

As for the weird values, it looks like a bug in jaudiotagger, doesn't it?

tesshucom:
It may be a bug because it returns  a (should not be readable)value.

When I searched for my data, I found an old mp3 with the same symptoms.
It has the format "ID3v1 ID3v2.3".

Rewrite the source for which overwriting of dirty data is avoided so that existing data is not affected.

tesshucom:
Improvement as below.

image

image

image

However, if the above code is executed for the ID3v1 file, an error will be displayed in the log only once.
(It does not seem to be per file.)

```
java.lang.UnsupportedOperationException: TODO:Not done yet
        at org.jaudiotagger.tag.id3.ID3v1Tag.getFields(ID3v1Tag.java:603) ~[jaudiotagger-2.2.5.jar!/:na]
        at org.airsonic.player.service.metadata.JaudiotaggerParser.getRawMetaData(JaudiotaggerParser.java:102) ~[classes!/:10.3.0-SNAPSHOT]
        at org.airsonic.player.service.metadata.MetaDataParser.getMetaData(MetaDataParser.java:47) [classes!/:10.3.0-SNAPSHOT]
        at org.airsonic.player.service.MediaFileService.createMediaFile(MediaFileService.java:488) [classes!/:10.3.0-SNAPSHOT]
        at org.airsonic.player.service.MediaFileService.updateChildren(MediaFileService.java:400) [classes!/:10.3.0-SNAPSHOT]
        at org.airsonic.player.service.MediaFileService.getChildrenOf(MediaFileService.java:190) [classes!/:10.3.0-SNAPSHOT]
        at org.airsonic.player.service.MediaScannerService.scanFile(MediaScannerService.java:249) [classes!/:10.3.0-SNAPSHOT]
        at org.airsonic.player.service.MediaScannerService.doScanLibrary(MediaScannerService.java:189) [classes!/:10.3.0-SNAPSHOT]
        at org.airsonic.player.service.MediaScannerService.access$000(MediaScannerService.java:45) [classes!/:10.3.0-SNAPSHOT]
        at org.airsonic.player.service.MediaScannerService$2.run(MediaScannerService.java:157) [classes!/:10.3.0-SNAPSHOT]

```
It does not occur with recent files.

PR #999, opened 2019-04-10
Compatible with ID3 v2.4

tesshucom:
(PR # 996 again because the commit history gets dirty)

So, if it is v2.4 form, it works well in the form read additionally.
(The additional processing type may be preferable to handle various formats from ID3v1 to ID3v24 transients.)
___

### Version check
For example Improper combination, when trying to get Fields of ID3v24 Tag for tag of file of ID3v1, error.

```
java.lang.UnsupportedOperationException: TODO:Not done yet
        at org.jaudiotagger.tag.id3.ID3v1Tag.getFields(ID3v1Tag.java:603) ~[jaudiotagger-2.2.5.jar!/:na]
        at org.airsonic.player.service.metadata.JaudiotaggerParser.getRawMetaData(JaudiotaggerParser.java:102) ~[classes!/:10.3.0-SNAPSHOT]
        at org.airsonic.player.service.metadata.MetaDataParser.getMetaData(MetaDataParser.java:47) [classes!/:10.3.0-SNAPSHOT]
        at org.airsonic.player.service.MediaFileService.createMediaFile(MediaFileService.java:488) [classes!/:10.3.0-SNAPSHOT]
        at org.airsonic.player.service.MediaFileService.updateChildren(MediaFileService.java:400) [classes!/:10.3.0-SNAPSHOT]
        at org.airsonic.player.service.MediaFileService.getChildrenOf(MediaFileService.java:190) [classes!/:10.3.0-SNAPSHOT]
        at org.airsonic.player.service.MediaScannerService.scanFile(MediaScannerService.java:249) [classes!/:10.3.0-SNAPSHOT]
        at org.airsonic.player.service.MediaScannerService.doScanLibrary(MediaScannerService.java:189) [classes!/:10.3.0-SNAPSHOT]
        at org.airsonic.player.service.MediaScannerService.access$000(MediaScannerService.java:45) [classes!/:10.3.0-SNAPSHOT]
        at org.airsonic.player.service.MediaScannerService$2.run(MediaScannerService.java:157) [classes!/:10.3.0-SNAPSHOT]

```

Therefore, version check is performed.

When loading, a read class is selected in Strategy.
Version check is enabled by cast of Tag.

### ifnull check
Avoiding dirty data done at # 996.
It may be considered as a poor implementation of jaudiotagger, but it may also occur if the format is hybrid to the ID3 specification.
Redundant but necessary to avoid this.

The data in the repository contains data applicable to this case.

image

image

image

Improvement as below by ifnull.

image

image

image

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -9,10 +9,13 @@ daysUntilClose: 180
 
 # Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
 exemptLabels:
-  - enhancement
-  - accepted
-  - critical
-  - neverstale
+  - 'enhancement'
+  - 'type: enhancement'
+  - 'accepted'
+  - 'status: accepted'
+  - 'critical'
+  - 'priority: critical'
+  - 'neverstale'
 
 # Set to true to ignore issues in a project (defaults to false)
 exemptProjects: false


jooola:
> (PR # 996 again because the commit history gets dirty)

Please use the git squash/rebase methods to clean the commit history, this will prevent from opening more and more PRs

PR #1000, opened 2019-04-10
Compatible with ID3 v2.4

tesshucom:
(PR #996 again because the commit history gets dirty)

If it is v2.4 form, it works well in the form read additionally.
(The additional processing type may be preferable to handle various formats from ID3v1 to ID3v24 transients.)
___

#### Version check
The processing added in this PR is executed additionally only for the ID3v24 file.

#### ifnull check
The processing added to this PR will not be overwritten with the value acquired from the ID3v24 field if the tag value of ID3v24 or earlier has already been read.

Avoiding dirty data done at # 996.
It may be considered as a poor implementation of jaudiotagger, but it may also occur if the format is hybrid to multiple ID3 specifications.

The data in the repository contains data applicable to this case.

image

image

image

Improvement as below by ifnull.

image

image

image

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
@@ -97,8 +97,8 @@ public class JaudiotaggerParser extends MetaDataParser {
                 metaData.setTrackNumber(parseTrackNumber(getTagField(tag, FieldKey.TRACK)));
                 metaData.setMusicBrainzReleaseId(getTagField(tag, FieldKey.MUSICBRAINZ_RELEASEID));
 
-                String songArtist = getTagField(tag, FieldKey.ARTIST);
-                String albumArtist = getTagField(tag, FieldKey.ALBUM_ARTIST);
+                metaData.setArtist(getTagField(tag, FieldKey.ARTIST));
+                metaData.setAlbumArtist(getTagField(tag, FieldKey.ALBUM_ARTIST));
 
                 if (tag instanceof AbstractID3Tag && 0 < audioFile.getTag().getFieldCount()) {
 
@@ -140,7 +140,7 @@ public class JaudiotaggerParser extends MetaDataParser {
                                     }
                                     break;
                                 case ID3v24Frames.FRAME_ID_ARTIST:
-                                    if (null == metaData.getArtist() && null == songArtist) {
+                                    if (null == metaData.getArtist()) {
                                         metaData.setArtist(f.toString());
                                     }
                                     break;
@@ -155,11 +155,11 @@ public class JaudiotaggerParser extends MetaDataParser {
                         });
                     }
                 }
-                if(StringUtils.isBlank(metaData.getArtist())) {
-                    metaData.setArtist(StringUtils.isBlank(songArtist) ? albumArtist : songArtist);
+                if (StringUtils.isBlank(metaData.getArtist())) {
+                    metaData.setArtist(metaData.getAlbumArtist());
                 }
-                if(StringUtils.isBlank(metaData.getAlbumArtist())) {
-                    metaData.setAlbumArtist(StringUtils.isBlank(albumArtist) ? songArtist : albumArtist);
+                if (StringUtils.isBlank(metaData.getAlbumArtist())) {
+                    metaData.setAlbumArtist(metaData.getArtist());
                 }
 
             }
--- a/airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
@@ -37,6 +37,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
+import org.springframework.util.ObjectUtils;
 
 import java.io.File;
 import java.util.SortedSet;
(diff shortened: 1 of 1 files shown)


jvoisin:
Protip: you can [rewrite your commit history]( https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History ), and then force-push, instead of opening several pull-requests :)

airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
```diff
@@ -96,8 +99,69 @@ public MetaData getRawMetaData(File file) {
...
+                                        metaData.setAlbumArtist(f.toString());
+                                    }
+                                    break;
+                                default:
+                                    break;
+                            }
+                        });
+                    }
+                }
+                if(StringUtils.isBlank(metaData.getArtist())) {
```
jvoisin: Shouldn't this be in an `else`?

tesshucom: > Protip: you can [rewrite your commit history](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History), and then force-push, instead of opening several pull-requests :)

 Sorry... You're absolutely right.

> Shouldn't this be in an `else`?

Generally it is very rare, but there may be cases where Artist does not exist in either id3v2.1 or id3v2.4 field.
However certainly, it may be that there is not enough consideration about the point indicated.

When the relationship between Artist and AlbumArtist is as follows.
(First of all, it is necessary to confirm this specification.)

image
Case 3 may now have a bad implementation.

jooola: Could you add links to the specifications in the code that back/approves your changes ?

airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
```diff
@@ -97,8 +97,8 @@ public MetaData getRawMetaData(File file) {
                 metaData.setTrackNumber(parseTrackNumber(getTagField(tag, FieldKey.TRACK)));
                 metaData.setMusicBrainzReleaseId(getTagField(tag, FieldKey.MUSICBRAINZ_RELEASEID));
 
-                String songArtist = getTagField(tag, FieldKey.ARTIST);
-                String albumArtist = getTagField(tag, FieldKey.ALBUM_ARTIST);
+                metaData.setArtist(getTagField(tag, FieldKey.ARTIST));
+                metaData.setAlbumArtist(getTagField(tag, FieldKey.ALBUM_ARTIST));
```
tesshucom: Set values ​​for artist and albumArtist (except id3v2.4).

airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
```diff
@@ -140,7 +140,7 @@ public MetaData getRawMetaData(File file) {
                                     }
                                     break;
                                 case ID3v24Frames.FRAME_ID_ARTIST:
-                                    if (null == metaData.getArtist() && null == songArtist) {
+                                    if (null == metaData.getArtist()) {
```
tesshucom: Remove unnecessary judgments.

airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
```diff
@@ -155,11 +155,11 @@ public MetaData getRawMetaData(File file) {
                         });
                     }
                 }
-                if(StringUtils.isBlank(metaData.getArtist())) {
-                    metaData.setArtist(StringUtils.isBlank(songArtist) ? albumArtist : songArtist);
+                if (StringUtils.isBlank(metaData.getArtist())) {
+                    metaData.setArtist(metaData.getAlbumArtist());
```
tesshucom: Use albumArtist instead of artist if artist does not resolve.

airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
```diff
@@ -155,11 +155,11 @@ public MetaData getRawMetaData(File file) {
...
                 }
-                if(StringUtils.isBlank(metaData.getAlbumArtist())) {
-                    metaData.setAlbumArtist(StringUtils.isBlank(albumArtist) ? songArtist : albumArtist);
+                if (StringUtils.isBlank(metaData.getAlbumArtist())) {
+                    metaData.setAlbumArtist(metaData.getArtist());
```
tesshucom: Use artist instead of albumArtist if albumArtist does not resolve.

airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
tesshucom: Certainly.

airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
tesshucom: I agree with that opinion!
However, I did this because the method of writing in this class is not uniform.
 - Rewrite the place related to this commit with StringUtils
 - Rewrite the null check of this class with StringUtils

Which one is better ...?

muff1nman: Not sure I understand what option 2 is, but in general I would say that new code should use StringUtils.isBlank when it makes sense.

tesshucom: A correction has been made.

https://github.com/airsonic/airsonic/blob/1c4a70af99229091c258033f9640e4bc27815ce4/airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java#L112-L155

PR #1008, opened 2019-04-12
Fix a bunch of possible XSS

jvoisin:
I didn't bother to try to exploit them,
but those variables were suspiciously looking like
XSS vectors. Anyway, it shouldn't hurt to escape them.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/HomeController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/HomeController.java
@@ -114,6 +114,7 @@ public class HomeController  {
                 map.put("genres", genres);
                 if (!genres.isEmpty()) {
                     String genre = getStringParameter(request, "genre", genres.get(0).getName());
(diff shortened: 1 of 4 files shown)


airsonic-main/src/main/java/org/airsonic/player/controller/HomeController.java
```diff
@@ -114,6 +114,7 @@ protected ModelAndView handleRequestInternal(HttpServletRequest request) throws
                 map.put("genres", genres);
                 if (!genres.isEmpty()) {
                     String genre = getStringParameter(request, "genre", genres.get(0).getName());
+                    genre = genre.replaceAll("[^\\p{IsAlphabetic}^\\p{IsDigit}]", "");
```
fxthomas: This breaks things for some genres (e.g. `Heavy Metal` has a space in it, likely others).

jvoisin: You're right, I should use proper escaping instead of a ghetto-regex.

PR #1015, opened 2019-04-13
Fix a stacktrace in AvatarController

jvoisin:
The parameter `id` was parsed as a string, then
casted to an int. We're now directly parsing it as an
Integer instead, gracefully handling errors as a side-effect.


jvoisin:
Any objection against merging this?

eharris:
Can you rebase this from master?  I believe that will show some checkstyle violations.

eharris:
There's a compilation error, and still checkstyle violations

Issue #911, opened 2019-03-21
Fancyzoom is non-free


jvoisin:
I was looking at [FancyZoom.js]( https://github.com/airsonic/airsonic/blob/8d3c0ec9a0e126e0e430de7c4fbf817199632b0d/airsonic-main/src/main/webapp/script/fancyzoom/FancyZoom.js ) to see how to replace its use of `eval` (see #909 for details),
and it turns out that it's not released under a Free (as in Freedom) license:

```
// Redistribution and use of this effect in source form, with or without modification,
// are permitted provided that the following conditions are met:
//
// * USE OF SOURCE ON COMMERCIAL (FOR-PROFIT) WEBSITE REQUIRES ONE-TIME LICENSE FEE PER DOMAIN.
//   Reasonably priced! Visit www.fancyzoom.com for licensing instructions. Thanks!
//
// * Non-commercial (personal) website use is permitted without license/payment!
//
// * Redistribution of source code must retain the above copyright notice,
//   this list of conditions and the following disclaimer.
//
// * Redistribution of source code and derived works cannot be sold without specific
//   written prior permission.
//
// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
// "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
// LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR
// CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
// EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
// PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
// PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
// LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
// NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
[...]

PR #1017, opened 2019-04-15
Replace FancyZoom with jquery.fancyzoom

jvoisin:
FancyZoom isn't a free-as-in-freedom software,
and is mostly a big pile of old-school javascript.
This commit replace it with a small jquery-powered
script under MIT license.


jvoisin:
This should close #911

jvoisin:
Rebased and tested locally :)

Issue #1018, opened 2019-04-15
Replace wz_tooltip.js with jquery-ui


jvoisin:
Since we're already using jquery-ui.js for stuff™, it might make sense to replace the 1300-lines-long-old-school-steaming-pile-of-javascript-wz_tooltip.js with it.

jvoisin:
Done in 195a0879

muff1nman:
Reverted #1019 due to #1112

eharris:
The current incarnation of this is in PR #1154.

jvoisin:
Done in 584ee96f

Issue #1001, opened 2019-04-10
Document the localization system


jvoisin:
Document the following:

- [ ] How to contribute to localizations
- [ ] How to add new strings
  - [ ] What files to change
- [ ] How to test the localizations

jooola:
Translation are handled by Transifex. So only the source file (english) needs to be updated.

I'll take care updating the transifex references and pull any new translation via a PR so we can review the changes.

https://www.transifex.com/airsonic/airsonic/

jvoisin:
Can you please add a some details about this in #942 ?

jooola:
>  How to contribute to localizations

Translate strings at Transifex https://www.transifex.com/airsonic/airsonic/ requesting translator permissions to project owner

> How to add new strings 

Edit the i18n source file and add the key + string 
https://github.com/airsonic/airsonic/blob/master/airsonic-main/src/main/resources/org/airsonic/player/i18n/ResourceBundle_en.properties

On my side I'll use the transifex-client (python package) to push to source file and update the targeted files

eharris:
I believe that 542256285fe5e9ba18b3bd080f18d69250d23bff will accomplish this (once it is merged).

prschopf:
Link https://www.transifex.com/airsonic/airsonic/ is not working.

jooola:
@prschopf We moved to weblate. 
https://hosted.weblate.org/projects/airsonic/airsonic/

But currently the translation are locked, while I really have to get the integration done. I cannot get to work properly.

comradekingu:
@jooola What is wrong with it? If you add https://hosted.weblate.org/user/kingu/ ad admin in https://hosted.weblate.org/projects/airsonic/airsonic/ I can have a look.

PR #1019, opened 2019-04-15
Replace wz_tooltip.js with jquery-ui

jvoisin:
Since we're already using jquery-ui for other stuff, it makes sense to use it
for tooltips too, instead of using an old-school-written-in-2009 1300 lines
  steaming pile of javascript named wz_tooltip.js


jvoisin:
This should close #1018.

jvoisin:
This is currently blocked by #1001, because I have no clue on how to update the translations/texts, and I need to remove the `<p>…</p>` characters from the tooltips texts.

muff1nman:
Is this one still blocked?

jvoisin:
Yup, I need to find some time to edit the localizations :/

jvoisin:
If anyone know some sed/awk, the question is: How to replace `helppopup.YYY.text=<p>XXX</p>` with `helppopup.YYY.text=XXX`. If nobody has a quick one-liner to drop, I'll allocate some time to write one at some point.

fxthomas:
Something like (modifies files in-place):

```
find airsonic-main/src/main/resources/ -name 'ResourceBundle_*.properties' -print0 | xargs -0 sed -i 's/\(helppopup\..*\.text\)\s*=\s*<p>\(.*\)<\/p>/\1=\2/'
```

jvoisin:
This is now ready to be reviewed'n'merged :)

PR #1026, opened 2019-04-19
Test stream for m4a vbr

muff1nman:


muff1nman:
This PR has two additional largish files (one m4a and one flac) to use during tests. I'd rather not commit these files to the git history to keep our repo size down. What I'd like to do is put these on an internet facing place so that we can fetch the samples down during integration/e2e testing. Looking around, vlc does something similar for their testing: https://github.com/videolan/vlc/blob/70ee5693df6cf6e6676e8bf38b3376e03a4ee933/test/Makefile.am#L75

Some ideas I've considered are:

1. Use git lfs
2. Serve them up on s3 (use my personal account?)
3. Distribute them via bittorrent or ipfs
4. Try to host them on github (the docs page?)
5. Commit them to history (i.e. PR as is)

All of these options have their upsides/downsides. Anyone have any opinions?

muff1nman:
This PR requires #1035. Expected response files were generated with ffmpeg.

jooola:
What about creating a dummy file (3 sec audio) that weight almost nothing ? Creating 1 dummy file in FLAC and trans-coding it to other format could do the trick, or use existing files like these :
http://techslides.com/sample-files-for-development

muff1nman:
@jooola I originally used the longer files because they more clearly exhibit issues with length mismatches and were causing people the most problems.

muff1nman:
In addition, I'd like to start doing some tests with videos and in order to address issues like #1053, it'll require larger files I think.

tari:
I'd prefer git lfs; storing large binary files is exactly what it's meant for, and that overlaps nicely with 4 because they're still stored on github.

I'm not sure how that interacts with on-disk repository size when checked out though.

jvoisin:
I vote for [git-lfs]( https://git-lfs.github.com/ ).

heyarne:
:+1: for git-lfs

PR #1027, opened 2019-04-22
Avoid logging sensitive URL parameters in the Subsonic API

fxthomas:
In case of exceptions, Airsonic logs the full URL that triggered it since 417583cc, including possibly sensitive query parameters such as the authentication string passed to the Subsonic API. When outputting logs, this replaces the value set for this parameter in the URL by the "\<hidden\>" string.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/util/Util.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/util/Util.java
@@ -23,6 +23,8 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.util.MultiValueMap;
+import org.springframework.web.util.UriComponentsBuilder;
 
 import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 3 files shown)


airsonic-main/src/main/java/org/airsonic/player/util/Util.java
```diff
@@ -133,6 +136,24 @@ public static String getURLForRequest(HttpServletRequest request) {
...
+     * @return The associated anonymized URL
+     */
+    public static String getAnonymizedURLForRequest(HttpServletRequest request) {
+
+        String url = getURLForRequest(request);
+        UriComponentsBuilder builder = UriComponentsBuilder.fromUriString(url);
+        MultiValueMap<String, String> components = builder.build().getQueryParams();
+
+        // See RESTRequestParameterProcessingFilter
+        if (components.containsKey("p")) builder.replaceQueryParam("p", URL_SENSITIVE_REPLACEMENT_STRING);
```
jvoisin: I think that it would be a bit better to explain that `p` is the password parameter too.

jvoisin: Wouldn't it would make sense to anonymize the username too?

fxthomas: Yup, you're correct! Done!

fxthomas: Done!

PR #1030, opened 2019-04-23
Use modern Spring mapping annotations

jvoisin:
Java is verbose enough, no need to make it worse.


airsonic-main/src/main/java/org/airsonic/player/controller/UploadController.java
```diff
@@ -71,7 +71,7 @@
     private SettingsService settingsService;
     public static final String UPLOAD_STATUS = "uploadStatus";
 
-    @RequestMapping(method = { RequestMethod.POST })
+		@PostMapping
```
muff1nman: Indentation is off here.

fxthomas:
Pulled on my server, I did not notice anything wrong but I'll keep it on for a few days to make sure.

PR #1031, opened 2019-04-23
Use builtin formatting for logs

jvoisin:
This should marginally improve performances,
but more importantly, it makes the code a bit
more clean.


airsonic-main/src/main/java/org/airsonic/player/service/upnp/FolderBasedContentDirectory.java
```diff
@@ -63,8 +63,8 @@
     public BrowseResult browse(String objectId, BrowseFlag browseFlag, String filter, long firstResult,
             long maxResults, SortCriterion[] orderby) throws ContentDirectoryException {
 
-        LOG.info("UPnP request - objectId: " + objectId + ", browseFlag: " + browseFlag + ", filter: " + filter +
-                ", firstResult: " + firstResult + ", maxResults: " + maxResults);
+        LOG.info("UPnP request - objectId: {0}, browseFlag: {1}, filter: {2}, firstResult: {3}, maxResults: {4}",
+            objectId, browseFlag, filter, firstResult, maxResults);
```
muff1nman: This is not slf4j compatible:  https://www.slf4j.org/faq.html#logging_performance

PR #1035, opened 2019-04-24
Handle ranges more correctly when streaming

### LINKED ISSUE #685: Playback occasionally stops with "upstream prematurely closed connection..." errors on server
## Problem description
Periodically when streaming music,  the player will cease playback (usually toward the end of a song, before starting the next one). I then see an "ERR_CONTENT_LENGTH_MISMATCH" in the browser console. Looking at server logs, I see many repeated instances of the same sort of error, which I assume are related:

```
2018/01/25 17:11:11 [error] 15161#0: *46668 upstream prematurely closed connection while reading upstream, client: REDACTED, server: REDACTED, request: "GET /airsonic/rest/stream.view?REDACTED", upstream: "http://127.0.0.1:8080/airsonic/rest/stream.view?REDACTED", host: "REDACTED"
```
I added
``` 
JAVA_ARGS=--server.connection-timeout=-1
```
to my sysconfig, but that doesn't seem to have had any effect.

## System information

 * **Airsonic version**: 10.1.1-RELEASE – December 16, 2017
 * **Operating system**: Centos 7
 * **Proxy server**: Nginx 1.12.2
 * **Client**: Chrome, DSub

## Additional notes
Airsonic is reading from a network drive, so I assume it is sometimes hitting a timeout threshold while waiting for the file read. I'm not finding a read timeout option in the embedded server config options though, and I'm very new to Tomcat / Java so don't know what else to try.

Thanks!!

EDIT:

More info. I am starting to suspect this has something to do with transcoding. It seems there are particular files (tending to be FLACs, is the current theory) that it struggles with. When it hits a file it doesn't like, it will continue to spam errors (the aforementioned ones, in browser console and nginx error log). It will also spam the following lines in the airsonic log:

```
2018-01-27 10:54:02.920  INFO --- o.a.p.io.PlayQueueInputStream            : gryphon listening to "Animal Crossing - Your Favourite Songs (Original Soundtrack)/03 Daytime In City Square.flac"
2018-01-27 10:54:02.922  INFO --- o.a.p.io.TranscodeInputStream            : Starting transcoder: [/var/airsonic/transcode/ffmpeg] [-i] [/var/airsonic/music/Kazumi Totaka _ Manaka Tominaga _ Shiho Fujii/Animal Crossing - Your Favourite Songs (Original Soundtrack)/03 Daytime In City Square.flac] [-map] [0:0] [-b:a] [192k] [-v] [0] [-f] [mp3] [-]
2018-01-27 10:54:08.750  INFO --- o.a.p.c.StreamController                 : Got HTTP range: 2756762-
2018-01-27 10:54:08.750  INFO --- o.a.p.io.PlayQueueInputStream            : gryphon listening to "Animal Crossing - Your Favourite Songs (Original Soundtrack)/03 Daytime In City Square.flac"
2018-01-27 10:54:08.752  INFO --- o.a.p.io.TranscodeInputStream            : Starting transcoder: [/var/airsonic/transcode/ffmpeg] [-i] [/var/airsonic/music/Kazumi Totaka _ Manaka Tominaga _ Shiho Fujii/Animal Crossing - Your Favourite Songs (Original Soundtrack)/03 Daytime In City Square.flac] [-map] [0:0] [-b:a] [192k] [-v] [0] [-f] [mp3] [-]
2018-01-27 10:54:14.523  INFO --- o.a.p.c.StreamController                 : Got HTTP range: 2756762-
```
(Some unrelated AudioScrobbler lines omitted)

It will continue spamming this same message every few seconds until I stop it from continuing to attempt playback.

EDIT2:

I have ruled out it being a network drive / timeout issue by copying an album containing offending files onto a local drive, adding that folder as a new Media source, and playing the duplicate version of the album. It fails on the exact same tracks, in the exact same way. I tracked down what looks like a very similar Madsonic issue here, which mentions problems with the audio scanner: http://forum.madsonic.org/viewtopic.php?t=2230 

The Madsonic author's suggested fixes seem irrelevant though as those options aren't available in Airsonic. 

EDIT3: 

I attempted rescanning my library with fast access mode enabled. No change.

tari:
Significant refactoring of the stream controller, to correctly handle Range requests by including a header as to the support for ranges in a given request (supported iff not transcoding) if a range is requested, and use chunked encoding when transcoding.

I believe this fixes #685 (it appears to on my system).


tari:
Cleaned up the logging and experimental things I added, I think it's good now.

airsonic-main/src/main/java/org/airsonic/player/controller/StreamController.java
```diff
@@ -151,28 +153,48 @@ public void handleRequest(HttpServletRequest request, HttpServletResponse respon
...
+                    if (range != null) {
+                        response.setStatus(HttpServletResponse.SC_PARTIAL_CONTENT);
+                        response.setHeader("Accept-Ranges", "bytes");
+
+                        // Both ends are inclusive
+                        long startByte = range.getFirstBytePos();
+                        long endByte = range.isClosed() ? range.getLastBytePos() : file.getFileSize() - 1;
+
+                        response.setHeader("Content-Range",
+                                String.format("bytes %d-%d/%d", startByte, endByte, endByte + 1));
```
muff1nman: Instead of `endByte + 1`, shouldnt this be `file.getFileSize()`?

airsonic-main/src/main/java/org/airsonic/player/io/RangeOutputStream.java
```diff
@@ -55,7 +55,7 @@ public RangeOutputStream(OutputStream out, HttpRange range) {
      * @return The possibly wrapped output stream.
      */
     public static OutputStream wrap(OutputStream out, HttpRange range) {
-        if (range == null) {
+        if (range == null || (range.getFirstBytePos() == 0 && !range.isClosed())) {
```
muff1nman: Hmm, can you provide any additional context for this change?

muff1nman:
Also looks like you'll need to add an explicit depdency for the following:

```
[WARNING] Used undeclared dependencies found:
[WARNING]    com.google.code.findbugs:jsr305:jar:3.0.2:compile
```

Should just be able to put it in the airsonic-main/pom.xml.

airsonic-main/src/main/java/org/airsonic/player/io/RangeOutputStream.java
muff1nman: Reading this again, and it makes sense although imo should be unnecessary assuming RangeOutputStream behaves correctly.

fxthomas:
I'll pull that and leave it running on my server for a while, I'll let you guys know if I see something going bad!

airsonic-main/src/main/java/org/airsonic/player/io/RangeOutputStream.java
jvoisin: Since it's not super-duper-obvious, a comment would be nice ♥

muff1nman:
Wrote some tests for this in #1026. With this PR the responses now actually line up with ffmpeg's generated output. Well done!

Besides getting travis to be happy, I'd like to see this PR reduced to the minimum amount of changes possible. I.e. no line changes, no refactoring, etc. Those can be done separately if really needed. The reason for this is two fold:

1. This is a critical part of code and the smaller amount of changes we can make the easier it is to understand the impacts.
2. A smaller changeset is easier to revert if necessary.

@tari let me know your thoughts on simplifying. I'd also be more than happy to send a PR against your PR to help.

fxthomas:
I mentioned I had been getting this issue lately in https://github.com/airsonic/airsonic/issues/685#issuecomment-495008596, but the current PR unfortunately doesn't appear to fix it, it just makes it appear 100% of the time on one specific album/song.

1. Go to Airsonic and click on "Play Album". The first song plays.
2. Wait till the end.
3. The play queue appears to go to the second song, but the playback stops.

The player looks like this when this happens (note the "Live Broadcast"):



The Firefox logs contain this message (and the stream request appears incomplete in the Network panel):

```
AbortError: The fetching process for the media resource was aborted by the user agent at the user's request.
```

Here are file sizes and durations (as reported by `ffprobe`):

* FLAC: 00:03:08.03
* Manual transcode using the same command as the logs: 7521175 bytes, 00:03:08.03
* Manual stream with `curl` (using Firefox' handy 'Copy as cURL', so this should be exactly the same request): 7520000 bytes, 00:03:08.00 (`mp3check` reports "130 bytes missing from last frame")
* After your PR, manual stream with `curl`: 9001796 bytes, 00:03:45

muff1nman:
@fxthomas can you send me your example file?

airsonic-main/src/main/java/org/airsonic/player/controller/StreamController.java
tari: Yes, this should be the actual file size.

airsonic-main/src/main/java/org/airsonic/player/io/RangeOutputStream.java
tari: Yeah, I wanted to short-circuit wrapping the stream if it's unnecessary. Added a comment to that effect.

fxthomas:
Sure, in your mailbox!

In the mean time, it seems that I'm just a poor dumb human making mistakes. I restarted Firefox, which made the issue go away with your PR. I also happened to take the wrong notes for the stream size/duration after PR, edited my earlier comment to reflect that. All I can say is... great job on the PR! :smile:

tari:
As for reducing the impact of these changes overall, I think the refactoring is necessary for everything to make sense (without it, it's very difficult to follow what should happen) but much of the reformatting and other changes could be split up. I'll do some rebasing so refactoring and reformatting are more separate, but if you have further opinions on simplification I'm happy to see them.

tari:
..and now it's been split into a number of smaller commits.

bschaefer99:
How exciting! Thanks team!

Lightsockie:
Most excellent :ok_hand:

Issue #608, opened 2017-10-30
Error on MySQL version 5.6 migration


hyptechdev2015:
## Problem description
```
Caused by: liquibase.exception.DatabaseException: Invalid default value for 'created_date' [Failed SQL: CREATE TABLE airsonichyptech.system_avatar (id INT AUTO_INCREMENT NOT NULL, name VARCHAR(512) NULL, created_date datetime DEFAULT NOW() NOT NULL, mime_type VARCHAR(512) NOT NULL, width INT NOT NULL, height INT NOT NULL, data BLOB NOT NULL, CONSTRAINT PK_SYSTEM_AVATAR PRIMARY KEY (id))]
	at liquibase.executor.jvm.JdbcExecutor$ExecuteStatementCallback.doInStatement(JdbcExecutor.java:301) ~[liquibase-core-3.5.1.jar:na]
	at liquibase.executor.jvm.JdbcExecutor.execute(JdbcExecutor.java:55) ~[liquibase-core-3.5.1.jar:na]
	at liquibase.executor.jvm.JdbcExecutor.execute(JdbcExecutor.java:107) ~[liquibase-core-3.5.1.jar:na]
	at liquibase.database.AbstractJdbcDatabase.execute(AbstractJdbcDatabase.java:1273) ~[liquibase-core-3.5.1.jar:na]
	at liquibase.database.AbstractJdbcDatabase.executeStatements(AbstractJdbcDatabase.java:1255) ~[liquibase-core-3.5.1.jar:na]
	at liquibase.changelog.ChangeSet.execute(ChangeSet.java:568) ~[liquibase-core-3.5.1.jar:na]
	... 38 common frames omitted
```

### Solutions

need change "datetime" type to "TIMESTAMP"

and add "zeroDateTimeBehavior=convertToNull" in **context.xml**
reference: https://airsonic.github.io/docs/database/
```
CREATE TABLE airsonichyptech.system_avatar (id INT AUTO_INCREMENT NOT NULL, name VARCHAR(512) NULL, created_date **TIMESTAMP** DEFAULT NOW() NOT NULL, mime_type VARCHAR(512) NOT NULL, width INT NOT NULL, height INT NOT NULL, data BLOB NOT NULL, CONSTRAINT PK_SYSTEM_AVATAR PRIMARY KEY (id))
```
```
<Resource name="jdbc/airsonicDB" auth="Container"
    type="javax.sql.DataSource"
    maxActive="20"
    maxIdle="30"
[...]

PR #1040, opened 2019-04-28
Fix for Issue 608

agodlydeciple:
applied fixed to legacy table creations that allows mysql/mariadb to not fail. Also updated cxf.version to 3.2.8 as it seems 3.2.6 is no longer available when trying to package the project.

This seems to be a cleaner fix than what was mentioned in issue #608

### DIFF
--- a/airsonic-main/src/main/resources/liquibase/db-changelog.xml
+++ b/airsonic-main/src/main/resources/liquibase/db-changelog.xml
@@ -4,7 +4,7 @@
         xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.1.xsd">
     <property name="binary_type" dbms="hsqldb" value="binary" />
     <property name="binary_type" value="blob" />
-    <property name="curr_date_expr" value="current_timestamp" />
+    <property name="curr_date_expr" value="NULL" />
     <property name="varchar_type" dbms="mysql" value="varchar(${mysqlVarcharLimit})" />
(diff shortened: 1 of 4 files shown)


muff1nman:
Im not sure I like the idea of this. I think we can come up with a better fix.

agodlydeciple:
@muff1nman I agree that it probably isn't the cleanest fix.  It was just the quickest and easiest way to get airsonic to use mysql without the need for editing a ton of source files. 

I am all for coming up with a different approach to fixing this issue.  

Do you have any thoughts on the subject?

muff1nman:
Let me get a mysql instance fired up again and do some experimentation. I coulda swore mysql was working at some point. What version are you using?

agodlydeciple:
mysql  Ver 15.1 Distrib 5.5.60-MariaDB, for Linux (x86_64) using readline 5.1 on CentOS 7.6

muff1nman:
I think that version of mysql is too old. Try a version at or newer than what is mentioned in the [docs](https://airsonic.github.io/docs/database/).

agodlydeciple:
Hmmm that's possible.  I am a little confused though.
image

MariaDB states it needs version 10.2.  I am using MariaDB on my instance and it is version 15.1 which seems to be based on mysql 5.5. So i wonder if there is a mismatch between the versions and the airsonic docs?

fxthomas:
Related to https://github.com/airsonic/documentation/pull/66 (doc updates to mention workarounds for MariaDB <10.2)

jvoisin:
Shall we close this PR, now that we have a workaround?

PR #1041, opened 2019-04-29
Remove the dependency on ant

jvoisin:
Apparently, `ant` isn't used for anything in airsonic.

### DIFF
--- a/airsonic-main/pom.xml
+++ b/airsonic-main/pom.xml
@@ -134,12 +134,6 @@
             <version>3.0.3</version>
         </dependency>
 
-        <dependency>
-            <groupId>org.apache.ant</groupId>
-            <artifactId>ant</artifactId>
-            <scope>runtime</scope>
-        </dependency>
-
         <!-- connection pool manager, fixes #860 (external database performance) -->  
         <dependency>
             <groupId>org.apache.commons</groupId>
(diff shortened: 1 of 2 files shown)


muff1nman:
This probably needs to be redone considering the update with master seems to have borked it.

fxthomas:
This appears to break the following lines in `airsonic-main/pom.xml`:

```xml
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-antrun-plugin</artifactId>
                <executions>
                    <execution>
                        <phase>generate-resources</phase>
                        <configuration>
                            <target>
                                <tstamp/>
                                <copy file="${basedir}/src/main/resources/org/airsonic/player/i18n/ResourceBundle_en.properties"
                                      tofile="${project.build.directory}/classes/org/airsonic/player/i18n/ResourceBundle.properties"/>
                                <echo file="${project.build.directory}/classes/build_number.txt">${buildNumber}</echo>
                                <!--suppress MavenModelInspection -->
                                <echo file="${project.build.directory}/classes/build_date.txt">${DSTAMP}</echo>
                                <echo file="${project.build.directory}/classes/version.txt">${project.version}</echo>
                            </target>
                        </configuration>
                        <goals>
                            <goal>run</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
```

jvoisin:
Sigh, you're right.
We're using a complete build system to simply copy a file and write a couple of strings.
I guess this should be doable via maven too.

Issue #953, opened 2019-04-01
Add a way to visualize the testsuite's coverage


jvoisin:
It would be cool to use something like [coveralls]( https://coveralls.io/ ) or [codecov]( https://codecov.io/ ) to provide a way to visualize the testsuite's coverage.

jvoisin:
[codecov]( https://codecov.io/gh/jvoisin/airsonic/tree/0ad81b31beef0c2e735b0a0c3e8565ba3fad15a8/airsonic-main/src/main/java/org/airsonic/player ) is working and simpler to use than coveralls. Do we want to use it? Currently it's only deployed for my fork.

PR #1012, opened 2019-04-13
Add a gitlab-ci.yml file

jvoisin:
Since I was a bit bored this morning, I started to replicate a bit the github repositories to gitlab (see #888 for details). This commit adds a `.gitlab-ci.yml` file to test airsonic inside gitlab-ci. It currently fails because of vulnerable dependencies, that are being fixed in other PR (#1006, …).

PR #1051, opened 2019-05-09
WIP: Use codecov for coverage visualization

jvoisin:


jvoisin:
This should close #953, and the result should be visible [here]( https://codecov.io/gh/airsonic/airsonic )

.travis.yml
```diff
@@ -8,6 +8,9 @@ script:
 matrix:
   include:
   - jdk: oraclejdk8
+    after_success:
+      - mvn cobertura:cobertura
+      - bash <(curl -s https://codecov.io/bash)
```
muff1nman: Have you considered just integrating this into the maven lifecycle? Keeping things wrapped into the maven lifecycle increases portability and makes things like #1012 much easier.

jvoisin: No idea how to use maven, unfortunately :/

jvoisin: I'm not sure that it's worth it:

- the .travis-ci.yml and .gitlab-ci.yml are simple enough
- everyone knows yaml, fewer know maven lifecycles
- it's unlikely that we'll want to send the coverage from many places: it'll only be from travis-ci, and should we move entirely to gitlab-ci, it'll just be 3 lines to move to an other yaml file
- having this in the yaml files is the recommended way to use codecov.

eharris:
I'm a fan of doing this or a similar tool.

muff1nman:
I'm not convinced this is configuring cobertura correctly as there is a lot of nonsense in the logs that probably shouldn't be there:

- stacktraces on instrumentation errors
- mulitple logging impl warnings/errors
- extraneous logging in general. Our builds are currently outputting about 2K lines of output during travis runs and we should keep it in that ballpark

jvoisin:
I switched to Jococo, which is maintained and less awful than cobertura: it seems that it's working correctly now \o/

almet:
This looks great to me, should we merge?

pom.xml
```diff
@@ -322,6 +322,25 @@
...
+                <version>0.8.5</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>prepare-agent</goal>
+                        </goals>
+                    </execution>
+                    <execution>
+                        <id>report</id>
+                        <phase>test</phase>
```
muff1nman: I'm not sure it makes sense to put this in the test phase.. What about the verify phase instead?

muff1nman: Or what about no phase at all and this is only run adhoc in travis?

almet: Just so anybody can get some context, [from the maven docs](https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html):

> test - test the compiled source code using a suitable unit testing framework. These tests should not require the code be packaged or deployed
> verify - run any checks on results of integration tests to ensure quality criteria are met

In Jacoco, the test phase seems to be used in the default templates, and this seems a good value to me. Verify could be used in case we want to meet a certain criteria, maybe if we want to ensure we have a minimum code coverage of XX%?

PR #1059, opened 2019-05-19
WIP: Remove custom implementation of boundedlist

jvoisin:
Apache commons are already providing an
implementation of boundedlist called CircularFiroBuffer.


jvoisin:
See #1060.

Issue #1048, opened 2019-05-08
CVE-2019-5021


muff1nman:
Since we are using an alpine base, our docker image needs to be updated.

jooola:
I would add docker image building process to the travis / gitlab pipelines.

muff1nman:
@jooola its part of the maven process actually so its already in there ;)

PR #1062, opened 2019-05-20
Update docker base image

muff1nman:


muff1nman:
Fixes #1048

jooola:
What about using `alpine:latest` ? I don't think we will face many bugs while sticking to latest.

muff1nman:
@jooola I really dislike the latest tag because it is so mutable. I think the major.minor versions are a good tradeoff.

PR #1064, opened 2019-05-20
Docs for creating a release

muff1nman:


jooola:
We need to force any developers that will release to expose theyre public key used to sign the war file to the documentation. Since we added a way to verify the integrity and the authenticity of the .war.
And add it to the documentation.

Or generate a Signing key, shared between maintainers.

Issue #902, opened 2019-03-18
Track seek bar gets uselessly dark after a song ends


kaysond:
This is on Firefox 65.0.2. Not sure if it affects other browsers as well.

Take a look at https://imgur.com/a/HRdvUPI
The top is when the page first loads. The bottom is after a song finishes and it transitions to the next one.

It looks like the problem is with the code below. On page load, the style is set to `scaleX(0)`. Presumably when the next song loads, it starts this progress bar, but when it finishes, it doesn't reset. The result is two overlaid bars causing the darkness.
```
<span class="mejs__time-loaded" style="transform: scaleX(1);"></span>
```

jooola:
Might be duplicate.

kaysond:
you'd think someone would've reported this already. I did search but couldn't find it

jvoisin:
Any chances you can issue a PR to fix this?

kaysond:
I did try to dig through it a little bit with firefox dev tools but didn't find an immediate answer. My guess is that it's a problem with the mediaelement library. I can keep looking

jooola:
This should have been fixed in 10.3, please test and close the ticket if so.

kaysond:
Doesn't look like it... The behavior has changed slightly though. Now in addition to having this glitch after transitioning to another track, it also happens immediately after playing the first track.

Interestingly, MS Edge does not seem to have the issue on first track play, but I couldn't actually get it to transition to the next track in the play queue without clicking play again... so not sure if it happens between tracks as well
![Untitled](https://user-images.githubusercontent.com/1147328/58121457-60d2b300-7bbc-11e9-84ff-9333f1ce227f.png)

I'll try to debug further

muff1nman:
Seeing this as well.

jooola:
Weird, I just added flac back to the mp3 audio trans-coding rule and I'm seing this as well.

tesshucom:
It seems that the fix has already been applied to the master.
Is it resolved by #1080?

PR #1080, opened 2019-05-23
Fix remaining issues with playback stopping intermittently

fxthomas:
This is a follow-up of #1035 and #685, and I would love some testers on this!

@tari's patch helped a lot, but I was still seeing intermittent playback failures when switching songs, though not always and not with the same message...

I'm now fairly sure that this is caused by some of our JS code trying to run `play()` while the player is still loading the stream, causing a race condition.

I experimented a bit with the `skip(index, position)` function at the core of the play queue and added some logging to the player. I came to the conclusion that some methods (I identified `player.play()`, `player.load()` or setting `player.currentTime`) will automatically start loading the stream in the background, and calling any of them while the player is still loading will crash and burn.

The following screenshot shows the player load a first stream, then a a second (see both logs from `playQueue.view`) ; the request is then interrupted by something else and the player does not try again.

Screenshot from 2019-05-24 00-52-30

The following patch makes sure that all code paths in `skip` load *AND* start the stream exactly once, and _then_ start playback when the _canplay_ event fires. This appears to fix everything for me. Tested on FF 68 and Chromium 74.

### DIFF
--- a/airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
+++ b/airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
@@ -577,6 +577,12 @@
         // the 'createMediaElementPlayer()' function above.
         //
         // player.play();
+
+        // FIXME: Remove once https://github.com/mediaelement/mediaelement/issues/2650 is fixed.
+        //
+        // Instead, we're triggering a 'waiting' event so that the player shows
+        // a progress bar to indicate that it's loading the stream.
+        player.dispatchEvent(new Event("waiting"));
     }
 
     function skip(index, position) {
--- a/airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
+++ b/airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
@@ -52,7 +52,7 @@
                 }
             }});
 
-        <c:if test="${model.player.web}">createPlayer();</c:if>
+        <c:if test="${model.player.web}">createMediaElementPlayer();</c:if>
 
         $("#playlistBody").sortable({
             stop: function(event, ui) {
@@ -145,8 +145,21 @@
         onNext(repeatEnabled);
     }
 
-    function createPlayer() {
-        $('#audioPlayer').get(0).addEventListener("ended", onEnded);
+    function createMediaElementPlayer() {
+
+        var player = $('#audioPlayer').get(0);
+
+        // Once playback reaches the end, go to the next song, if any.
+        player.addEventListener("ended", onEnded);
+
+        // FIXME: Remove once https://github.com/mediaelement/mediaelement/issues/2650 is fixed.
+        //
+        // Once a media is loaded, we can start playback, but not before.
+        //
+        // Trying to start playback after a song has endred but before the
+        // 'canplay' event for the next song currently causes a race condition
+        // in the MediaElement.js player (4.2.10).
+        player.addEventListener("canplay", function() { player.play(); });
     }
 
     function getPlayQueue() {
@@ -532,6 +545,40 @@
         }
     }
 
+    function loadMediaElementPlayer(song, position) {
+        var player = $('#audioPlayer').get(0);
+
+        // Is this a new song?
+        if (player.src != song.streamUrl) {
+            // Stop the current playing song and change the media source.
+            player.src = song.streamUrl;
+            // Inform MEJS that we need to load a new media source. The
+            // 'canplay' event will be fired once playback is possible.
+            player.load();
+            // The 'skip' function takes a 'position' argument. We don't
+            // usually send it, and in this case it's better to do nothing.
+            // Otherwise, the 'canplay' event will also be fired after
+            // setting 'currentTime'.
+            if (position && position > 0) {
+                player.currentTime = position;
+            }
+
+        // Are we seeking on an already-playing song?
+        } else {
+            // Seeking also starts playing. The 'canplay' event will be
+            // fired after setting 'currentTime'.
+            player.currentTime = position || 0;
+        }
+
+        // FIXME: Uncomment once https://github.com/mediaelement/mediaelement/issues/2650 is fixed.
+        //
+        // Calling 'player.play()' at this point may work by chance, but it's
+        // not guaranteed in the current MediaElement.js player (4.2.10).  See
+        // the 'createMediaElementPlayer()' function above.
+        //
+        // player.play();
+    }
+
     function skip(index, position) {
         if (index < 0 || index >= songs.length) {
             return;
@@ -541,16 +588,12 @@
         currentStreamUrl = song.streamUrl;
         updateCurrentImage();
 
+        // Handle ChromeCast player.
         if (CastPlayer.castSession) {
             CastPlayer.loadCastMedia(song, position);
+        // Handle MediaElement (HTML5) player.
         } else {
-            if ($('#audioPlayer').get(0).src != song.streamUrl) {
-                $('#audioPlayer').get(0).src = song.streamUrl;
-                $('#audioPlayer').get(0).load();
-                console.log(song.streamUrl);
-            }
-            $('#audioPlayer').get(0).currentTime = position ? position : 0;
-            $('#audioPlayer').get(0).play();
+            loadMediaElementPlayer(song, position);
         }
 
         updateWindowTitle(song);


airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
```diff
@@ -146,7 +146,27 @@
...
+        var player = $('#audioPlayer').get(0);
+        player.addEventListener("ended", onEnded);
+        player.addEventListener("canplay", function() { player.play(); });
+
+        // Verbose debug logging for player events
+        player.addEventListener("canplay", function(ev) { console.debug("Received audio event: canplay"); console.debug(ev); })
+        player.addEventListener("loadedmetadata", function(ev) { console.debug("Received audio event: loadedmetadata"); console.debug(ev); })
+        player.addEventListener("seeking", function(ev) { console.debug("Received audio event: seeking"); console.debug(ev); })
+        player.addEventListener("seeked", function(ev) { console.debug("Received audio event: seeked"); console.debug(ev); })
+        player.addEventListener("canplay", function(ev) { console.debug("Received audio event: canplay"); console.debug(ev); })
```
muff1nman: Why is there two canplay?

muff1nman:
Got some AbortError in Firefox, but the next song started playing just fine. Not sure what its abortin as it happens after the pause and ended events.

```
Received audio event: play playQueue.view:226:56
play { target: null, isTrusted: false, detail: {…}, eventPhase: 0, bubbles: false, cancelable: false, returnValue: true, defaultPrevented: false, composed: false, timeStamp: 248967, … }
playQueue.view:226:101
Received audio event: playing playQueue.view:227:59
playing { target: null, isTrusted: false, detail: {…}, eventPhase: 0, bubbles: false, cancelable: false, returnValue: true, defaultPrevented: false, composed: false, timeStamp: 248975, … }
playQueue.view:227:107
Server side script session id timed out. New session automatically created engine.js:1:8190
Received audio event: pause playQueue.view:228:57
pause { target: null, isTrusted: false, detail: {…}, eventPhase: 0, bubbles: false, cancelable: false, returnValue: true, defaultPrevented: false, composed: false, timeStamp: 412222, … }
playQueue.view:228:103
https://www.barracks.andrewdemaria.com/airsonic/stream?player=79&id=11686 playQueue.view:640:17
Received audio event: ended playQueue.view:229:57
ended { target: null, isTrusted: false, detail: {…}, eventPhase: 0, bubbles: false, cancelable: false, returnValue: true, defaultPrevented: false, composed: false, timeStamp: 412225, … }
playQueue.view:229:103
AbortError: The operation was aborted. 
Received audio event: loadedmetadata playQueue.view:222:66
loadedmetadata { target: null, isTrusted: false, detail: {…}, eventPhase: 0, bubbles: false, cancelable: false, returnValue: true, defaultPrevented: false, composed: false, timeStamp: 412680, … }
playQueue.view:222:121
Received audio event: canplay playQueue.view:221:59
canplay { target: null, isTrusted: false, detail: {…}, eventPhase: 0, bubbles: false, cancelable: false, returnValue: true, defaultPrevented: false, composed: false, timeStamp: 412683, … }
playQueue.view:221:107
Received audio event: canplay playQueue.view:225:59
canplay { target: null, isTrusted: false, detail: {…}, eventPhase: 0, bubbles: false, cancelable: false, returnValue: true, defaultPrevented: false, composed: false, timeStamp: 412683, … }
playQueue.view:225:107
Received audio event: play playQueue.view:226:56
play { target: null, isTrusted: false, detail: {…}, eventPhase: 0, bubbles: false, cancelable: false, returnValue: true, defaultPrevented: false, composed: false, timeStamp: 412689, … }
playQueue.view:226:101
Received audio event: playing
```

muff1nman:
The AbortError might be due to the fact that currentTime is changed right after loading a new song as aluded to here: https://bugzilla.mozilla.org/show_bug.cgi?id=1507193#c4. 

Why change currentTime to position when loading a new song?

muff1nman:
One thing I'm still experiencing is abrupt skips. I.e. the next song starts playing before the current one is over. I wonder if that's because I'm on a degraded network separate from the server and its missing data during which the onended event is fired.

This link maybe supports the above hypothesis:  https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/ended_event

> The ended event is fired when playback or streaming has stopped because the end of the media was reached **or because no further data is available**. This event occurs based upon HTMLMediaElement (<audio> and <video>) fire ended when playback of the media reaches the end of the media.

However I think this behavior isn't directly related to playback stopping which I have yet to see with this patch.

muff1nman:
Tangent: Looks like not providing content length for transcoded music does the following in chrome: image

Not sure why mediaelement doens't do that in firefox.

airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
fxthomas: Mostly debugging. I intend on removing the extra lines after I'm sure this works.

fxthomas:
Re: changing `currentTime`, we currently set it every time, which unfortunately also starts loading the stream. It's a parameter for `skip`, so I don't like to ignore it, but at least this patch avoids setting it when.there's no need to.

Re:Chrome appearance, I noticed the same thing. I usually use FF, and it only flashes "Live Broadcast" for a fraction of a second before showing the progress bar. (I personally like having the progress bar, but it's not related to the PR, I think.)

Re: the AbortError you saw, can you check that the data streamed from Airsonic is complete? Did you check media files with transcoding or without? I can reliably reproduce the issues I experience with FLAC files, less so with files already in MP3.

chaosbiber:
This PR has been working fine for me, but I have only compared it to 10.3.1.
After the last release update the playback stopped between **every** song in Firefox, accompanied by a "Client unexpectedly closed connection" server message. Chrome behaviour improved with 10.3.1 but with the state of this PR both browsers have been playing different files without stopping. I've also noticed the "Live Broadcast" instead of the bar in Chrome for transcoded files but working playback is more important. No AbortErrors for me.

fxthomas:
Nice, thank you very much for testing this! I haven't seen any playback errors anymore since using this either, so we're on a good track.

I rebased/edited the commit bit to make that ready for merging, feel free to suggest improvements.

jvoisin:
Shouldn't this be fixed upstream in mediaelement.js?

airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
```diff
@@ -530,6 +563,38 @@
...
+                player.currentTime = position;
+            }
+            // Log the song URL.
+            console.log(song.streamUrl);
+
+        // Are we seeking on an already-playing song?
+        } else {
+            // Seeking also starts playing. The 'canplay' event will be
+            // fired after setting 'currentTime'.
+            player.currentTime = position ? position : 0;
```
jvoisin: I don't understand this line: If position is non-zero, the use position, otherwise, use zero?
If position can be nil, the javascript-way of casting it to zero would be `player.currentTime = position || 0;`.

fxthomas: Ah, thanks! That's what I was going for.

fxthomas:
Not sure, we were really using it incorrectly before this commit.

And implementing the promises API correctly is probably a big change considering how MEJS works right now.

jvoisin:
I guess we can get this merged as soon as you clean the comments up?

fxthomas:
Yup, go ahead!

jvoisin:
As soon as you remove the comments :P 
Or you want me to do it?

fxthomas:
Ah, I thought you meant the PR comments!

The fact that MEJS had a completely different interface than HTML5 ME threw me off at the beginning, I'd rather leave a big comment explaining that. I'll remove the debug lines but leave the rest, does that sound good?

jooola:
Would be nice to ship this asap in 10.3.2 because it is getting really annoying!

jooola:
I just tested this PR + the current state of master @ 4f59c72bbe8514a1959a6aca6d4a52e064cc6bd9 .
It is working but same as above, the browser console is raising an error: `AbortError: The operation was aborted.`

On Firefox 67.0, debian 10

jvoisin:
But comments, in mean the commented debug lines, not the useful ones, sorry :)

@jooola Could it be a caching issue in your browser?

jooola:
> @jooola Could it be a caching issue in your browser?

Nope, tried again with a fresh browser.

fxthomas:
Is transcoding enabled? Can you try running a tool like mp3check (if
playing as an MP3 stream) on the whole data Airsonic is playing?

jooola:
I only removed the flac transcoding since it was broken, I will add it back.

I'll check this with your tool (if it exists on Linux).

But even with flac disabled it shouldn't raise anything right ?

fxthomas:
If the original MP3 file is corrupted (even if it's very slightly, at the end) I think this error can appear. At least Debian (https://packages.debian.org/source/sid/mp3check) and Arch (https://aur.archlinux.org/packages/mp3check/) have packages. Let me know if you find something!

Anyway, debug lines removed!

airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
```diff
@@ -530,6 +545,38 @@
...
+            player.load();
+            // The 'skip' function takes a 'position' argument. We don't
+            // usually send it, and in this case it's better to do nothing.
+            // Otherwise, the 'canplay' event will also be fired after
+            // setting 'currentTime'.
+            if (position && position > 0) {
+                player.currentTime = position;
+            }
+            // Log the song URL.
+            console.log(song.streamUrl);
```
jvoisin: I don't think that we should ever output things in the console, except critical errors and stacktraces.

jooola:
So I checked my mp3 file and the only error I get is:
`31967 bytes of junk before first frame header`

But I also noticed that if the song isn't played long enough everything is working fine. Also the error happen on flac files, even if trans-coding for flac has been disabled.

airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
fxthomas: You're right, done.

fxthomas:
I also noticed that with this PR we loose the nice "loading media" infinite progress bar that currently show up, like this:

image

Without this the user has no idea that something is still loading, so I'd like to try and find a way to keep it.

jooola:
> I also noticed that with this PR we loose the nice "loading media" infinite progress bar that currently show up when changing sources, like this:
Without this the user has no idea that something is still loading, so I'd like to try and find a way to keep it.

See https://github.com/airsonic/airsonic/issues/902#issuecomment-499261928

I also just noticed that the last track of my play queue is playing endlessly without having the repeat mode on. Not sure whether this is a separate issue.

fxthomas:
@joola Thanks, I fixed that.

Also, I tried to make the progress bar show up, at the cost of sometimes seeing the dreaded `AbortError`, see a972868.

Can you guys re-test the PR and let me know if you find issues?

muff1nman:
I'm wondering if it makes sense to call `player.pause()` when changing sources. Testing with it seems to prevent the AbortError issue on firefox.

jooola:
Ok I just builded a new war from latest master 6baa2d09d6db135e8ed34ebab1e7fa1dfd843d13 and your PR. Damn build tests are getting better and better !

With flac transcoding off:
- The seekbar works as expected, and isn't dark and complete at song start.
- Abort error is still showing up.

With flac transcoding on:
- The seekbar don't work as expected, and is completely dark at song start.
- Abort error is still showing up.

Also the endlessly playing track isn't fixed.

Damn maybe I forgot to update the branch or something like this. Gonna rebuild to be sure...

jooola:
Could you uncomment the debugging code for now, so we can get more details on what's going on. And comment it back only once everything is working and ready to be merged ?

fxthomas:
@jooola Which commit did you test? Can you try with just 5e406f0f71a73b1a54e3c5ebd272b54963f81a4e?

At this point I think we're going to have to try and debug inside MediaElement.js itself, although I don't like it too much...

fxthomas:
@muff1nman Will try, good idea.

jooola:
The last test I made was with a97286892f7954b1c37470d5be2c08ba4153620e .
You rebased so I don't recall were was my first test.

Should I cherrypick the commit you suggested or keep the entire branch and go back at this commit ?

airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
muff1nman: Is line 174 not the same as line 170?

fxthomas: You're right, I missed that ;)

fxthomas:
Nah, this was not rebased, you tested the correct commit. It's just that while I'm not too sure that a972868 doesn't introduce issues, 5e406f0 should work fine.

Kieltux:
Hello, I updated yesterday my Subsonic 5.2.1 installation to Airsonic 10.3.1 (build with maven at my server). With or without transcoding I did have the problem, that the webplayer stops after every song. The browser notification shows the next song, but the next song was not played. Then I did a downgrade to 10.2.1 and everything works.
Now I just build 10.3.1 again with the fix from fxthomas and everything is fine so far.

harawata:
I also experienced a similar issue, but with Safari 12.1.1 (14607.2.6.1.1) (no transcoding).
Since jQuery was upgraded to version 3, Safari does not play the next song in the Play queue when a song ends.
There is no error in the Safari's console nor in the server log.

After some digging, I found that the Safari's problem is triggered by this [spec change](https://github.com/jquery/jquery/commit/5cbb234dd3273d8e0bbd454fb431ad639c7242c1) in jQuery 3 and the 'ended' event handler bound during `init()` is not invoked when a song ends (as if it were not bound).
Simply moving the `init()` call to the end of the body resolved the problem, so it seems to be a pure JavaScript (or jQuery) issue in this particular case.


```diff
diff --git a/airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp b/airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
index 965c327..69cc561 100644
--- a/airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
+++ b/airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
@@ -26,7 +26,7 @@
     </style>
 </head>
 
-<body class="bgcolor2 playlistframe" onload="init()">
+<body class="bgcolor2 playlistframe">
 
 <span id="dummy-animation-target" style="max-width: ${model.autoHide ? 50 : 150}px; display: none"></span>
 
@@ -867,5 +867,9 @@
     };
 </script>
 <script type="text/javascript" src="https://www.gstatic.com/cv/js/sender/v1/cast_sender.js?loadCastFramework=1"></script>
-
+<script type="text/javascript">
+    $(function(){
+        init();
+    });
+</script>
 </body></html>
```

Although this PR might be good for the AbortError, Safari needs some additional JavaScript adjustment.
Without the above `init()` change, various event handlers stopped working after this patch was applied.

fxthomas:
Okay, I think I got at least a short reproducible case for Firefox, see [here](https://github.com/fxthomas/mediaelement-test-aborterror).

Can somebody else clone this repo and test?

I'll send an issue to MEJS if this sounds like a real issue on other's computers.

RedGreen007:
Alright, SO. 

Browser: Firefox
OS: Rasbian 4.14 (airsonic) Windows 10 (pc accessing said server)
Java: 8 (i think)
Proxy: Nada
Client: Webplayer.

When FLAC is played on the webplayer it will play until about the 1:10 mark, then stop playing while the timer continues to move. It will then resume where it left off after a couple of seconds. The buffer bar never reaches the end. Though it stopped skipping tracks it does this now. It's almost like it buffers for a while, then stops and theres a glitch in the buffer and then it starts up again. Not sure.

DOES NOT DO THIS WITH MP3! It will almost instantly buffer an MP3 to full and play with zero issues.

Browser: Chrome
Everything else: The same

Chrome does not suffer from these issues it appears. Transcoding I believe is disabled.

fxthomas:
Thanks! Can you add the version of Airsonic you tested? I believe you used my branch, right?

fxthomas:
I added an issue with the MEJS folks, let's see what they have to say on it before I dive into the source code.

RedGreen007:
Edited. I think that is the correct information.

muff1nman:
@fxthomas - cloned and confirmed the mejs one has issues with AbortError but the mejs_fix plays the next song fine with no AbortError.

muff1nman:
@RedGreen007 which version of airsonic. Comments on this PR are specifically for airsonic built on this branch.

RedGreen007:
Wow, it didn't save my edit.


Version | 10.4.0-SNAPSHOT – June 12, 2019
-- | --
Server | Apache Tomcat/8.5.14 (Debian), java 1.8.0_65, Linux (144.0 MB / 225.6 MB)

muff1nman:
The `onEnded` function can be removed with this change, correct?

RuralHunter:
I tested the patch with the latest snapshot. It doesn't stop on the second song but... it skips the second song and goes to the third one! Seems it still has problem when starting the second song.

RedGreen007:
So, I was looking at the log file and wanted to copy this and see if it helps at all.


2019-06-17  14:22:31.684  INFO --- o.a.p.io.PlayQueueInputStream            :  192.168.1.109: admin listening to Opus Eponymous/02. Con Clavi Con  Dio.flac
--
2019-06-17 14:22:31.790  INFO ---  o.a.p.c.StreamController                 : 192.168.1.109: Client  unexpectedly closed connection while loading  http://192.168.1.126:7979/airsonic/stream?player=1&id=24019  (java.io.IOException: Broken pipe)
2019-06-17 14:22:31.843  INFO ---  o.a.p.io.PlayQueueInputStream            : 192.168.1.109: admin  listening to Opus Eponymous/02. Con Clavi Con Dio.flac
2019-06-17 14:23:05.443  INFO ---  o.a.p.c.StreamController                 : 192.168.1.109: Client  unexpectedly closed connection while loading  http://192.168.1.126:7979/airsonic/stream?player=1&id=24019  (java.net.SocketTimeoutException)
2019-06-17 14:23:34.104  INFO ---  o.a.p.io.PlayQueueInputStream            : 192.168.1.109: admin  listening to Opus Eponymous/02. Con Clavi Con Dio.flac
2019-06-17 14:23:56.041  INFO ---  o.a.p.c.StreamController                 : 192.168.1.109: Client  unexpectedly closed connection while loading  http://192.168.1.126:7979/airsonic/stream?player=1&id=24019  (java.net.SocketTimeoutException)

fxthomas:
I think I've managed to track down the root cause of this issue. I updated the test repo with some more info about it.

The way I see it, we have 2 options there : wait for the MEJS devs to release a fix and use the workaround for now, or ship a modified version of MEJS that works for us. I'm in favor of the second, in part because the workaround I have is imperfect and still has issues.

jvoisin:
Or, help the MEJS dev to fix the issue, so we can use the upstream version, even if we have to use a snapshot instead of a release.

fxthomas:
Of course that's even better, hoping it's easy to fix without breaking MEJS! :smile:

muff1nman:
Crazy idea - can we rip out mejs?

fxthomas:
That basically amounts to rewriting the player UI. Doable, of course, but not a small amount of work.

muff1nman:
Sorry I deleted my comment (which was "should we scrap me.js") as I had second thoughts about distracting from the issue at hand. But yes I agree that it is a non trivial task. However its also not unsurmountable and might also be better long term.

muff1nman:
But that aside, I think it would be okay to ship a modified me.js. It would also allow us to test it within the larger airsonic context.

Adnn:
Hi,

Encountering a problem on macOS 10.14 with the next song never starting automatically (on both latest Chrome and Firefox), I found some related issues on your github, and might give this PR a try on my server.

I checked the build instructions, but they do not explain how to build the Docker image. Could someone provide the command line to achieve this build?

Thank you for reading,

muff1nman:
@Adnn https://github.com/airsonic/documentation/pull/69/files

jmbannon:
Built the standalone, so far so good! Host is ubuntu 16.04, java 8, Chrome client.

muff1nman:
@fxthomas id like to get this merged - anything I can do to help here?

fxthomas:
I didn't have a lot of time to work on this last week, real life happened ;)

I am confident in the first commit (been running my Airsonic instance with it for a few weeks now), what do you think about creating a new PR with it as a workaround? This PR can stay until we manage to completely fix this in MEJS.

muff1nman:
Why not merge this entire PR? We can fix the MEJS thing later as the firefox AbortError doesn't prevent smooth playback.

Adnn:
After building latest master with fxthomas:try-to-fix-playback-intermittent-stops merged in, I have been listening to a few albums and did not have any problem with next song not playing.

A strange behaviour I am observing is that the last song of the list is repeating (i.e., as soon as it is finished, it restarts from the beginning).

fxthomas:
@Adnn https://github.com/airsonic/airsonic/pull/1080/commits/5e406f0f71a73b1a54e3c5ebd272b54963f81a4e was supposed to fix it, but apparently it's still a bit racey and doesn't work everytime (including for me).

@muff1nman @jvoisin I'll remove the tentative fix for the loop issue @Adnn was talking about as well as the debug logs in the JS console, and we can merge it.

fxthomas:
Okay, ready to merge as soon as Travis finishes testing it!

RedGreen007:
So...as the resident noob to Github, does this mean that I can now go back to the base airsonic...repo? Am I using words right? hahahahahaha

I am still running your test branch and just been following this thread. :p

muff1nman:
It means you can use the master branch now. However there is no release with this change yet so you'd need to build master yourself.

RedGreen007:
Perfect. Thank you 👍

RedGreen007:
Still stopping with Firefox. Works fine in Edge and Chrome. Also got an error while compiling. Had to shut off testing.

jvoisin:
Please do tell us more about this error

RedGreen007:
[ERROR] Failures:
[ERROR]   SearchServiceTestCase.testSearchTypical:237 (15) Specify album 'Ravel - Chamber Music', and get songs. The first song is  expected:<0[1 - Gaspard de la Nuit - i. Ondine]> but was:<0[2 - Gaspard de la Nuit - ii. Le Gibet]>
[INFO]
[ERROR] Tests run: 161, Failures: 1, Errors: 0, Skipped: 0
[INFO]
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for Airsonic 10.4.0-SNAPSHOT:
[INFO]
[INFO] Airsonic ........................................... SUCCESS [  2.579 s]
[INFO] Subsonic REST API .................................. SUCCESS [ 13.869 s]
[INFO] Sonos API .......................................... SUCCESS [ 15.400 s]
[INFO] Airsonic Main ...................................... FAILURE [  01:06 h]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:06 h
[INFO] Finished at: 2019-07-07T15:35:48-07:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.22.0:test (default-test) on project airsonic-main: There are test failures.
[ERROR]
[ERROR] Please refer to /home/pi/airsonic/airsonic-main/target/surefire-reports for the individual test results.
[ERROR] Please refer to dump files (if any exist) [date]-jvmRun[N].dump, [date].dumpstream and [date]-jvmRun[N].dumpstream.
[ERROR] -> [Help 1]
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR]
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
[ERROR]
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <goals> -rf :airsonic-main

I bypassed it by running "mvn clean package -DskipTests"

jvoisin:
Your stacktrace looks completely unrelated to the playback issue.
Please open an other issue about it.

RedGreen007:
I think that worked hahaha. Sorry I just wanted to post it here cause it happened right after the changes got put back into the master. I wasn't sure if it was related. I just hit create new issue and copy and pasted the error there.

fxthomas:
@all, if some of you can go over at #1254 and let me know if you encounter playback issues with it, I would appreciate it. I don't have any on my machines, but additional testers are nice!

PR #1089, opened 2019-06-04
Minor cleanup of integration-test/pom.xml

jvoisin:
- Bump Cucumber's version
- Bump the docker client version
- Remove the now-useless-and-deprecated cucumber parallel plugin


jvoisin:
Meh, the testsuite shouldn't be green if it's failing :/

muff1nman:
It's not failing, rather the changes make it so that it isn't run.

PR #1090, opened 2019-06-05
Show missing Chromecast cover art (at full size)

seldridge:
This fixes a problem where Chromecast cover art was not showing up on the receiver. For whatever reason, the URL that was being generated was right, but the addition of a size specification was resulting in a redirect to the Airsonic login page with an "Incorrect username or password" message.

This removes the size specification (which seems to be a net benefit to me?) and gets the album art working again.

This fixes the second bullet from:
  - https://www.reddit.com/r/airsonic/comments/brf7rm/airsonic_1031/eog2omd/


muff1nman:
Hmm interesting. Ideally I'd like to know why the size was causing issues. Let's leave this open for a bit longer in the case one of us has time to dig a bit deeper.

Issue #1095, opened 2019-06-07
Use inotify to trigger library scan


jooola:
Someone asked me whether his airsonic server could trigger a library scan after changes where made in the music folders. I found inotify which is a tools used to watch change in a directory, and it is embed in the Linux kernel since 2.6 ([thanks wikipedia](https://en.wikipedia.org/wiki/Inotify)).

It would be nice the queue a file scan every time changes are made in the music folders. Of course we should limit the scan to 1 per hour, or we will be scanning the library endlessly for some fast changing libraries. But that would be a nice killer feature.

jvoisin:
Unfortunately, Inotify doesn't work recursively (you have to put watches
at every level of your library), doesn't work for remote filesystems
(like NFS), and is Linux-only :/

Having a monitoring feature would be cool, but I don't see a way to have
one for remote filesystems :/

jooola:
Inotify can work recursively, it depend whether the tool has implemented this feature (watching every sub folders).

Yes, I agree that it only works on linux, I also couldn't get it working using kvm/qemu 9p virtio shared folders.

So I ended up building a bash script for those who has this need. Will open a PR.

But I guess any similar solution could be implemented, maybe a java library that make the required abstraction and use different tools depending on the host OS.

eharris:
I agree that trying to use `inotify` is problematic and probably of too limited scope to be worth the effort within airsonic itself.  I'd much rather see the airsonic have much better auto-detection of changes (as the library is used/browsed), and better scope (re-scanning only the portions of the library that can be detected as having changed).

I've also been thinking about implementing some sort of lazy/incremental scanning solution, that will either continually or within specified timeframes and with very low resource use/impact, scan the parts of the library that are the "oldest" in terms of when they were last scanned.

almet:
I'm not really familiar with the Java ecosystem, but it seems that there already exists [a high level abstraction to do exactly this](https://docs.oracle.com/javase/8/docs/api/java/nio/file/WatchService.html).

PR #1096, opened 2019-06-07
Add contrib/library_autoupdater.sh

jooola:
This script is using inotify to watch your music folders and trigger a new library scan 10 minutes after  a notification has been send from inotify.

Improvement would be very welcomed in future PRs !


jooola:
Related to #1095

Issue #1093, opened 2019-06-06
The videoplayer is broken


jvoisin:
cf1f86f2 broke the video player, by moving the javascript code to an external file, where patterns like `${…}` aren't replaced.

We can either:

1. Revert the commit
2. Declare that since nobody noticed in 7 weeks that it was broken that we can get rid of it
3. Refactor the javascript/jsp so that we can have a working separate javascript file without trivial xss in it, eg by assigning the `model.…` variable to javascript variables, and sanitizing them.

jvoisin:
I'll try to do 3. tonight.

muff1nman:
Either 1 or 3 are fine with me.

jvoisin:
Fixed in  #1097

PR #1097, opened 2019-06-07
Fix the video player that was broken in cf1f86f

jvoisin:


jvoisin:
This should fix #1093

PR #847, opened 2018-12-11
Update dependencies - DO NOT MERGE

randomnicode:
This PR is just testing grounds. Subsequent PRs will break up most of the changes here into more meaningful chunks.

Updates some libs.

The major ones are:
- JDOM -> JDOM2 (mostly import changes)
- EHCache 2.5 -> 3.2.3 (differences with tiered cache mechanisms, mostly localized to CacheFactory)
- Lucene 3 -> 7.5 (some api changes, mostly localized to SearchService)
- JFreeChart (some api changes)

PR #1113, opened 2019-06-09
WIP: Lucene update

jvoisin:
This PR is heavily based on #847, some might even call it a complete rip-off.

### DIFF
--- a/airsonic-main/pom.xml
+++ b/airsonic-main/pom.xml
@@ -148,12 +148,6 @@
         </dependency>
 
         <dependency>
-            <groupId>org.apache.lucene</groupId>
-            <artifactId>lucene-queries</artifactId>
-            <version>7.5.0</version>
-        </dependency>
-
-        <dependency>
             <groupId>org.apache.ant</groupId>
             <artifactId>ant</artifactId>
             <scope>runtime</scope>
--- a/airsonic-main/pom.xml
+++ b/airsonic-main/pom.xml
@@ -131,7 +131,26 @@
         <dependency>
             <groupId>org.apache.lucene</groupId>
             <artifactId>lucene-core</artifactId>
-            <version>3.0.3</version>
+            <version>7.5.0</version>
+        </dependency>
+
+
+        <dependency>
+            <groupId>org.apache.lucene</groupId>
+            <artifactId>lucene-analyzers-common</artifactId>
+            <version>7.5.0</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.lucene</groupId>
+            <artifactId>lucene-queryparser</artifactId>
+            <version>7.5.0</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.lucene</groupId>
+            <artifactId>lucene-queries</artifactId>
+            <version>7.5.0</version>
         </dependency>
 
         <dependency>
--- a/airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
@@ -172,7 +172,7 @@ public class MediaScannerService {
         try {
 
             // Maps from artist name to album count.
-            Map<String, Integer> albumCount = new HashMap<String, Integer>();
+            Map<String, Integer> albumCount = new HashMap<>();
             Genres genres = new Genres();
 
             scanCount = 0;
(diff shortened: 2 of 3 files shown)


tesshucom:
On my clone server it has already been updated to 7.7.1 and is used by Some Japanese people.
It might be a partial help.

(Because of the long Japanese language processing, I was forced to change the class structure for unit testing.
 In addition, processing different from Airsonic such as multi genre and purge processing of unnecessary data is included.
So partial....)

[main](https://github.com/tesshucom/jpsonic/tree/master/jpsonic-main/src/main/java/com/tesshu/jpsonic/service)
[test](https://github.com/tesshucom/jpsonic/tree/master/jpsonic-main/src/test/java/com/tesshu/jpsonic/service)

Due to Lucene's destructive updates, there are some syntactic differences.
ex)
How to write a logical expression is different.
```
        BooleanQuery.Builder subMusicFoldersQuery = new BooleanQuery.Builder();
        musicFolders.forEach(musicFolder -> {
            if (indexType == IndexType.ALBUM_ID3 || indexType == IndexType.ARTIST_ID3) {
                subMusicFoldersQuery.add(new TermQuery(new Term(FieldNames.FOLDER_ID, musicFolder.getId().toString())), Occur.SHOULD);
            } else {
                subMusicFoldersQuery.add(new TermQuery(new Term(FieldNames.FOLDER, musicFolder.getPath().getPath())), Occur.SHOULD);
            }
        });
        mainQuery.add(subMusicFoldersQuery.build(), Occur.MUST);
```


null is no longer allowed.
```
        if (!(isEmpty(criteria.getFromYear()) && isEmpty(criteria.getToYear()))) {
            query.add(IntPoint.newRangeQuery(FieldNames.YEAR, 
                isEmpty(criteria.getFromYear())
                    ? Integer.MIN_VALUE
                    : criteria.getFromYear(),
                isEmpty(criteria.getToYear())
                    ? Integer.MAX_VALUE :
                    criteria.getToYear()),
                Occur.MUST);
        }
```
Also, the field types used in this PR document definition may be different from mine.

Migration is possible.
But I do not know if there is a motive that is commensurate with hardship...
In my case I needed Japanese.

jvoisin:
This is a bit off-topic, but why aren't you trying to upstream more of your changes in airsonic?

tesshucom:
I apologize if I am a guilty.
However, I did not mean that I could estimate time and risk in advance.
I just took a lot of trial and error stacks.

It often seems that Lucene's topic has risen but it has always been negative.
Therefore I tried it personally.
As a result I feel that it contains some areas that are likely to be controversial.

fxthomas:
I, for one, welcome better support for complex scripts, it's currently a bit painful to search for Japanese stuff... I'd be glad to test and include this if you can make a PR! :)

jvoisin:
Me too !

tesshucom:
I also think that way.
And It may also be necessary to introduce some elements at an appropriate time to approach it.


- Divide into several classes for each function
- Improvement of index version control (generation rotate mechanism) which is dead

What about this? I suggest.
Class division seems to be useful.
At least the current "one class configuration" was a barrier to the lucene update.

- I want to avoid destructive updates.
- There is no test, so I want to add a test.
- Tests are difficult to add because the features are combined too much.

It is the state now.

Although generational changes seem irrelevant, it reduce the burden on users at the time of release.
___

I found this link in past issues.
[Subsonic utilise Elasticsearch comme base de données orientée documents](https://www.elastic.co/fr/blog/subsonic-utilise-elasticsearch-comme-base-de-donnees-orientee-documents#)
(The author is French and uses French.)
Some countries other than English have such ideas.

I do not want the Airsonic to be complicated at all, and Multilingual support is an example of an anti-pattern for a specific purpose.
Making Airsonic multilingual is difficult and should not be the immediate goal.
However, design improvements can also reduce the burden of localization.

It is sad, the latest version of Subsonic has evolved so cleverly that it is impossible to search Japanese...

jvoisin:
Splitting the file is a good start, I do agree :)

tesshucom:
I will create a PR from now on as a suggestion.
Because Analyzer is different, it may take some time.

I mainly use JapaneseAnalyzer.
What is required of Airsonic is [StandardAnalyzer](https://lucene.apache.org/core/7_7_1/core/org/apache/lucene/analysis/standard/StandardAnalyzer.html).
The Analyzer currently used by Airsonic and the parsing of the StandardAnalyzer should be slightly different.

It is because [UAX#29](https://unicode.org/reports/tr29/) was introduced.
The handling of underbars and single quotes changes slightly, but it is faster to include the results in a test case.
French is a little sensitive.

fxthomas:
If I understand it correctly, if we start using `StandardAnalyzer`, replacing it with `JapaneseAnalyzer` is easy, right?

tesshucom:
That's right but, my clone contains minor spec changes as well as Lucene updates and localization.
It is radical to introduce everything at the same time.

I think that it is better to create a minimum PR according to the Guidelines for Contributing.
 - Split SearchService
 - Lucene update

These will improvement make other improvements very easy and safety.
Other than that, I think that judgment is necessary individually.

PR #1116, opened 2019-06-10
Refactor transcoding/downsampling bitrate limits

### LINKED ISSUE #867: Max bitrate setting "No limit" is not accurate
## Problem description

There is one option of max bitrate in the user setting page "No limit". I checked the source code and believe the actual meaning of it should be "Default". See the code starting here:
https://github.com/airsonic/airsonic/blob/adc224147a6266d729af8a5fa8d7f11f58598676/airsonic-main/src/main/java/org/airsonic/player/service/TranscodingService.java#L199

The value of option "No limit" is 0 which is TranscodeScheme.OFF. Then the actual bitrate will be set to a default value later here:
https://github.com/airsonic/airsonic/blob/adc224147a6266d729af8a5fa8d7f11f58598676/airsonic-main/src/main/java/org/airsonic/player/service/TranscodingService.java#L208

eharris:
Previously, transcodes just used default bitrate limits which might
either be too low (degrading the quality unnecessarily), or would be too
high which wasted bandwidth.  Fixes #867
This also now takes VBR sources into account, and gives them a little
extra bitrate when transcoding to better reproduce the quality of the
source.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/AvatarUploadController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/AvatarUploadController.java
@@ -118,7 +118,7 @@ public class AvatarUploadController  {
 
             // Scale down image if necessary.
             if (width > MAX_AVATAR_SIZE || height > MAX_AVATAR_SIZE) {
-                double scaleFactor = (double) MAX_AVATAR_SIZE / (double) Math.max(width, height);
+                double scaleFactor = MAX_AVATAR_SIZE / (double) Math.max(width, height);
                 height = (int) (height * scaleFactor);
                 width = (int) (width * scaleFactor);
                 image = CoverArtController.scale(image, width, height);
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/StatusChartController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/StatusChartController.java
@@ -134,7 +134,7 @@ public class StatusChartController extends AbstractChartController {
 
         XYItemRenderer renderer = plot.getRendererForDataset(dataset);
         renderer.setSeriesPaint(0, Color.blue.darker());
-        renderer.setSeriesStroke(0, new BasicStroke(2f));
+        renderer.setSeriesStroke(0, new BasicStroke(2.0f));
 
         // Set theme-specific colors.
         Color bgColor = getBackground(request);
--- a/airsonic-main/src/main/java/org/airsonic/player/service/jukebox/PlayerTest.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/jukebox/PlayerTest.java
@@ -48,7 +48,7 @@ public class PlayerTest implements AudioPlayer.Listener {
         });
         gainSlider.addChangeListener(new ChangeListener() {
             public void stateChanged(ChangeEvent e) {
-                float gain = (float) gainSlider.getValue() / 1000.0F;
+                float gain = gainSlider.getValue() / 1000.0F;
                 player.setGain(gain);
             }
         });
--- a/airsonic-main/src/main/java/org/airsonic/player/util/StringUtil.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/util/StringUtil.java
@@ -200,25 +200,25 @@ public final class StringUtil {
         // More than 1 TB?
         if (byteCount >= 1024L * 1024 * 1024 * 1024) {
             NumberFormat teraByteFormat = new DecimalFormat("0.00 TB", new DecimalFormatSymbols(locale));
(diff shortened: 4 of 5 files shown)


airsonic-main/src/main/java/org/airsonic/player/service/TranscodingService.java
```diff
@@ -202,15 +202,32 @@ public Parameters getParameters(MediaFile mediaFile, Player player, Integer maxB
 
         boolean hls = videoTranscodingSettings != null && videoTranscodingSettings.isHls();
         Transcoding transcoding = getTranscoding(mediaFile, player, preferredTargetFormat, hls);
-        if (transcoding != null) {
-            parameters.setTranscoding(transcoding);
+        Integer bitRate = mediaFile.getBitRate();
+
+        if (mediaFile.isVideo()) {
+            if (bitRate == null) {
+                bitRate = 20000;
```
jvoisin: Wouldn't it make sense to use `VideoPlayerController.DEFAULT_BIT_RATE` instead of a magic constant?

eharris: That is not the maxBitRate, that is the bitRate for the media file itself.  Hopefully it is never null, but I didn't want to allow for that possibility or have to keep checking and handling null values, so this just forces it to be a very large value.

Note that the `VideoPlayerController.DEFAULT_BIT_RATE` will be used since it is less than this large value.

eharris: I guess I should note that the reason I did it this way is that I am planning for future changes that will allow downsampling of video and other types of media files rather than solely mp3 files.

airsonic-main/src/main/java/org/airsonic/player/service/TranscodingService.java
```diff
@@ -202,15 +202,32 @@ public Parameters getParameters(MediaFile mediaFile, Player player, Integer maxB
...
             if (maxBitRate == null) {
-                maxBitRate = mediaFile.isVideo() ? VideoPlayerController.DEFAULT_BIT_RATE : TranscodeScheme.MAX_192.getMaxBitRate();
+                maxBitRate = VideoPlayerController.DEFAULT_BIT_RATE;
+            }
+        } else {
+            if (bitRate == null) {
+                bitRate = TranscodeScheme.MAX_320.getMaxBitRate();
+            } else if (mediaFile.isVariableBitRate()) {
+                // Assume VBR needs approx 20% more bandwidth to maintain equivalent quality in CBR/ABR
+                bitRate = Math.min(bitRate * 6/5, 320);
```
muff1nman: I'm not following on the logic here. Can you please expand?

airsonic-main/src/main/java/org/airsonic/player/service/TranscodingService.java
muff1nman: If bitRate is null why even try to set a default? bitRate being null would imply that we don't know what the bitRate of the media file is. In that case the existing behavior of never downsampling seems better than guessing whether to transcode or not.

eharris: I didn't want to preclude the possibility that airsonic may not know the bitrate in some circumstances (perhaps a media type that Airsonic doesn't know how to detect bitrate but ffmpeg can still transcode?), so I just decided to assume it is max bitrate and let the transcoder fail if it needs to rather than prematurely refusing to do what was asked.

airsonic-main/src/main/java/org/airsonic/player/service/TranscodingService.java
eharris: I'm trying to preserve as much quality as possible when transcoding (unless the bitrate is limited by other settings).  Since VBR files can encode higher quality content at a lower average bitrate, I assume that a CBR/ABR transcode needs 20% more bandwidth to make sure that complex sections of audio don't run out of bits and end up lower quality.  But there's also an upper limit of 320kbps.

Granted the 20% figure is still just an approximation, but I based it on the bitrate comparisons here: https://trac.ffmpeg.org/wiki/Encode/MP3

eharris:
I may need to adjust this to quantize the bitrates to "standard" values, as it seems that `ffmpeg` with standard CBR encoding doesn't allow "in-between" values (e.g. 173kbps) and rounds up to the next standard rate (passing 175kbps yields a stream at 192kbps) which then causes the track to get cut off early if range-enabled transfers for transcodes are allowed.

This is unfortunate, because it would be nice to allow for a continuous range in bitrate since it's only CBR encoding that doesn't.  But it could also be handled by way of an `ffmpeg` wrapper that rounded down to the closest standard bitrate rather than up.

eharris:
I moved this to a branch so it can be updated.  The new location is here:
https://github.com/eharris/airsonic/tree/bitrate-limits

eharris:
The new PR is #1124

Issue #1098, opened 2019-06-07
Wtf is /rest/videoPlayer.view


jvoisin:
1. Why do we have a video player in the rest API?
2. Why is it using Flash (and some old-school javascript libraries along with it)?
3. Can we get rid of it?

To play videos, there is already the decently-working `/videoPlayer.view` page.

jooola:
I'm more a make a one thing and make it right kind of guy. So I would even totally get rid of the video player. Other tools are doing video streaming way better than *sonic.

We already made chat api endpoint 'Gone', and some new video endpoint 'Not implemented'. Same could be applied here.

https://airsonic.github.io/docs/api/

heyarne:
I'd support that as well: https://github.com/airsonic/airsonic/issues/921#issuecomment-480834513

jvoisin:
`/rest/videoPlayer` isn't a subsonic API endpoint.

jvoisin:
@muff1nman I would like to have your opinion on this one :)

heyarne:
So if I understand this correctly it iseems to be used by DSub, see [here](https://github.com/daneren2005/Subsonic/search?q=getVideoUrl&unscoped_q=getVideoUrl). The actual implementation is in [RESTMusicService.java](https://github.com/daneren2005/Subsonic/blob/166d395e3e5de223527fec96d8b17dd42ed9a991/app/src/main/java/github/daneren2005/dsub/service/RESTMusicService.java#L892), even though there are also methods to defer to `stream` and `hls`.

I can't test this right now, but my guess would be that it might

a. not be used directly
b. broken if it's used to embed a fragment of HTML that uses `swf` to play videos.

muff1nman:
It looks like this method wasn't migrated correctly when the REST api got migrated. See 5b2297af968dfe903b45170e8204d81116f1de80 for an example of how the other methods were migrated. Its also interesting that this method isn't documented in the subsonic api.

jvoisin:
Can we delete it?

muff1nman:
It still might be useful for apps like dsut, whats broken about it other than it not being annotated correctly?

jvoisin:
It's using flash, so I don't think that it's working on DSub :D

muff1nman:
> **muff1nman**
jvoisin: where do you see flash in the videoPlayer method in SubsonicRESTController?
**jvoisin (@freenode_jvoisin:matrix.org)**
[...]

PR #1118, opened 2019-06-11
Remove the ghetto REST flash-based video player

jvoisin:
Remove the terrible REST flash-based video player. This includes flash files, related javascript, removing the controller, the jsp file, and purging every single trace of Flash-related code from airsonic!


jvoisin:
This should close #1098

airsonic-main/src/main/java/org/airsonic/player/security/GlobalSecurityConfig.java
```diff
@@ -162,7 +162,7 @@ protected void configure(HttpSecurity http) throws Exception {
                     .and().authorizeRequests()
                     .antMatchers("/recover*", "/accessDenied*",
                             "/style/**", "/icons/**", "/flash/**", "/script/**",
-                            "/sonos/**", "/crossdomain.xml", "/login", "/error")
```
muff1nman: This is unrelated to this PR, correct?

jvoisin: Nope, it's related: since we're not using Flash anymore, we can remove the crossdomain.xml file.

eharris:
@jvoisin Thanks for this!  But it looks like the internationalization strings still reference the "Flash" player, which should probably be changed to something more generic like "Media" or maybe HTML5?

jvoisin:
Done in 2d06995416f589ad9664b4ca1d03963326b3ca0d

PR #1123, opened 2019-06-12
Change back to allow transcodes to allow seeking

### LINKED ISSUE #1117: Reverting patch broke ability for web player to seek within transcoded streams
## Problem description

Commit 3efa2d3 (reverting an earlier feature commit) broke the ability for the web player to seek within a transcoded stream.

### Steps to reproduce

1. Install v10.3.1, and enable `Enable seeking on transcodes` in the Transcoding settings page.
2. Play a file that gets transcoded or downsampled using the web player.
3. Observe that the seek bar works, and can seek within the content that has already been received.
4. Build and install the current master branch which contains the above commit and do the same.
5. Note that the seek bar in the web player doesn't work anymore, just shows `Live Broadcast` where the seek bar should be.

## System information

 * **Airsonic version**: 10.4.0-SNAPSHOT
 * **Operating system**: Debian 9 stretch
 * **Java version**: java-8-openjdk
 * **Proxy server**: Apache builtin
 * **Client**: Chrome for OSX Version 74.0.3729.131 (Official Build) (64-bit)

## Additional info

The original patch was #798.  The predecessor was #759.

eharris:
This switches back to allowing range rather than chunked requests so that
seeking works in the web player, but does so by safer means than previous
solutions, by slightly over-estimating the transcoded size, then sending
dummy bytes to the client to fill any gap.  Fixes #1117.  Addresses #685.


jvoisin:
@tari @muff1nman any opinion on this, since you were the ones discussing it in #1117

tari:
As I said there, because this estimate can still be wrong I'm not comfortable making it the only option. Reintroducing the "allow seeking" user-facing option and changing transcode-with-seek behavior (when enabled by the user) to do this seems okay, though I'm not sure how player in general would handle having garbage inserted at the end of a file.

eharris:
I'm kind of holding off on making more changes to this until #1124 gets merged, because they touch some common parts of the code.  And I have further changes planned that will make this much safer even without a configuration option to enable it.

As to how players handle extra data tacked on to pad out the transfer to the proper length, I kind of disagree with the term "garbage", and would point out that the code already did that in some circumstances for streaming sources, so appears to be at least somewhat of a time-tested mechanism.

I should also point out that the change that broke seeking for transcodes is a very recent change that probably hasn't seen that much exposure to the user community at large.  I suspect that there might be a decent amount of fallout from the loss of that capability and its other side effects.

tari:
The existing practice of inserting garbage into streams has also made me nervous when looking at the code, but I haven't touched it because it appears to work and I haven't been interested in that use case.

I agree that people will probably miss seeking, but I know people will also miss having streams that work correctly. Not all users will care about both though, so an option seems to make sense. Ultimately I'd hope there is a more correct and integrated solution, but the existing change was done this way because it's not possible to fix known problems to a high degree of reliability without making this tradeoff.

eharris:
@tari will you check if my last commit is acceptable to you as far as attempting to ensure safety when allowing seeking in transcodes, without the need for a user configurable option?

tari:
That's an interesting heuristic. I'm still not really happy with this solution, but it seems okay enough to work around what is mostly an unfortunate set of assumptions in the original API design that we're going to be stuck with (since we don't control all the clients). Maybe eventually we can adopt DASH as a backwards-incompatible streaming solution, and support the existing API by spooling transcoded files in their entirety to temporary storage so we can always know the exact size- trading latency (presumably less important at that point) for accuracy.

The real test will be running this for a while and seeing if it does things wrong, but I'm okay with giving it a try.

eharris:
@tari Thanks for being willing to try this solution.  I also just added a commit to add a warning if the stream output exceeds the expected length, so we can have some indicator when that happens in case it causes problems for players.

I think this PR is ready to be merged.

eharris:
@tari Also, after this gets merged I would be willing to take a stab at addressing your stream padding concerns by working on a method to append actual valid silence frames (in the appropriate format for the output stream) rather than the current method.  I'm not sure how that will work when a silence frame may be truncated at the end of the stream, but hopefully players will see a truncated frame as invalid and will ignore it.

airsonic-main/src/main/java/org/airsonic/player/service/TranscodingService.java
```diff
@@ -485,6 +487,59 @@ private boolean isTranscodingStepInstalled(String step) {
...
+
+        // Verify that were able to predict the length
+        if (parameters.getExpectedLength() == null) {
+            return false;
+        }
+
+        // Check if last configured step uses the bitrate, if so, range should be pretty safe
+        for (String step : steps) {
+            if (step != null) {
+                return step.contains("%b") ? true : false;
```
jvoisin: `return step.contains("%b")` would be enough I think.

airsonic-main/src/main/java/org/airsonic/player/service/TranscodingService.java
```diff
@@ -485,6 +487,59 @@ private boolean isTranscodingStepInstalled(String step) {
...
+            return null;
+        }
+
+        if (maxBitRate == null) {
+            LOG.error("Unknown bit rate for " + file + ". Unable to estimate transcoded size.");
+            return null;
+        }
+
+        // Over-estimate size a bit (2 seconds) so don't cut off early in case of small calculation differences
+        return (duration + 2) * (long)maxBitRate * 1000L / 8L;
```
jvoisin: Please don't use magic constants like `1000L / 8L` without at least adding a comment :)

airsonic-main/src/main/java/org/airsonic/player/service/TranscodingService.java
eharris: You're right, will correct that.

airsonic-main/src/main/java/org/airsonic/player/service/TranscodingService.java
eharris: That function was just moved from `StreamController`, I didn't write it... :-)

fxthomas:
Didn't Airsonic previously truncate the duration up to the second, assume CBR and discard extra bytes in the output? It was working reasonably well, on both the web player and mobile clients (DSub).

Another thing : I personally don't care much about this, but how does overestimating the length work with players offering gapless playback (DSub does as far as I know)?

fxthomas:
So far so good, I've been testing this over the last few weeks and everything appears to be working well.

fxthomas:
Have my tentative approval, I saw no issues running this on my server.

I'll try to give it a full review of the code itself, but might not be able to do that before next week, so if somebody else can review and approve you can go ahead and merge.

tari:
It's been a while now, but IIRC this seemed to work okay for me too.

airsonic-main/src/main/java/org/airsonic/player/controller/StreamController.java
```diff
@@ -231,25 +233,37 @@ public void handleRequest(HttpServletRequest request, HttpServletResponse respon
...
                 byte[] buf = new byte[BUFFER_SIZE];
+                long bytesWritten = 0;
 
                 while (!status.terminated()) {
                     if (player.getPlayQueue().getStatus() == PlayQueue.Status.STOPPED) {
                         if (isPodcast || isSingleFile) {
                             break;
                         } else {
-                            sendDummy(buf, out);
+                            sendDummyDelayed(buf, out);
```
fxthomas: Out of curiousity, any idea what the goal of sleeping here is? It's probably inherited from Subsonic and I'm probably missing something, but `Thread.sleep` feels like a hacky workaround...

fxthomas:
Ready to merge as soon as the test failures are handled :smile:

almet:
Integration tests are failing, and from my understanding, it's only due to the fact that they should be updated : they are currently checking that the sent bytes are equal to the actual ones, but as this very pull request is made against this idea… Integration tests should be updated.

I believe that a new `Then response bytes are larger` test should be added in `/integration-test/src/test/java/org/airsonic/test/cucumber/steps/api/StreamStepDef.java`.

fxthomas:
I'd really like to merge that in the next release, so I added the 10.6.0 milestone. I'll update that if we decide go to 11.0 instead!

airsonic-main/src/main/java/org/airsonic/player/controller/StreamController.java
eharris: I think originally that code was meant to keep the stream/connection alive while waiting for something else to occur (but I'm not really sure what that was).

fxthomas: Well, in any case I don't see why we shouldn't keep it here for now until we figure out what this means.

eharris:
I fixed the checkstyle issues and the transcoding tests by updating the compared byte sequences and adding back the check that the length headers match.  I decided not to try to change the tests to allow a non-specified amount of "additional padding" bytes since that just makes things more convoluted (and probably also overly permissive) when the comparison byte sequences would most likely need to be generated by the transcoding output code path anyway.

As mentioned previously, an even better solution would be to have the range requests only be allowed by configuration for the specific players that need them.  That probably should have been done as part of the original change rather than just removing the range capability for transcodes completely, but this PR at least gets us back to the expected prior behavior in a somewhat safer manner and avoids most of the undesired side-effects caused by that change.  Making those improvements is left for a later PR.

fxthomas:
Merged, thank you for the help!

Issue #894, opened 2019-03-01
10.2.1 - Missing FFMPEG causes "Transcoded" MP3s the size of original FLACs


4oo4:
## Problem description

I recently upgraded to 10.2.1 and at some point accidentally uninstalled ffmpeg, in the past when I haven't had it installed or symlinked to the correct place, I remember that the player would just freeze and refuse to play the file. However, I accidentally did this again, but Airsonic ended up downloading a 100+ MB file for what was supposed to be an MP3 of a 1 minute track.

### Steps to reproduce

1. Break the symlink in `/var/airsonic/transcode/ffmpeg` (or uninstall ffmpeg)
2. Attempt to play a high-resolution FLAC file in the Ultrasonic app, and download will be very slow and create very large files that are still labeled as MP3. Somehow it's still a playable MP3 file though?

## System information

 * **Airsonic version**: 10.2.1
 * **Operating system**: Ubuntu 16.04
 * **Proxy server**: nginx 1.15.8
 * **Client**: Ultrasonic Android App

Cheers

eharris:
This is kind of as-designed.  If the transcoder isn't installed, the code falls back to serving the original file, on the assumption that trying to play something is better than nothing.  The transcoder failure and fallback is also noted in the logs, and it is up to the server administrator to make sure that things are configured and working properly.

That being said, if a bitrate limit is configured, and is not being heeded, then that is a problem.  Other issues similar to this are being worked on as part of #1124.

eharris:
Closing this since the fallback functionality is working as designed.

4oo4:
@eharris Thanks, good to know. I think what confused me was that when it was serving the original file, which was a FLAC, the file extension was still MP3.

PR #1124, opened 2019-06-14
Refactor transcoding/downsampling bitrate limits

### LINKED ISSUE #867: Max bitrate setting "No limit" is not accurate

eharris:
Previously, transcodes just used a hardcoded default bitrate which might either be too low (degrading quality unnecessarily), or too high (wasting bandwidth unnecessarily).  Fixes #867

Because the transcoding bitrate (with these changes) now gets set to approximately the same bitrate as the source, whether the source is VBR or not is taken into account and gives them a little extra bitrate when transcoding to better approximate the quality of the source.  This won't necessarily perfectly account for different codecs that may be better quality with the same bitrate (e.g. m4a vs. mp3), but it should be reasonably close.

It should be noted that the reason I made the changes the way I have is that I am planning for future changes that will allow downsampling of video and other types of media files rather than solely mp3 files.


eharris:
Looks like the tests need to be changed/fixed for this one.

eharris:
It should be noted that this change is especially helpful for people listening from mobile devices, as it can significantly reduce the data use (and cellular data plan impact) for certain types of content (e.g. audio books).

Since this patch allows the server to automatically use a lower bitrate based on the source bitrate, on my mobile devices this means even though I restrict my bitrate to 128mbps for music when on cellular data, if I happen to listen to an audiobook, this lets it it automatically use a lower (e.g. 64kbps) bitrate, which can really help me save data and allow me to do more listening and worry about my usage less.

eharris:
I'm probably going to need help getting the `travis-ci` tests fixed, since I don't have access to that, and the transcode outputs need to be byte-for-byte identical, which means they depend on the version of `ffmpeg` and libs that are installed.

eharris:
Since this branch hadn't been merged yet, I went ahead and added some more transcoding cleanup. 
With these changes there is now no longer a need for a separate config for Downsampling, which I've always thought was confusing because it wasn't that obvious that it only applied for mp3->mp3.  And I never liked having a hardcoded special case for mp3 anyway.

So I've completely removed all references to downsampling, it's all just transcoding, and airsonic is now smart enough to transcode only if it needs to, either because the format needs to be changed or the bitrate needs to be reduced.

@muff1nman @jooola @jvoisin @fxthomas please review, as I've got other changes waiting that touch some of the same areas of code, and I don't want to keep adding to this PR, or have to have someone deal with merge conflicts.  Thanks!

eharris:
Fixed a bug related to high bitrate sources (flac), and got rid of arbitrary value.

eharris:
Note that the reason the travis-ci flac streaming test is failing is due to the file not being transcoded.  This is intentional (and correct) because the client is not requesting a specific format and there is no bitrate limit in effect, so the code now properly delivers the file without transcoding.

airsonic-main/src/main/java/org/airsonic/player/domain/TranscodeScheme.java
```diff
@@ -105,7 +105,7 @@ public static TranscodeScheme valueOf(int maxBitRate) {
 
     public static TranscodeScheme fromMaxBitRate(int maxBitRate) {
         for (TranscodeScheme transcodeScheme : TranscodeScheme.values()) {
-            if (maxBitRate == transcodeScheme.getMaxBitRate()) {
```
jvoisin: Don't we want to get the closest number fo `maxBitRate`?

airsonic-main/src/main/java/org/airsonic/player/service/TranscodingService.java
```diff
@@ -195,27 +193,46 @@ public Parameters getParameters(MediaFile mediaFile, Player player, Integer maxB
...
             }
-        } else if (maxBitRate != null) {
-            boolean supported = isDownsamplingSupported(mediaFile);
-            Integer bitRate = mediaFile.getBitRate();
-            if (supported && bitRate != null && bitRate > maxBitRate) {
-                parameters.setDownsample(true);
+        } else {
+            if (mediaFile.isVariableBitRate()) {
+                // Assume VBR needs approx 20% more bandwidth to maintain equivalent quality in CBR
+                bitRate = bitRate * 6/5;
```
jvoisin: Nit: why `6/5` and not `1.2f` ?

airsonic-main/src/main/java/org/airsonic/player/service/TranscodingService.java
```diff
@@ -195,27 +193,46 @@ public Parameters getParameters(MediaFile mediaFile, Player player, Integer maxB
...
+            if (TranscodeScheme.fromMaxBitRate(bitRate) != null) {
+                bitRate = TranscodeScheme.fromMaxBitRate(bitRate).getMaxBitRate();
+            }
+        }
+
+        if (maxBitRate == 0 || (bitRate != 0 && bitRate < maxBitRate)) {
+            maxBitRate = bitRate;
+        }
+
+        if (transcoding != null && ((maxBitRate != 0 && (bitRate == 0 || bitRate > maxBitRate)) ||
```
jvoisin: This could likely be simplified via ` else` related to the condition above.

eharris: It can't because it should happen in both cases that the above if handles, either video or audio files.

airsonic-main/src/main/java/org/airsonic/player/service/TranscodingService.java
eharris: This is doing integer math because the operand is an integer, not floating point.

airsonic-main/src/main/java/org/airsonic/player/domain/TranscodeScheme.java
eharris: Yes, and that's what it does, it gets the closes scheme that is less than or equal to `maxBitRate`

airsonic-main/src/main/java/org/airsonic/player/service/TranscodingService.java
eharris: Also some of the values that it uses are modified within the previous if, which means it would be testing the wrong values.

eharris:
I agree that some transcoding tests would be good, but since I don't have access to the travis instance where the tests are run, I can't be sure that any changes I would make would work properly without a lot of trial and error checkins.  I'd prefer to leave that to someone who has access so as not to clutter up the PR with a bunch of test commits.

muff1nman:
@eharris you don't need access to travis to run the tests. You can run all of the tests using: `mvn verify  -P integration-test` as seen in the `.travis.yml` file.

eharris:
I discovered a problem that doesn't seem to affect operation in normal usage, but is nonetheless wrong, which is that the reported `Content-Type` is wrong in some circumstances.  This is due to some parts of the code making incorrect assumptions (and arriving at a different conclusion) about whether or not a file will ultimately be transcoded.

Unfortunately, fixing the pieces of code that results in the incorrect `Content-Type` problem is not trivial, but I'm working on it.  I don't think this should be merged until I have a fix in place.

eharris:
#894 points out that in the event of a transcoder failure, if the resulting fallback (original) content would exceed a configured bitrate limit, then the content should not be served.

eharris:
I updated this PR to resolve the merge conflicts introduced in mainline since this PR was created, but travis is currently broken due to tomcat-embed-core-8.5.43.jar: CVE-2020-1938

fxthomas:
I understand that this PR removes the need for a dedicated downsampling command because it's now setting the bitrate limits inside the transcoding commands. Is that correct?

Gonna test-drive it on my instance, will report back if anything happens!

airsonic-main/src/main/java/org/airsonic/player/service/TranscodingService.java
```diff
@@ -195,27 +193,46 @@ public Parameters getParameters(MediaFile mediaFile, Player player, Integer maxB
...
+            if (mediaFile.isVariableBitRate()) {
+                // Assume VBR needs approx 20% more bandwidth to maintain equivalent quality in CBR
+                bitRate = bitRate * 6 / 5;
             }
-        } else if (maxBitRate != null) {
-            boolean supported = isDownsamplingSupported(mediaFile);
-            Integer bitRate = mediaFile.getBitRate();
-            if (supported && bitRate != null && bitRate > maxBitRate) {
-                parameters.setDownsample(true);
+            // Make sure bitrate is quantized to valid values for CBR
```
fxthomas, line 220: Not sure I understand how the quantization works here?

airsonic-main/src/main/java/org/airsonic/player/service/TranscodingService.java
```diff
@@ -195,27 +193,46 @@ public Parameters getParameters(MediaFile mediaFile, Player player, Integer maxB
...
+        }
+
+        if (mediaFile.isVideo()) {
+            if (maxBitRate == 0) {
+                maxBitRate = VideoPlayerController.DEFAULT_BIT_RATE;
+            }
+        } else {
+            if (mediaFile.isVariableBitRate()) {
+                // Assume VBR needs approx 20% more bandwidth to maintain equivalent quality in CBR
+                bitRate = bitRate * 6 / 5;
```
fxthomas, line 218: Fair enough, without a first pass it's impossible to determine in advance. Might need a note in the docs though.

airsonic-main/src/main/java/org/airsonic/player/service/TranscodingService.java
eharris: The main purpose of imposing a bitrate limit is to control bandwidth usage.  To do this, most codecs (like mp3) have to be used in CBR rather than VBR mode, which only allows specific quanta for bitrate values (e.g. values like 32, 56, 64, 96, 128kbps).  These values are provided in `domain/TranscodeScheme.java`.

airsonic-main/src/main/java/org/airsonic/player/service/TranscodingService.java
eharris: Yes, updating the docs would be good.  Happy to create a PR to do that once this is merged (finalized).

airsonic-main/src/main/java/org/airsonic/player/service/TranscodingService.java
fxthomas: Thanks! ;)

fxthomas:
Good for me, seems to be working just fine after spending 8 hours listening to music of various formats :smile:

PR #1129, opened 2019-06-15
Disallow deleting your own user or removing admin role

eharris:
It makes no sense and is potentially dangerous to allow a user to delete their own account, or to be able to remove the admin role from themselves, so this removes the ability to do so.


muff1nman:
Shouldn't there also be a server side check here as well?

eharris:
@muff1nman I added validation to reject these on the server side too.  Didn't bother with translations for the error messages since users won't even be given the option with the prior changes.

Issue #1121, opened 2019-06-11
Do we want to add more checks in checkstyle?


jvoisin:
Currently, we're only using [RedundantImport]( http://checkstyle.sourceforge.net/config_imports.html#RedundantImport ) and UnusedImports in checkstyle, but there are a lot of other [interesting checks]( http://checkstyle.sourceforge.net/checks.html ) available, like:

- [EmptyLineSeparator]( https://checkstyle.sourceforge.io/config_whitespace.html#EmptyLineSeparator ) — #1196
- [EqualsAvoidNull]( http://checkstyle.sourceforge.net/config_coding.html#EqualsAvoidNull ) — #1183
- [EqualsHashCode](http://checkstyle.sourceforge.net/config_coding.html#EqualsHashCode) — 167d1d9c2071adb84dfe61347f6197847b0dca6f
- [FileTabCharacter]( https://checkstyle.sourceforge.io/config_whitespace.html#FileTabCharacter ) — #1206
- [FinalLocalVariable]( https://checkstyle.sourceforge.io/config_coding.html#FinalLocalVariable )
- [Indentation]( https://checkstyle.sourceforge.io/config_misc.html#Indentation ) — #1194
- [MethodParamPad]( https://checkstyle.sourceforge.io/config_whitespace.html#MethodParamPad ) — #1197
- [ModifiedControlVariable]( https://checkstyle.sourceforge.io/config_coding.html#ModifiedControlVariable )
- [NoWhiteSpaceBefore]( https://checkstyle.sourceforge.io/config_whitespace.html#NoWhitespaceBefore ) — #1197
- [OneStatementPerLine]( https://checkstyle.sourceforge.io/config_coding.html#OneStatementPerLine ) — #1201
- [ParenPadCheck]( https://checkstyle.sourceforge.io/config_whitespace.html#ParenPad ) — #1196 
- [SimplifyBooleanExpression]( https://checkstyle.sourceforge.io/config_coding.html#SimplifyBooleanExpression ) — #1200
- [SimplifyBooleanReturn]( https://checkstyle.sourceforge.io/config_coding.html#SimplifyBooleanReturn ) — #1200
- [StringLiteralEquality](http://checkstyle.sourceforge.net/config_coding.html#StringLiteralEquality) — 10b41160c0a255b7914cac7e1254db5eba4b241b
- [TypecastParenPad]( https://checkstyle.sourceforge.io/config_whitespace.html#TypecastParenPad ) — #1196 
[...]

PR #1130, opened 2019-06-15
Split SearchService

tesshucom:
This is a draft of SearchService's split refactoring related to #1113.
I divided the class to withstand Lucene update and added the test cases to the high importance class.

The relationship between the divided classes is as follows.
(Filter and Tokenizer may be added after updating)

The granularity of the division prioritizes that differences in test results can be easily seen during Lucene updates, and that improvements can easily be made individually after the update.
___

class
___
#### Additional point of view

- SearchService is interfaced and holds the previous methods completely.
- The following is written in the SearchService comment: ``Performs Lucene-based searching and indexing.``
- Now, ``startIndexing`` ``index`` ``stopIndex`` can be deleted from SearchService. MediaScannerService refers to SearchService, but can be replaced with IndexManager (now, IndexManager is a deligate in SearchService).
- This means that the SearchService design is completely split up in search and indexing.

It is not included in PR because it is not related to Lucene's update. If it is desirable to make a change, will commit at the same time.

Is there a cool name for SearchServiceTermination ...?


airsonic-main/src/main/java/org/airsonic/player/service/search/AnalyzerFactory.java
```diff
@@ -0,0 +1,140 @@
...
+   * 
+   * @return analyzer for index
+   */
+  public Analyzer getAnalyzer() {
+    if (null == analyzer) {
+      try {
+        analyzer = new CustomAnalyzer();
+      } catch (Exception e) {
+        /*
+         * It is a transient description, and it is appropriate to detect an IOException at Lucene
```
jvoisin: I think you meant "temporary" :)

airsonic-main/src/main/java/org/airsonic/player/service/search/DocumentFactory.java
```diff
@@ -0,0 +1,175 @@
...
+
+  /**
+   * Create a document.
+   * 
+   * @param album
+   *        target of document
+   * @return document
+   * @since legacy
+   */
+  public Document createDocument(Album album) {
```
jvoisin: Can't this one be made generic, an merged with `public Document createAlbumDocument(MediaFile mediaFile)` ?

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexType.java
```diff
@@ -0,0 +1,209 @@
...
+        fieldNames(
+            FieldNames.ARTIST),
+        boosts(
+            entry(FieldNames.ARTIST, 2F))),
+
+    ;
+
+  public static final class FieldNames {
+
+    private FieldNames() {
```
jvoisin: Is this necessary?

airsonic-main/src/main/java/org/airsonic/player/service/search/AnalyzerFactory.java
tesshucom: That's right.
Here, a new version of Analyzer is generated after the update.

airsonic-main/src/main/java/org/airsonic/player/service/search/DocumentFactory.java
tesshucom: createAlbumDocument (MediaFile mediaFile) is an Album of File structure.
createDocument (Album album) is for Id3.
(Maybe it is hard to understand the name ...?)

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexType.java
tesshucom: It is necessary, but it is difficult to answer immediately if it is said about the description place and scope.
"Don't put it in a public enum if you can"
Is this all right?

tesshucom: It used to exist as a private FIELD_TITLE in SearchService.
As a result of the division, it was referred to from here and there.

I'll check if package access is possible.
If possible, exclude from enum.

airsonic-main/src/main/java/org/airsonic/player/service/search/DocumentFactory.java
jvoisin: ok

tesshucom:
I erased.

tesshucom:
Currently Airsonic has a reproducible Path search bug.
The content is very related to this PR ... what next?
Do you want to handle here or issue ?

jvoisin:
I would prefer to have it in a new issue :)

tesshucom:
If you scan with a specific pass pattern, the data is corrupted and incorrect search.
(I thought it was related to this PR, but it might not be related ...)
I'll Investigate a little more the playback procedure and then write a new issue.

tesshucom:
As for the scan, it seems that my confirmation was missing.
It used to be confused with the problem of nesting of directories with reported examples and children not being updated.

Issued for shuffle false search example instead.

tesshucom:
Earlier, I wrote that my clones contained some spec changes.
That is the problem above.
Folders are excluded from queries, and path comparisons exact-match.
This is a simpler solution as Japanese are traditionally plagued by native language pass strings.

Doing the same with Airsonic means that the functionality is lost.
It is better to update as it is or to change the analysis of the FOLDER field to [PathHierarchyTokenizer](https://lucene.apache.org/core/7_7_1/analyzers-common/org/apache/lucene/analysis/path/PathHierarchyTokenizer.html).
Changing to PathHierarchyTokenizer eliminates the #1139 bug.
In addition, folder search becomes intuitive because folder delimiters are simplified to path separators.

(Strictly speaking, spanOr affects scoring.
``Or phrase`` is an exact-match, path search is preferably adjusted by Boost value.)

In any case there is a solution.
The benefits of Lucene's version upgrade outweigh the slight increase in processing.
(Memory improvement, index compression rate, search speed improvement, etc.)
If that is determined to be necessary, it is good to change to a better process.

jvoisin:
I like the idea of changing the analysis of the `FOLDER` field to `PathHierarchyTokenizer` or a new dedicated `PathTokenizer`.

Worse case, increasing the size of the index is completely acceptable in my opinion.

tesshucom:
Sounds good.
If the FOLDER path is a media file path, I might have been convinced.
I do not know a little about the meaning of the Music folder path appearing in the query, but I think such a correspondence that does not break the function is a better.

___

The relationship of the current Analizer, that is, the field Tokenizer, Filter is as follows.

image

image

___

For early updates, I can suggest the following design.
There are few (relatively) differences.

image

image

I can create a PR as a draft.
You can also find the differences that you are interested in from the test case results and then filter them by adding more filters.
(For example, processing the processing that you do not like underbar with regular expression etc.)

#### KeywordTokenizer

 - Normalization processing existed in MediaType and Genre, but it is unnecessary if it is stored as a keyword.
 - The legacy normalization process is dirty code to prevent word segmentation in order to share Tokenizer.
 - When storing in storage without character processing, cross reference with the database becomes possible. This opens up new possibilities, such as multi-genre support on the Lucene side.

#### CJKWidthFilter

 - This is generally necessary. I think it is an oversight of the legacy design.
 - Stopwords are now passed through in upper case input.
 - In rare cases, the characters of CDDB may be entered in upper case. I can not search for such tag data.

#### EnglishPossessive

 - If this is attached, I think that there is not much difference between single coats.
 - The interpretation of UAX29 is a little different from legacy, but you may not need to stick with being completely the same.

I think that the explanation about the difference other than this is good after PR creation.

If you want to output the difference in behavior beautifully, I think it is better to separate this PR and the next PR.
If this PR is merged into the master or the verification branch, I can add them to both PR if want to add a new test case.

If you wish, I'll add to this PR as it is, but what should we do?

tesshucom:
It solved by doing some experiments.

In my clone, I wrote that the folder contains specification changes.
not ``SpanOr`` but ``Or``.
(I thought intuitively it should be so )

When I verified this accurately, the current SpanOr implementation can not be migrated to 7.x due to technical limitations.

Legacy saves the path as follows:

``
doc.add(new Field(FieldNames.FOLDER, mediaFile.getFolder(), Field.Store.NO,　Field.Index.NOT_ANALYZED_NO_NORMS
``

It is a field that is not analyzed and does not affect scoring.

In 7.x, SpanOr can not be used for NO_NORMS.
(This is an appropriate change and it is strange that SpanOr can be used in 3.x)
It is unclear what the original implementation intended for SpanOr, but it is correct to change the implementation and the case to Or.

When Path is processed with Single Token by KeywordAnalyzer instead of NO_NORMS, Spanor can be used.
(Of course it is an undesirable implementation. For verification.)
In this case, if there are Music Folders such as Music, Music2 and Music3, misanalysis occurs.
For example, shuffle search will pick up the mess.
The result that can be observed by the user is a bug close to #1139.
But the cause is a potential defect different from #1139.


- Rewrite implementation and test case to SpanOr.

This is a necessary measure.

___

There is another legacy mysterious behavior.

``
"+((art:abc* f:abc*) (art:def* f:def*)) +spanOr([f:" + PATH1 + "])", 
``

The path of Music Folder is specified in the main query, and comparison with the input value is performed.

I think this is not a meaningful specification.
(I do not understand unless I ask the author)

 - Folder is specified as NO_NORMS at the time of document generation. At least not related to scoring
 - However, since Folder is explicitly added to IndexType, it is analyzed by MultiFieldQueryParser. Due to the insufficient design of Analyzer, the input is affected by analysis by MultiFieldQueryParser and there is no substantial meaning as a search of Folder


The correspondence method is as follows:

 - The intention is unknown but to be implement as it is (It does not function as a proper search condition).
 - To be omit it because it causes false search.
 - Direct search of music folders is considered as a requirement, and correct as appropriate.

In the case of the third policy, search will be realized using PathHierarchyTokenizer, KeywordTokenizer and PathTokenizer etc...

___

Besides replacing the simple API syntax, rewriting SpanOr to Or is a condition for updating.
UAX#29 needs some confirmation about the handling of the underbar.

I can create a PR keeping this in mind.

jvoisin:
It's ok to break retro-compatibility: nobody is going to complain that the search is working way better :D

tesshucom:
I will create a PR from this.

If you want to increase the variation of the test and apply it after and before the update, it is useful if the PR so far is merged into the master or the verification branch.
If you want to make a decision on discard including push after this, I will push on this PR.
(Which one is better?)

___

If you want to postpone the decision to discard , there are also ideas to improve the Reader and include it in the evaluation..

The overall performance by the upgrade tends to improve, but the search overhead time increases by about 1 ms.
It will be improved by reusing the leader.

Reader is open every search and not closed 😣
it is recommended to fix.
I'll push in succession if necessary.

___

[MediaScannerService](https://github.com/airsonic/airsonic/blob/master/airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java)#
deleteOldIndexFiles() probably does not work.

It is a mechanism to change the generation of an existing index.
Database schemas also have generation management, but they can be thought of as similar roles.

This does not have to be immediate, but a fix in the future is ​​desirable.

jvoisin:
I'm ok with merging this PR so you can work on top of the master, since it's making the code a bit more clear: everyone prefers to review small PR, instead of massive ones.

But I would like that at least someone else review this PR, since it's moving a lot of things around.

Please close the `Reader` and fix `deleteOldIndexFiles` method is possible :)

tesshucom:
For the time being, i will proceed locally.
Once the policy is decided, I'll PR it to the desired location.

I hope the tests are done as thoroughly as possible.
I will do my best as much as possible, but it would be better if different people would be involved.

airsonic-main/src/main/java/org/airsonic/player/service/search/FieldNames.java
```diff
@@ -0,0 +1,74 @@
+package org.airsonic.player.service.search;
+
+/**
+ * Enum that symbolizes the field name used for lucene index.
+ * This class is a division of what was once part of SearchService and added functionality.
+ */
+class FieldNames {
```
muff1nman: Any particular reason for the changing of the field names?

airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceImpl.java
```diff
@@ -0,0 +1,298 @@
...
+  /**
+   * Common processing of random method.
+   * @param count Number of albums to return.
+   * @param searcher
+   * @param query
+   * @param id2ListCallBack Callback to get D from id and store it in List
+   * @return result
+   * @throws IOException
+   */
+  private final <D> List<D> createRandomFiles(
```
muff1nman: I think this method needs to be renamed. How about createRandomDocsList?

airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceTermination.java
```diff
@@ -0,0 +1,199 @@
...
+ * 
+ * Since SearchService operates as a proxy for storage (DB) using lucene,
+ * there are many redundant descriptions different from essential data processing.
+ * This class is a transfer class for saving those redundant descriptions.
+ * 
+ * Exception handling is not termination,
+ * so do not include exception handling in this class.
+ */
+@Component
+public class SearchServiceTermination {
```
muff1nman: This definitely needs a different name. It seems this class is mainly a holder for various utility functions related to searching. A generic name therefore might be SearchServiceUtilities?

airsonic-main/src/main/java/org/airsonic/player/service/search/FieldNames.java
tesshucom: The value has been changed simply to shorten the query.
(I just want to reduce the line breaks in assert code in [QueryFactoryTestCase](https://github.com/airsonic/airsonic/blob/53698e0d1f445edde555f67cecc166194710ff69/airsonic-main/src/test/java/org/airsonic/player/service/search/QueryFactoryTestCase.java)
Conversely, if you are reducing the readability of the query, restore it.


The reason for making a literal into another class is that it is not originally a value that SearchService should hold by design.
The class name has been FieldNames to avoid confusion with the org.apache.lucene.document.Field class

airsonic-main/src/test/java/org/airsonic/player/service/search/AnalyzerFactoryTestCase.java
```diff
@@ -0,0 +1,574 @@
+
```
muff1nman: Please use 4 space indentation. @jvoisin please add this to the list in #1121

tesshucom:
I`ll correct the point you pointed out.

There are currently inevitable CVE issues in Airsonic.
I may not be able to respond immediately.

```
spring-security-core-4.2.12.RELEASE.jar
(cpe:/a:pivotal_software:spring_security:4.2.12,
 org.springframework.security:spring-security-core:4.2.12.RELEASE) : CVE-2019-11272

spring-security-ldap-4.2.12.RELEASE.jar
(org.springframework.security:spring-security-ldap:4.2.12.RELEASE,
 cpe:/a:pivotal_software:spring_security:4.2.12,
 cpe:/a:pivotal_software:spring-ldap:4.2.12) : CVE-2019-11272

tomcat-embed-core-8.5.40.jar
(org.apache.tomcat.embed:tomcat-embed-core:8.5.40,
 cpe:/a:apache_software_foundation:tomcat:8.5.40,
 cpe:/a:apache_tomcat:apache_tomcat:8.5.40,
 cpe:/a:apache:tomcat:8.5.40) : CVE-2019-10072

jackson-databind-2.9.9.jar
 (cpe:/a:fasterxml:jackson:2.9.9,
  cpe:/a:fasterxml:jackson-databind:2.9.9,
  com.fasterxml.jackson.core:jackson-databind:2.9.9) : CVE-2019-12814
```

Jdom2.x exists in Airsonic classpath.
```
$ mvn dependency:tree | grep -i jdom
[INFO] +- org.jdom:jdom:jar:2.0.2:compile
```

In this case, jackson-databind causes a security issue.
There is no effective countermeasure at present.

The 2.9.9.1 or 2.10.0 release of jackson-databind is planned and is pending for several days.
It may be easier if this is released.

muff1nman:
@tesshucom id rebase with master. those cve issues should be fixed there.

tesshucom:
I understand.
It was already solved in upstream!

airsonic-main/src/main/java/org/airsonic/player/service/search/FieldNames.java
tesshucom: I revised that according to your advice.

airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceImpl.java
tesshucom: I revised that according to your advice.

airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceTermination.java
tesshucom: I revised that according to your advice.

airsonic-main/src/test/java/org/airsonic/player/service/search/AnalyzerFactoryTestCase.java
tesshucom: I revised that according to your advice.

muff1nman:
Looks good. Can you rebase? (I can too if you want).

tesshucom:
I'll leave it to you which granularity of rebase is appropriate!

muff1nman:
Rebased in 767b39ed5b9a74a99a566ae0b2236bb71cf3374a

Issue #233, opened 2017-01-22
Bulk export playlists


issuemover631:
<a href="https://github.com/spamawaymate"><img src="https://avatars1.githubusercontent.com/u/11852681?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [spamawaymate](https://github.com/spamawaymate)**
_Sunday Jan 22, 2017 at 20:15 GMT_
_Originally opened as https://github.com/Libresonic/libresonic/issues/232_

----

Hi,
Firstly, I just wanted to say a big thank you to all of the people involved in this project.
I'm not sure if this is the right place, but I wondered if it would be possible to bulk export playlists from libresonic?

eharris:
I'm actually wondering if it wouldn't be better just to change to have all playlists stored in some suitable playlist format on disk.  Then they would be kept up to date, no export would be needed and the playlist files on disk could easily be updated by other means (beets smart playlists anyone?) which would partially address or resolve #803, #752, #605, #370, and possibly some part of others (#556, #459).

I'm thinking something along the line of using XSPF files stored in `playlists/<owner>/<id>_<name>.xspf` or maybe just `playlists/id.xspf`.

PR #1131, opened 2019-06-15
Added ability to delete the default admin user

### LINKED ISSUE #640: Allow to delete default admin user
## Problem description

Currently, after installation, Airsonic creates default user `admin` that is used to make first configuration.
The problem is that there is no way to delete/disable such user - it rises security problem as any "bad" person already knows existing login.

Possible solutions:
- Allow to delete such user if any other user with admin rights present
- Allow to rename user on installation
- Allow to disable this user

eharris:
This transitions to solely using Admin Roles rather than a hardcoded admin
account.  This allows for deleting the initial admin account, which had
previously been a security weakness since the account name was not able to
be changed.  Fixes #640.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/LoginController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/LoginController.java
@@ -51,9 +51,11 @@ public class LoginController {
         map.put("brand", settingsService.getBrand());
         map.put("loginMessage", settingsService.getLoginMessage());
 
-        User admin = securityService.getUserByName(User.USERNAME_ADMIN);
-        if (User.USERNAME_ADMIN.equals(admin.getPassword())) {
-            map.put("insecure", true);
+        User admin = securityService.getUserByName("admin");
+        if (admin != null) {
+            if (admin.getUsername().equals(admin.getPassword())) {
+                map.put("insecure", true);
+            }
         }
 
         return new ModelAndView("login", "model", map);
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/SubsonicRESTController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/SubsonicRESTController.java
@@ -2093,7 +2093,10 @@ public class SubsonicRESTController {
         if (u == null) {
             error(request, response, ErrorCode.NOT_FOUND, "No such user: " + username);
             return;
-        } else if (org.airsonic.player.domain.User.USERNAME_ADMIN.equals(username)) {
+        } else if (user.getUsername().equals(username)) {
+            error(request, response, ErrorCode.NOT_AUTHORIZED, "Not allowed to change own user");
+            return;
+        } else if (securityService.isAdmin(username)) {
             error(request, response, ErrorCode.NOT_AUTHORIZED, "Not allowed to change admin user");
             return;
         }
@@ -2145,7 +2148,15 @@ public class SubsonicRESTController {
         }
 
         String username = getRequiredStringParameter(request, "username");
-        if (org.airsonic.player.domain.User.USERNAME_ADMIN.equals(username)) {
+        org.airsonic.player.domain.User u = securityService.getUserByName(username);
+
+        if (u == null) {
+            error(request, response, ErrorCode.NOT_FOUND, "No such user: " + username);
+            return;
+        } else if (user.getUsername().equals(username)) {
+            error(request, response, ErrorCode.NOT_AUTHORIZED, "Not allowed to delete own user");
+            return;
+        } else if (securityService.isAdmin(username)) {
             error(request, response, ErrorCode.NOT_AUTHORIZED, "Not allowed to delete admin user");
             return;
         }
--- a/airsonic-main/src/main/java/org/airsonic/player/dao/UserDao.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/dao/UserDao.java
@@ -147,10 +147,6 @@ public class UserDao extends AbstractDao {
      * @param username The username.
      */
     public void deleteUser(String username) {
-        if (User.USERNAME_ADMIN.equals(username)) {
-            throw new IllegalArgumentException("Can't delete admin user.");
-        }
-
         update("delete from user_role where username=?", username);
         update("delete from player where username=?", username);
         update("delete from " + getUserTable() + " where username=?", username);
--- a/airsonic-main/src/main/java/org/airsonic/player/service/PlaylistService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/PlaylistService.java
@@ -27,7 +27,6 @@ import org.airsonic.player.dao.PlaylistDao;
 import org.airsonic.player.domain.MediaFile;
 import org.airsonic.player.domain.PlayQueue;
 import org.airsonic.player.domain.Playlist;
-import org.airsonic.player.domain.User;
 import org.airsonic.player.service.playlist.PlaylistExportHandler;
 import org.airsonic.player.service.playlist.PlaylistImportHandler;
 import org.airsonic.player.util.FileUtil;
@@ -296,7 +295,9 @@ public class PlaylistService {
         }
         InputStream in = new FileInputStream(file);
         try {
-            importPlaylist(User.USERNAME_ADMIN, FilenameUtils.getBaseName(fileName), fileName, in, existingPlaylist);
+            // With the transition away from a hardcoded admin account to Admin Roles, there is no longer
+            //   a specific account to use for auto-imported playlists, so use the first admin account
+            importPlaylist(securityService.getAdminUsername(), FilenameUtils.getBaseName(fileName), fileName, in, existingPlaylist);
             LOG.info("Auto-imported playlist " + file);
         } finally {
             FileUtil.closeQuietly(in);
--- a/airsonic-main/src/main/java/org/airsonic/player/service/SecurityService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/SecurityService.java
@@ -166,11 +166,22 @@ public class SecurityService implements UserDetailsService {
      * Returns whether the given user has administrative rights.
      */
     public boolean isAdmin(String username) {
-        if (User.USERNAME_ADMIN.equals(username)) {
-            return true;
-        }
         User user = getUserByName(username);
         return user != null && user.isAdminRole();
+    }
+
+    /**
+     * Returns username with admin rights.
+     *
+     * @return username of admin user
+     */
+    public String getAdminUsername() {
+        for (User user : userDao.getAllUsers()) {
+            if (user.isAdminRole()) {
+                return user.getUsername();
+            }
+        }
+        return null;
     }
 
     /**
--- a/airsonic-main/src/main/java/org/airsonic/player/domain/User.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/domain/User.java
@@ -26,7 +26,6 @@ package org.airsonic.player.domain;
  */
 public class User {
 
-    public static final String USERNAME_ADMIN = "admin";
     public static final String USERNAME_GUEST = "guest";
 
     private final String username;
@@ -125,7 +124,7 @@ public class User {
     }
 
     public boolean isSettingsRole() {
-        return isSettingsRole;
+        return isAdminRole || isSettingsRole;
     }
 
     public void setSettingsRole(boolean isSettingsRole) {


eharris:
I would recommend that #1129 is merged before this.

airsonic-main/src/main/java/org/airsonic/player/dao/UserDao.java
```diff
@@ -147,7 +151,7 @@ public void createUser(User user) {
      * @param username The username.
      */
     public void deleteUser(String username) {
-        if (User.USERNAME_ADMIN.equals(username)) {
+        if (settingsService.getAdminUser().equals(username)) {
             throw new IllegalArgumentException("Can't delete admin user.");
```
jvoisin: Maybe show the username of the current adminUser?

airsonic-main/src/main/java/org/airsonic/player/service/SecurityService.java
```diff
@@ -162,11 +162,20 @@ public User getUserByEmail(String email) {
         return userDao.getAllUsers();
     }
 
+    public String[] getAllUsernames() {
+        List<User> users = userDao.getAllUsers();
+        String[] usernames = new String[users.size()];
+        for (int i = 0; i < users.size(); i++) {
```
jvoisin: nit: I would have used a `foreach` here.

eharris: I would have too, except for-each doesn't provide the index, and since I'm building an array I need the index.

jvoisin:
I'll test this one as soon as #1129 is merged :)

airsonic-main/src/main/java/org/airsonic/player/dao/UserDao.java
eharris: I'd be happy to add that if you think it's needed.  The only reason I didn't change the exception message is that I tend to prefer trying to keep minimal changesets (like avoiding unnecessary or unrelated formatting changes).

jvoisin: I do agree with your reasoning, as I tend to do the very same, but in this case, the change is so trivial that I would prefer to see it done here, instead of having to review a second commit simply changing this line.

But on the other hand, I guess they'll be more commits on top of this to polish the feature a bit more, so feel free to ignore this comment :)

eharris: In reality, with the changes from #1129, its very unlikely that this will ever happen, but I went ahead and added the username as requested.

eharris:
Due to changes in another PR, this will have merge conflicts, but I'll fix them after #1129 is merged.

airsonic-main/src/main/java/org/airsonic/player/controller/GeneralSettingsController.java
```diff
@@ -77,6 +81,9 @@ protected void formBackingObject(Model model) throws Exception {
             }
         }
 
+        String currentAdminUser = settingsService.getAdminUser();
```
jvoisin: This variable is unused.

airsonic-main/src/main/java/org/airsonic/player/validator/UserSettingsValidator.java
```diff
@@ -99,6 +99,9 @@ public void validate(Object obj, Errors errors) {
             }
         }
 
+        if (command.isDeleteUser() && settingsService.getAdminUser().equals(username)) {
+            errors.rejectValue("deleteUser", null, "Cannot delete the primary admin user");
```
jvoisin: Maybe we should show the username in the log?

airsonic-main/src/main/webapp/WEB-INF/jsp/generalSettings.jsp
```diff
@@ -80,6 +80,18 @@
 
         <tr><td colspan="2">&nbsp;</td></tr>
 
+        <tr>
+            <td><fmt:message key="generalsettings.adminuser"/></td>
+            <td>
+                <form:select path="adminUser" cssStyle="width:15em">
+                    <c:forEach items="${command.usernames}" var="username">
+                        <form:option value="${username}" label="${username}"/>
```
jvoisin: Please escape the values with `fn:escapeXml` otherwise we'll have XSS issues here.

airsonic-main/src/main/java/org/airsonic/player/controller/GeneralSettingsController.java
eharris: Ah yeah, I refactored out the part that used that

airsonic-main/src/main/webapp/WEB-INF/jsp/generalSettings.jsp
eharris: Done

airsonic-main/src/main/java/org/airsonic/player/validator/UserSettingsValidator.java
eharris: Done

muff1nman:
This seems bizzare to me. What is the benefit of adding an AdminUser setting? IMO we should either go all the way to Roles where there can be multiple admin users or leave it as is.

eharris:
This PR just makes it so that the "primary admin" user can be any account rather than being hardcoded as the user with the username of `admin`, and is aimed at the minimal change-set that will allow that.

Roles do already allow for multiple accounts to have "admin" privileges, but I think it still makes sense to have a special "primary" admin account that has special protections and actions, such as not being allowed to be deleted (even by other users with the admin role).  For simplicity and a clean PR, this doesn't add anything additional, but I could see things like the primary admin account automatically having access to all media directories, whereas now individual accounts, even if they have the "admin" role do not.  There may be other "special" things that would be useful as well.

But if you would prefer not to have anything past the more general "admin" role, I can probably rework the PR to do that.

eharris:
@muff1nman Did you still want changes to this in order to merge it?

muff1nman:
My gut feeling is that we should either leave it as is or go all the way to an admin role that can be assigned. Having a `special` admin seems like a bad idea in my mind. If we need further segmentation between lower and higher level admins, that probably means there should be more than one role.

eharris:
I've updated this PR as @muff1nman has requested to not have a "special" admin account and instead use only roles for admin.

A side effect of this change is that automatic playlist imports do not work any longer, since there is no "primary" user to own them, and there is not (yet) any way to auto-import playlists for individual users.  One way of resolving this is described in https://github.com/airsonic/airsonic/issues/233#issuecomment-513936052

eharris:
Fixed the merge conflicts from other merges.

airsonic-main/src/main/java/org/airsonic/player/controller/SubsonicRESTController.java
```diff
@@ -2093,7 +2093,10 @@ public void updateUser(HttpServletRequest request, HttpServletResponse response)
         if (u == null) {
             error(request, response, ErrorCode.NOT_FOUND, "No such user: " + username);
             return;
-        } else if (org.airsonic.player.domain.User.USERNAME_ADMIN.equals(username)) {
+        } else if (user.getUsername().equals(username)) {
+            error(request, response, ErrorCode.NOT_AUTHORIZED, "Not allowed to change own user");
```
jvoisin: What about "You're not allowed to update the properties of your own user" instead?

airsonic-main/src/main/java/org/airsonic/player/controller/SubsonicRESTController.java
```diff
@@ -2145,7 +2148,15 @@ public void deleteUser(HttpServletRequest request, HttpServletResponse response)
...
 
         String username = getRequiredStringParameter(request, "username");
-        if (org.airsonic.player.domain.User.USERNAME_ADMIN.equals(username)) {
+        org.airsonic.player.domain.User u = securityService.getUserByName(username);
+
+        if (u == null) {
+            error(request, response, ErrorCode.NOT_FOUND, "No such user: " + username);
+            return;
+        } else if (user.getUsername().equals(username)) {
+            error(request, response, ErrorCode.NOT_AUTHORIZED, "Not allowed to delete own user");
```
jvoisin: What about "You're not allowed to delete your own user" instead?

airsonic-main/src/main/java/org/airsonic/player/service/PlaylistService.java
```diff
@@ -27,13 +27,12 @@
...
 import org.airsonic.player.domain.PlayQueue;
 import org.airsonic.player.domain.Playlist;
-import org.airsonic.player.domain.User;
 import org.airsonic.player.service.playlist.PlaylistExportHandler;
 import org.airsonic.player.service.playlist.PlaylistImportHandler;
 import org.airsonic.player.util.FileUtil;
 import org.airsonic.player.util.Pair;
 import org.airsonic.player.util.StringUtil;
-import org.apache.commons.io.FilenameUtils;
+//import org.apache.commons.io.FilenameUtils;
```
jvoisin: This can be deleted I think

airsonic-main/src/main/java/org/airsonic/player/service/PlaylistService.java
```diff
@@ -296,8 +295,11 @@ private void importPlaylistIfUpdated(File file, List<Playlist> allPlaylists) thr
         }
         InputStream in = new FileInputStream(file);
         try {
-            importPlaylist(User.USERNAME_ADMIN, FilenameUtils.getBaseName(fileName), fileName, in, existingPlaylist);
-            LOG.info("Auto-imported playlist " + file);
+            // FIXME - With the transition away from a hardcoded admin account to Admin Roles, there is no longer
```
jvoisin: Should this be fixed before merging this PR?

airsonic-main/src/main/java/org/airsonic/player/service/PlaylistService.java
eharris: It could be, but I left it in but commented out so its there whenever a solution to the playlist import issue is decided.

airsonic-main/src/main/java/org/airsonic/player/service/PlaylistService.java
eharris: No one has suggested a good solution to this issue, and I don't want fixing this security issue to wait on one.  This side effect has only become a problem because @muff1nman wanted not to keep a "primary" admin user, so I've changed the PR to comply with that request.  @muff1nman which would you prefer, keeping this version with this side effect, or going back to the original version of this PR that kept a "primary" admin user?

airsonic-main/src/main/java/org/airsonic/player/controller/LoginController.java
```diff
@@ -51,9 +51,11 @@ public ModelAndView login(HttpServletRequest request, HttpServletResponse respon
         map.put("brand", settingsService.getBrand());
         map.put("loginMessage", settingsService.getLoginMessage());
 
-        User admin = securityService.getUserByName(User.USERNAME_ADMIN);
-        if (User.USERNAME_ADMIN.equals(admin.getPassword())) {
-            map.put("insecure", true);
+        User admin = securityService.getUserByName("admin");
```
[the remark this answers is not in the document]
eharris: Just because this is the only place it's ever used, and only once, and only because the initial install uses this value.  I could change it to a constant, but that kind of implies that it might be reasonable to use elsewhere, and it never should be.

airsonic-main/src/main/java/org/airsonic/player/controller/SubsonicRESTController.java
eharris: Since these messages are in the REST API, I kept them very terse, as no one should ever see them except developers.

airsonic-main/src/main/java/org/airsonic/player/controller/SubsonicRESTController.java
eharris: Since these messages are in the REST API, I kept them very terse, as no one should ever see them except developers.

airsonic-main/src/main/java/org/airsonic/player/service/PlaylistService.java
eharris: I'm probably going to work on the solution I proposed in https://github.com/airsonic/airsonic/issues/233#issuecomment-513936052 at some point which would resolve this playlist issue as well as many others, but that is almost certainly going to be a low priority.

muff1nman: For now could you chose the first admin user found? Really I'd like to see the playlist owner field be nullable but that will probably take some work todo.

eharris: I'd rather not put in a half-baked hack, since it seems unlikely that playlist importing is all that usable in its current form anyway(*).  I'd rather get the security regarding the hardcoded admin account in, and then later get some focus on improving/fixing the playlist functionality.

(*) Is there even any documentation on how the playlist importing stuff is supposed to work that would help anyone actually use it?  For instance, how are files in the playlist supposed to have their paths specified?  How does it pick which media folder is used as the base? etc...

muff1nman: I used it a year ago. I don't think the negligible security gained here is worth losing the feature over.

eharris: @muff1nman So then would you prefer that I revert back to the original version of this PR that kept a "primary" admin user, in order to preserve the playlist functionality?  I don't want to put playlist overhaul in as part of this PR.

muff1nman: Nope I'd prefer you kept it as is. For the playlist user you can:

1. Use the first admin user
2. Allow the playlist import user to be configurable
3. Change the playlist to allow a nullable user field

Although I still think 3 would be ideal/workable, 1 or 2 would be considerably easier and be much better than straight out removing this functionality.

eharris: As requested, PR has been modified to use the first admin account for playlist imports.

airsonic-main/src/main/java/org/airsonic/player/service/SecurityService.java
```diff
@@ -166,13 +166,24 @@ public User getUserByEmail(String email) {
...
+     *
+     * @return username of admin user
+     */
+    public String getAdminUsername() {
+        for (User user : userDao.getAllUsers()) {
+            if (user.isAdminRole()) {
+                return user.getUsername();
+            }
+        }
+        return null;
```
muff1nman: I'd throw an exception here instead of returning null.

eharris: I don't really like returning exceptions from simple lookup functions, especially when returning null for "not found" is such a universally accepted (and expected) pattern.  I think it restricts potential use cases unnecessarily and violates separation of concerns.  Why should this function need to enforce that the higher level system thinks not having a user with a particular role is an exception-worthy problem?

In any case, I believe that when this function returns null, the calling site will still generate an exception when it tries to use the null value, and does so without this function needing to have any concept of higher level user permissions concerns.

I also didn't want to rule out the possibility that someone might want to completely lock down their installation by disabling all admin accounts (even though doing so would require directly changing the db).

Returning null also allows for if/when someone changes the code to implement your suggestion number 3 above (making the owner of imported playlists nullable), although I admit that at that point this function might just be removed entirely, but who knows?

eharris: @muff1nman If throwing an exception is a requirement to get your approval, say so and I will change.

muff1nman: > I don't really like returning exceptions from simple lookup functions, especially when returning null for "not found" is such a universally accepted (and expected) pattern.

In Java land its not universally accepted. In other languages, go, c/c++, python, yes I would agree.

>  Why should this function need to enforce that the higher level system thinks not having a user with a particular role is an exception-worthy problem?

Its in the SecurityService and warrants a bit more checking around assumptions to ensure we fail fast. With this change we still have an assumption that there should always be at least one admin user. When that assumption changes, we should get this exception and we can change the code to deal with the new assumptions then.

> I also didn't want to rule out the possibility that someone might want to completely lock down their installation by disabling all admin accounts (even though doing so would require directly changing the db)

In the case the user does this anything that uses this function should rightly be broken. If the calling code can be robust in the scenario of no admin users, its on the function to handle the exception. 

> Returning null also allows for if/when someone changes the code to implement your suggestion number 3 above (making the owner of imported playlists nullable), although I admit that at that point this function might just be removed entirely, but who knows?

And when that happens we can change/remove this code too :)

eharris: It should be noted that the other user lookup functions already in SecurityService (e.g. `getUserByName()`, `getUserByEmail()`) also use returning null to indicate "not found", so throwing an exception rather than returning null would be inconsistent with those.

airsonic-main/src/main/java/org/airsonic/player/service/SecurityService.java
```diff
@@ -166,13 +166,24 @@ public User getUserByEmail(String email) {
...
         return user != null && user.isAdminRole();
     }
 
+    /**
+     * Returns username with admin rights.
+     *
+     * @return username of admin user
+     */
+    public String getAdminUsername() {
+        for (User user : userDao.getAllUsers()) {
```
fxthomas, line 179: Minor thing: we should probably make sure the order is consistent (order by id or user name), I'm not sure of the behavior of all databases, and this might be confusing if the order somehow changes between runs.

fxthomas:
I'm seeing no issue for now in my instance after removing the `admin` user. Let's see if it stays this way for a few days and I think we can merge this.

fxthomas:
I haven't see any issue, this sounds good to go!

fxthomas:
Let's go ahead and merge it. Thanks for the lot of work on this y'al!

PR #1132, opened 2019-06-15
Replace underscores with spaces in paths for display elements

### LINKED ISSUE #709: Feature: Config option to display underscores as spaces in directory names
It would be a nice feature to have a config option to convert underscores in directory names to spaces when displayed in the web interface.  Global for the server would be ok, but even better would be per-user.

Most of my media files are purposefully named with underscores to avoid having to quote things when doing file management and when using tab completion.  But when they are displayed in the web interface, having underscores is ugly and takes up more space than spaces.

eharris:
Originally I thought it might be worth making this configurable, but I decided against it because I doubt anybody really cares about preserving underscores in display elements that come from filenames, and replacing them with spaces makes the UI so much cleaner in so many places.  Fixes #709.


jooola:
Nice ! This is something we need, but should we make this option configurable ? I hope this will enhance the Last.fm features since it doesn't work with underscores in artists / albums/ title names.

eharris:
This only does the replace on filename/directory elements, not on any of the metadata from files.  I didn't want to mess with metadata because I haven't really seen any files with underscores in the metadata, and I figured it was more likely that someone might care about preserving the metadata exactly.

tesshucom:
I think it's great and it's a display device that is generally very popular.
As with some products(foobar, WMP etc), it may be too considered to replace semicolons with commas in artist name and genre names.
However, it is important whether the change of domain object is appropriate.
Metadata exists not only in databases but also in search engines, and cooperates via domain objects.

As an extension of this idea, a method such as getViewName () is often added to prevent the impact on the existing implementation.
It is also done to avoid the domain object extension, and to be implemented completely as a view.

eharris:
@tesshucom you raise valid concerns, if metadata was actually being modified or reformatted.  But this change does not affect any of the metadata that comes from media files in any way.  The changes only affect the display of filename/last path segment which is only used in the fallback case when no metadata is available.

tesshucom:
Focusing mainly on the display, you are right.
However, after registering meta to DB at scan time, all subsequent processing uses domain object via dao.
It is recommended to keep in mind that it also affects updating MediaFile.
Tag editing, some URL generation is also affected.
These should be committed as design changes.

Sorry, my opinion may be a stubborn general theory.

tesshucom:
Sorry, my opinion may be a stubborn general theory.
It is difficult to handle analysis rules, so I might have been nervous.

Under normal circumstances, it would be reasonable to deal with  design management the following place.
(the initial analysis specification)

``
MetaDataParser#guess**
``

I was confused because there was a duplicate code.

tesshucom:
Sorry, my opinion may be a stubborn general theory.
It is difficult to handle analysis rules, so I might have been nervous.

Under normal circumstances, it would be reasonable to deal with  design management the following place.
It's the initial analysis specification.

``
MetaDataParser#guess**
``

> Or to be solved when creating domain object etc.

Already, it has been designed like that.
I was confused because there was a duplicate code.
It's unclear why there's already a trick in the getter.

If you balance it with the search specification, you will probably see something like this:

```
    public static void main(String[] args) {

        String fileName = "_aaaa _b ccc_ddd_";
        
        Collector<String, StringBuilder, String> join =
                Collector.of(StringBuilder::new, 
                   StringBuilder::append, StringBuilder::append, StringBuilder::toString);
        String title =
        Arrays.asList(fileName.split("\\s")).stream()
            .map(s -> s.replaceAll("^_", ""))
            .map(s -> s.replaceAll("_$", ""))
             // .map(s -> s.replaceAll("_", " ")) // It is not erased in search index
            .map(s -> s.concat(" "))
            .collect(join)
            .trim();
        System.out.println(title);

    }
```
The result is like this: ``aaaa b ccc_ddd``.

eharris:
I'm not sure why the search provisions/code matters, because as far as I know, the search only applies to actual (found and indexed) media files, whereas the code this PR modifies is only used for displaying the filename/pathname, which may or may not contain media files.

tesshucom:
Have you confirmed the scope of the subsequent processing?

image

eharris:
`getName()` is called from many places so it's hard to verify them all, but in most cases there is metadata present, so the cases that this PR affects are quite limited.  I can say is I've been running this mod in my branch for quite a while and have not seen any any negative effects.

This PR is the minimal change that improves the behavior significantly in a set of very specific circumstances, including resolving a feature/bug report that has been open for quite a long time (16 months now).  I'm all for improving the code structure and layout, and if you'd like to submit a PR addressing the concerns you have raised, I would be very happy to see that, but as a better solution has not been forthcoming in the past year, I don't think that we should wait on that.  Perfect is the enemy of progress, and this change does not preclude someone doing that at some point subsequently.

tesshucom:
I myself do not oppose this PR.
If the display changes only at a specific location, there are other ways.

 - I just wanted to see where the risk or loss of functionality was by changing domain objects that rarely change, as no tests were shown.
 - There may be similar requirements. (Because such work is desired for everyone) For reference at that time

eharris:
Perhaps it is worth noting that `MediaFile.getTitle()` already exists and is used to retrieve the title metadata, which may not be present.  So `getName()` is used in situations when it is preferred to get something (anything) to be used for display or comparison, even if it is not necessarily perfectly accurate or even predictable.  The uses are all transient (not persisted), so any difference in behavior from previous versions should be inconsequential.  In fact `getName()` appears to be used as if it were `getDisplayName()`, but I don't really want to rework the naming for this PR since it would touch many different places.  (Again, this could be done in another PR if desired).

Here are the notable locations where `getName()` is used and the impact this PR would have:

service/upnp/MediaFileUpnpProcessor.java: Used as a display name.
service/MusicIndexService.java: Helps, used for comparison and sorting, since removing of articles doesn't work in the current code when underscores are included.
service/LastFmService.java: Helps, since presumably matching on LastFM should improve
service/sonos/SonosHelper.java: Helps, used as display name.
controller/SubsonicRESTController.java: Used as display name.
controller/CoverArtController.java: Uses getName (which returns the title) as the album or artist info.  This is probably incorrect, but should be addressed in a separate PR.
service/upnp/FolderBasedContentDirectory.java: Used as display name
domain/MediaFileComparator.java: Used for comparison/sorting
ajax/MultiService.java: Helps, used as display name

tesshucom:
- This PR should be actively applied to LastFmService.
 - As for sorting, there is a possibility of degradation and it needs to be considered separately.
 - You may be cautious about other display names. Because the removal of the underscores causes an unsearchable case. If you really want to erase, as already suggested, there are also removal rules that do not interfere with the search. However, it is only technically possible and needs should be considered.

> This PR is the minimal change that works within the already existing code structure and improves the behavior significantly in a set of very specific circumstances, including resolving a feature/bug report that has been open for quite a long time (16 months now). I'm all for improving the code structure and layout, and if you'd like to submit a PR addressing the concerns you have raised, I would be very happy to see that, but as a better solution has not been forthcoming in the past year, I don't think that we should wait on that. 

Rationality is worth considering.
"Improvement" is a powerful word, but whether it is an improvement should be judged separately.
(LastFm linkage should be in accordance with the LastFm specification.)

I can understand the feeling that the underscores is not beautiful.
However, guidelines on how word breaks should be treated in software exist in Unicode.
Products around the world that follow this will treat ``_ABC_DEF_`` as is.

View of Airsonic  has not contradicted it until now.
With regard to modernization, this PR is a retrospective specification change.

The search specification is special because it uses an old library before UAX#29.

eharris:
@tesshucom Addressing your last points:

* Since I do not use LastFM, I would feel uncomfortable adding changes there without testing.  I'm sure a PR addressing LastFM would be welcomed.
* All sorting operations this affects are self-consistent.  Although some orderings may change, I don't believe there is any negative effect.  If there is, please point out those cases.
* As stated previously, this PR only affects filename/filepath display elements.  If there are specific search cases that need support of including underscores in path elements, those should probably be using `getPath()` rather than `getName()` and only `getName()` is being modified here.  Again, I'm sure a PR that addresses or improves search aspects would be welcomed.

tesshucom:
- If you do not want to affect LastFM, you should have examined the Impact range
 - Changing sort order is a sign of bad effect
 - If you display AB_CD as ABCD , the user does not hit the search after copy&paste ABCD. It is not support but over-processing.

#### Guidelines for Contributing

In order to avoid controversy based on personal sense, I think that the UNICODE guidelines will be helpful.
An underscore is part of a word, not a punctuation mark or delimiter.
Fortunately, most Airsonic displays match this one.

 - Users can remove underscores from files and tags at their own discretion.
 - Airsonic will follow when the user removes the underscore.
 - We can also guarantee consistency with other products that follow the UNICODE guidelines.

If some special specifications are introduced easily, consistency with other processes and other products will be an issue later.

It's not like a good idea to add a unique perspective to a search specification too.
The part which is not control increases by user's intention.

eharris:
@tesshucom:

I don't know why you keep bringing up concerns about search and tagging impacts.  This change affects neither.  Please restrict your comments to actual issues involved in this PR.

As far as I can tell your concerns about Unicode and UAX#29 are also unrelated.  As far as I can see, UAX#29 doesn't even mention underscores, except as a substitution for spaces in examples.  Additionally, word breaks are also not germane to this change, as filenames are not parsed, they are whole entities and have no need of word-break rules.

You stated previously "This PR should be actively applied to LastFmService".  As I already stated, I don't use lastFM, but you are welcome to submit a PR that does that if you think it is needed.

I disagree that changes to sort order are a sign of bad effect.  If anything, the effects are beneficial.

As far as the "No Breakage without discussion", there's been lots of discussion on this PR, and the ticket requesting this change has been open for over a year, so I believe that is satisfied.  Besides, I don't think this can be considered a large change to functionality.

muff1nman:
How about this.

Can we add a new method called `getDisplayName`. It calls `getName` and does whatever humanizing is desired. Then callers can be modified to use that new method on a case by case basis in subsequent PRs.

eharris:
@muff1nman If we go that direction, I would think we could just rename `getName` to `getDisplayName`, as that is essentially what it is already.  I don't think there's anyplace we wouldn't want to use it.  Since there is already a `getTitle` function that returns the metadata title, `getName` is just a convenience function that bridges between the "view metadata" and "view filesystem" modes, and ensures that the `MediaFile` entity still has something displayable even if there's no metadata for it.

tesshucom:
If we have successfully updated to Lucene 8.1.1, it will be easy relatively to switch between Analyzers and recompile.
If Airsonic supports the latest version of Lucene, you can recompile it if you wish, replacing it with an Analyzer that is specialized for each country with specialized processing.

This is more informative and inspirational than English-speaking people imagine.
As far as I know, there is no servers that handle a large amount of music that can easily introduce for each country language dedicated processing .

This is one of the reasons why I avoid mixing excess unique logic implementations into search functions, and stick to global practices(UNICODE).

Therefore, I may often prioritize engineer's judgment over sensibility judgment.
Because I am not good at English, I may be violent in the way of communication.
I'm sorry.

eharris:
@tesshucom I think maybe your last comment on this thread was meant for another ticket?  This ticket does not relate to search at all.

tesshucom:
No, it is involved.
As stated in the process, users equate search and visible things to some extent.

It was not a good idea to replace the scope of impact without at least investigating it.
If for some reason you give a completely different display, you should be somewhat careful and have a design commitment.

eharris:
This is nowhere near "completely different".  I think most people equate underscores with spaces, this merely treats them as such.  Can you point to a single other language that utilizes underscores as lexical content?

tesshucom:
Read UNICOEDE's definition of Clusters.
If no special control role is granted, all the letters are part of a word.
For that reason In the latest Lucene, there is no removal of underscores in StandardTokenizer (≒ EnglishTokenizer).

The 3.x StandardTokenizer is the current ClassicTokenizer(We do not know how long it will be supported).
Selecting this does not remove all underscores.

Of course this depends on Analizer, but in English traditionally underscores are treated politely.

> most people equate underscores with spaces, 

This is not a generalizable opinion.

eharris:
I think you are bootstrapping an unrelated concern (search) into this PR merely based on the opinion that BOTH of these are significant concerns:

* some users might expect underscore to not be a word boundary
AND
* that the presentation layer's treatment of underscores must be consistent with the search usage

In my opinion, neither of those concerns alone outweigh the very large benefit to the look and usability of this change, and both of them together is ever so much smaller due to the geometric relationship of `AND` (small number * small number = very small number).

I'm tired of this discussion, and ask for a ruling from the maintainers.

jvoisin:
I'm on @eharris 's side for this one, as I think that
-  we want this behaviour to happen by default
- this PR can and should be merged

But I'll defer to our beloved project leader, @muff1nman for the final decision :)

tesshucom:
The final decision from the maintainer is justice.
However, my opinion is based on legacy source code and general design theory, not personal feeling.

What MediaFile#name indicates is not a display name.
This is the primitive specification created in the first stage.
Careful designers will notice that getAlbumName exists and getArtistName does not exist.

Subsonic/Airsonic is a typical MVC model, but there is no viewObject.
That is, the domain value (file structure or id3) is output to the presentation as it is.

Legacy view implementations are file structure based, not ID3 based.
Therefore, the display of the file name as it is is a intended specification, not an imperfect implementation.
(Index, directory snippet, search result artist, etc.)

The reason why they have not been rewritten until now is that.
It only needs to be able to switch to an ID3-based view, but it wasn't added because of the hard work.
That said, no one destroyed the file structure based design.

I am not against inventing the concept of data flow that is neither file structure nor id3.
If possible, give these types of changes specially prominent labels.

fxthomas:
Well, I had a quick read at the lengthy discussion. I'm not sure I got everything, but running a quick search in MusicBrainz yields artists like https://musicbrainz.org/artist/6e67dca0-80c0-4dbd-9063-faa176f99e1e. How does this PR handle them?

(I have no idea what their music sound like, and I am intrigued!)

eharris:
@fxthomas As noted above, this PR doesn't affect Metadata at all, the original PR only caused the displayed names of **file** elements to substitute spaces for underscores.  Since then, I've added another cleanup to convert all continuous sequences of whitespace characters to a single space, since most browsers collapse whitespace anyway.

But since you pointed it out, the fact that MusicBrainz' own search algorithm appears to conflate underscore (`_`) with dash (`-`) and other "inter-word space" characters is a perfect reason why this PR would be useful.

As far as the specific groups and releases that you mention from MusicBrainz, I don't really care how they are treated.  If a group wants to intentionally name themselves or something they produce after punctuation (or other nonsense) characters in an attempt to be "cute", then they deserve whatever happens to them, including marginalization.  Just because an exception case **can** be made for any rule, doesn't mean that's a good basis for halting development progress that would dramatically help in 99.999% of cases.

fxthomas:
That was a really narrow edge case, but I couldn't resist, sorry! Interestingly, MusicBrainz also uses Lucene under the hood, so their search syntax might be interesting to look at.

The PR seems good to me otherwise ; the `getDisplayName` method makes it clear that it is only used for display and nothing more.

fxthomas:
Let me add that I think that @tesshucom's concerns as well as potential edge cases could be resolved by making this configurable as initially planned ; it would also clearly document in the settings UI that this is the expected behavior and not a bug. That would be my preferred solution.

An alternative solution would be to use `alt` text to display the full title in the tooltip, but it is less easy to discover.

tesshucom:
What fxthomas says is probably a safe response.
Replacing a file structure-based view with something else, no matter how trivial, means a conversion of the specification that is not backward compatible.
Depending on the committer, Airsonic will determine finally that it has started misleading specifications.

If it's explicitly labeled "Specification Change" this PR, I think it's a future Airsonic policy.
The committer can determine that the project is likely to have similar modifications in the future.

The inability to show either attitude is also an option.

eharris:
If someone comes up with a real (non-theoretical) reason why this deserves to be configurable, I'd be happy to add that ability.  I just don't currently feel like it warrants the effort since I can't see a use case when someone would actually want to preserve underscores in filenames.  If they do want to preserve underscores, they can easily do so by providing it in the metadata, which is not altered by this PR.

tesshucom:
Perhaps the scope I see is different from eharris.

 - If the affected area extends to Rest, the document must be changed. API is file structure or ID3.
https://airsonic.github.io/docs/api/
``We will try not to break with the Subsonic API.``

 - Even if the impact is limited to (Web, Sonos, UPnP) other than Rest it is better to announce that the inconsistency with the API is not a bug but a specification.

 - We proposed getDisplayName () as one of the proposals, but this is essentially a requirement of the Controller. Maintainability is greatly reduced because it breaks the MVC boundary. 
Short code is good. Exactly. At the same time bugs are easy to mix and it is more difficult to maintain a robust design.

This is not a strong demand and I respect Airsonic's policy.
Because I can remove unnecessary code with clone.
But in that case,  I don't think I'll commit anymore.
(There are some known bugs.)
Because verification of systems with different domain designs requires extra effort.

eharris:
I don't see this as breaking the Subsonic API in any way.  Yes, path elements will have some characters translated, but there's no possibility that will actually cause any breaking effects unless some  (purely theoretical) thing was using both the API and *also* using the filesystem directly as well.  To me that theoretical possibility seems like nonsense.  The file "structure" (meaning hierarchy) is still perfectly maintained.

eharris:
And if your complaint is that we changed to use `getDisplayName()` several places, I'd be perfectly happy to go back to the original PR which was a simple 2-line change in one function.

tesshucom:
I think you should do whatever you want.
Clone servers can eliminate design misleading.
These are the good things about open source.
I already showed a relatively conservative view.
> There may be similar requirements. For reference at that time.

There is a simpler and no design risky way than changing two lines.
Renaming files is easier and rational than removing \0 from a tag.

PR #1135, opened 2019-06-15
Don't create arrays for varargs functions

jvoisin:
There is no need to create a new array of Objects
when calling varargs methods like `update`.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/dao/RatingDao.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/dao/RatingDao.java
@@ -89,7 +89,7 @@ public class RatingDao extends AbstractDao {
      */
     public Double getAverageRating(MediaFile mediaFile) {
         try {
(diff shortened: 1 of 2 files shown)


airsonic-main/src/main/java/org/airsonic/player/dao/RatingDao.java
```diff
@@ -104,7 +104,7 @@ public Double getAverageRating(MediaFile mediaFile) {
      */
     public Integer getRatingForUser(String username, MediaFile mediaFile) {
         try {
-            return getJdbcTemplate().queryForObject("select rating from user_rating where username=? and path=?", new Object[]{username, mediaFile.getPath()},Integer.class);
+            return getJdbcTemplate().queryForObject("select rating from user_rating where username=? and path=?", username, mediaFile.getPath(),Integer.class);
```
eharris: You changed the number of parameters.

eharris:
Turns out I was wrong, these aren't safe to do after all.

PR #1136, opened 2019-06-15
Gracefully handle failed version check attempts

jvoisin:
This issue what found by darkling, and reported on irc

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/VersionService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/VersionService.java
@@ -25,6 +25,7 @@ import org.apache.commons.io.IOUtils;
 import org.apache.http.client.ResponseHandler;
 import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.HttpGet;
+import org.apache.http.conn.ConnectTimeoutException;
 import org.apache.http.impl.client.BasicResponseHandler;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClients;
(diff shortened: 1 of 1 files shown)


airsonic-main/src/main/java/org/airsonic/player/service/VersionService.java
```diff
@@ -243,6 +244,11 @@ private void readLatestVersion() throws IOException {
         try (CloseableHttpClient client = HttpClients.createDefault()) {
             ResponseHandler<String> responseHandler = new BasicResponseHandler();
             content = client.execute(method, responseHandler);
+        } catch (ConnectTimeoutException e) {
+          LOG.error("Got a timeout when trying to reach {}", VERSION_URL);
```
eharris: Maybe this ought to be a warning rather than error?

airsonic-main/src/main/java/org/airsonic/player/service/VersionService.java
```diff
@@ -243,6 +244,11 @@ private void readLatestVersion() throws IOException {
         try (CloseableHttpClient client = HttpClients.createDefault()) {
             ResponseHandler<String> responseHandler = new BasicResponseHandler();
             content = client.execute(method, responseHandler);
+        } catch (ConnectTimeoutException e) {
+          LOG.warn("Got a timeout when trying to reach {}", VERSION_URL);
+          latestFinalVersion = getLocalVersion();
+          latestBetaVersion = getLocalVersion();
```
muff1nman: We should not set these two variables. Leaving them unset should make the UI not display a new version being available.

muff1nman:
Indentation should be 4 spaces.

eharris:
This needs to be rebased on master, which would pull in the checkstyle checks.

Issue #1144, opened 2019-06-25
Failed lyrics fetch causes huge traceback in log


eharris:
## Problem description

Looks like I'm getting failures of lyrics fetches (in my case the status code seems to be 500, Internal Server Error).  when this happens, a huge traceback is generated in the log.

I'm attaching a couple of log examples.

## System information

 * **Airsonic version**: 10.4.0-SNAPSHOT (latest master)
 * **Operating system**: Debian 9 stretch
 * **Java version**: java-8-openjdk-amd64
 * **Proxy server**: Apache SSL reverse proxy
 * **Client**: Happens from other clients in addition to webapp

## Additional notes

[catalina.out.log](https://github.com/airsonic/airsonic/files/3323392/catalina.out.log)

eharris:
This might be a problem with the `%27` quotes in the lookup.  If I remove those manually and query for the lyrics, the lookup seems to work and doesn't generate an `Internal Server Error`

Row:
The chartlyrics SearchLyricDirect API seems to be broken, it is not possible to find songs that contain the single quote character. Unfortunately, thier contact form seems broken too.

E.g.
http://www.chartlyrics.com/hHZoWzStn0q0JQ56qB-9OA/The+Mighty+Doors+of+the+Speargod's+Hall.aspx

I tried a couple of different ones:
http://api.chartlyrics.com/apiv1.asmx/SearchLyricDirect?artist=Amon+Amarth&song=The+Mighty+Doors+of+the+Speargod's+Hall

http://api.chartlyrics.com/apiv1.asmx/SearchLyricDirect?artist=Amon+Amarth&song=The+Mighty+Doors+of+the+Speargod's+Hall

http://api.chartlyrics.com/apiv1.asmx/SearchLyricDirect?artist=Amon+Amarth&song=The+Mighty+Doors+of+the+Speargod%27s+Hall

http://api.chartlyrics.com/apiv1.asmx/SearchLyricDirect?artist=Amon+Amarth&song=The+Mighty+Doors+of+the+Speargods+Hall

Perhaps it's time to switch to another API?

andyforceno:
Might I suggest the [Genius API](https://docs.genius.com/)? From an ideal user perspective, it would be nice to have a few options for lyrics providers to choose between. In reality, I know the amount of work that takes for the developers. If I knew the first thing about Java, I would contribute, but sadly, I do not.

PR #1145, opened 2019-06-25
Handle LyricsService exceptions

eharris:
This fixes the huge traceback problem when there is an error returned from the lyrics server that is exhibited in #1144, but doesn't address the underlying problem of why the lyrics fetches are failing in this manner.


eharris:
I think the integration tests for this only failed due to a since-resolved error?

jvoisin:
Rebase on master to re-trigger the CI run to prove it?

eharris:
done.

PR #1151, opened 2019-07-01
Precompile jsp

### LINKED ISSUE #1150: Experiment with pre-compiling JSP
… with the expectation that it would reduce the first loading times for each new page visited after a restart (as suggested by @muff1nman on IRC).

muff1nman:
Fixes #1150

### DIFF
--- a/airsonic-main/pom.xml
+++ b/airsonic-main/pom.xml
@@ -638,6 +638,27 @@
                 <groupId>net.nicoulaj.maven.plugins</groupId>
                 <artifactId>checksum-maven-plugin</artifactId>
             </plugin>
+            <plugin>
+                <groupId>org.eclipse.jetty</groupId>
+                <artifactId>jetty-jspc-maven-plugin</artifactId>
+                <version>9.4.19.v20190610</version>
+                <executions>
+                    <execution>
+                        <id>jspc</id>
+                        <goals>
+                            <goal>jspc</goal>
+                        </goals>
+                        <phase>compile</phase>
+                        <configuration>
+                            <!-- Since we are using spring boot - we do not have a web.xml so there is no sense
+                            enabling merging. Furthermore, we output the webXmlFragment to the target/classes
+                            directory such that it will be on the classpath for use during runtime -->
+                            <mergeFragment>false</mergeFragment>
+                            <webXmlFragment>${project.build.outputDirectory}/precompiled-jsp-web.xml</webXmlFragment>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
         </plugins>
     </build>
     <profiles>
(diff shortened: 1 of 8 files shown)


airsonic-main/pom.xml
```diff
@@ -631,6 +631,24 @@
...
+                <version>9.4.19.v20190610</version>
+                <executions>
+                    <execution>
+                        <id>jspc</id>
+                        <goals>
+                            <goal>jspc</goal>
+                        </goals>
+                        <phase>compile</phase>
+                        <configuration>
+                            <mergeFragment>false</mergeFragment>
```
jvoisin: It would be nice a have a small comment to explain why this is set to false :)

airsonic-main/src/main/java/org/airsonic/player/spring/RegisterPrecompiledJSPInitializer.java
```diff
@@ -0,0 +1,65 @@
...
+
+@Component
+public class RegisterPrecompiledJSPInitializer implements ServletContextInitializer {
+
+    private static final Logger logger = LoggerFactory.getLogger(RegisterPrecompiledJSPInitializer.class);
+
+    @Override
+    public void onStartup(ServletContext servletContext) {
+
+        if(!SettingsService.isDevelopmentMode()) {
```
jvoisin: Can you please de-invert the condition to make it easier to read?

airsonic-main/src/main/java/org/airsonic/player/spring/RegisterPrecompiledJSPInitializer.java
```diff
@@ -0,0 +1,65 @@
...
+                IOUtils.toInputStream("</web-app>", Charset.defaultCharset()));
+        WebXml webXml = new WebXml();
+        boolean success = parser.parseWebXml(new InputSource(webXmlIS), webXml, false);
+        if (!success) {
+            throw new RuntimeException("Error parsing precompiled jsp XML " + precompiledJspWebXml);
+        }
+        for (ServletDef def :  webXml.getServlets().values()) {
+            logger.trace("Registering precompiled JSP: {} -> {}", def.getServletName(), def.getServletClass());
+            ServletRegistration.Dynamic reg = servletContext.addServlet(def.getServletName(), def.getServletClass());
+            reg.setLoadOnStartup(99);
```
jvoisin: Why this magic constant?

muff1nman: See https://stackoverflow.com/questions/809775/what-does-the-servlet-load-on-startup-value-signify. Basically it needs to be between 0 and 128. The 99 I believe is so that these servlets dont take precedence over more important things.

jvoisin: Can you add a small comment? :)

muff1nman:
RE: travis failures. Looks like the dependency checker is being helpful here. We should not rely on tomcat classes to do the webxml parsing otherwise this wont work outside of tomcat. Will address.

jvoisin:
This might be a stupid question: Are the precompiled jsp invalidated when airsonic is upgraded?

muff1nman:
They are class files inside the war - so yes.

muff1nman:
@jvoisin did you have any other concerns on this one? If not can you merge?

PR #1152, opened 2019-07-02
Suppress traceback for common null pointer exception

eharris:
This happens often with containers that support video but only contain audio.  The code raises the `NullPointerException` itself, so this just suppresses the traceback in that case.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/CoverArtController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/CoverArtController.java
@@ -628,10 +628,10 @@ public class CoverArtController implements LastModified {
             try {
                 in = getImageInputStreamForVideo(mediaFile, width, height, offset);
                 BufferedImage result = ImageIO.read(in);
-                if (result == null) {
-                    throw new NullPointerException();
+                if (result != null) {
+                    return result;
                 }
-                return result;
+                LOG.warn("Failed to process cover art for " + mediaFile + ": {}", result);
             } catch (Throwable x) {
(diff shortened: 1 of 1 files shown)


airsonic-main/src/main/java/org/airsonic/player/controller/CoverArtController.java
```diff
@@ -632,6 +632,8 @@ public BufferedImage createImage(int size) {
                     throw new NullPointerException();
                 }
                 return result;
+            } catch (NullPointerException x) {
+                LOG.warn("Failed to process cover art for " + mediaFile + ": " + x);
```
muff1nman: I think this should just be changed such that the NPE isn't thrown in the first place. Maybe do:

```
if result != null:
  return result
else:
  LOG.warn...

eharris: The NPE is meant to catch (and warn) about the problem while still ending up falling back to the call to `createAutoCover()`

muff1nman:
Actually looking at this again, any NPE is already caught by the `catch (Throwable x)` clause since a NPE is a Throwable. So this change doesn't do anything...

eharris:
Look again. The parameter list to the warn is different for the NPE case.

muff1nman:
Ah I see it now, however using exceptions for logic like this is not a good practice IMO. It should just be written with an if statement.

eharris:
I agree, I just left it the same way since I didn't know if that was an intentional style.

eharris:
Changed to an if.

PR #1153, opened 2019-07-02
Added webm to default video file types

eharris:


eharris:
Doesn't look like it to me.  `webm` already has mime type assigned, and has since Sindre on 2016-05-1

Issue #1105, opened 2019-06-08
label element have a fixed width


jooola:
In the webapp, the `label` element have a fixed with of 5 em.

This make many things really ugly.

![image](https://user-images.githubusercontent.com/19195485/59149601-80206b80-8a17-11e9-93a1-c089e5db5ec3.png)
![image](https://user-images.githubusercontent.com/19195485/59149608-9fb79400-8a17-11e9-8436-6871f92930d1.png)

jooola:
@jvoisin By curiosity why did you removed the issue label ?

jvoisin:
Because most of the issues are, well, … issues. I think that it makes more sense to consider that issues are issues by default, and t add labels to indicate otherwise.

PR #1154, opened 2019-07-02
Replace wz_tooltip with jquery-ui

jvoisin:
Since we're already using jquery-ui for other stuff, it makes sense to use it
for tooltips too, instead of using an old-school-written-in-2009 1300 lines
of steaming pile of javascript named wz_tooltip.js


jvoisin:
This time, I didn't break the settings as in #1105

airsonic-main/src/main/webapp/WEB-INF/jsp/helpToolTip.jsp
```diff
@@ -8,11 +8,28 @@ PARAMETERS
...
 <fmt:message key="common.help" var="help"/>
 
 <div id="placeholder-${param.topic}" style="display:none">
-    <div style="font-weight:bold;"><fmt:message key="helppopup.${param.topic}.title"><fmt:param value="Airsonic"/></fmt:message></div>
-    <div><fmt:message key="helppopup.${param.topic}.text"><fmt:param value="Airsonic"/></fmt:message></div>
+    <div style="font-weight:bold;">
+			<fmt:message key="helppopup.${param.topic}.title">
+				<fmt:param value="Airsonic"/>
+			</fmt:message></div>
+    <div>
```
muff1nman: shouldnt this be a closing tag?

airsonic-main/src/main/webapp/WEB-INF/jsp/helpToolTip.jsp
```diff
@@ -8,11 +8,28 @@ PARAMETERS
...
+</div>
 </div>
-<img src="${imageUrl}" alt="${help}" title="${help}" onmouseover="TagToTip('placeholder-${param.topic}', BALLOON, true, ABOVE, true, OFFSETX, -17, PADDING, 8, WIDTH, -240, CLICKSTICKY, true, CLICKCLOSE, true)" onmouseout="UnTip()"/>
+ <img src="${imageUrl}" alt="<fmt:message key="helppopup.${param.topic}.text"><fmt:param value="Airsonic"/></fmt:message>"
```
eharris: Single space indent?

airsonic-main/src/main/webapp/WEB-INF/jsp/helpToolTip.jsp
```diff
@@ -10,26 +10,26 @@ PARAMETERS
...
 });
 </script>
 
 <style>
-	div.ui-tooltip {
-		max-width: 400px;
-	}
+  div.ui-tooltip {
+    max-width: 400px;
+  }
```
eharris: These indents should be standard 4-space not 2.

eharris: Apologies for this.  I didn't see that the rest of the file is already 2 space indent.  Probably better to keep it consistent rather than just change these few lines.

jvoisin:
Shall we merge this one?

eharris:
I've been meaning to test this one but haven't had time.  I'll try to apply it to my repo later today and check it out.

eharris:
I applied this to test it out and there are some problems:

* Settings->General page, several rendering problems near "Index".
* Settings->Advanced page, the rendering near CAPTCHA has problems.

There may be others, this is just what I've noticed so far.

In addition, I have a few suggestions:

* I don't like the upward drop-shadow.  Can this be changed to downward instead?
* I would prefer if the background text color and maybe text and outline was different, to make the tooltip stand out more from the rest of the page.  Maybe keep the yellow background and different text style from the original tooltip style?

jvoisin:
As I have poor design skills, I didn't touch the colours yet, but the escaping is fixed, and I removed the ugly dropshadow.

eharris:
I went ahead and merged the systemd/tomcat9 commit since it is unrelated to the tooltip stuff

eharris:
I checked the latest changes, and the page layout issues seem to be fixed.  However, now the embedded formatting tags in the help strings aren't being rendered (e.g. `<p>,<em>,<br>`).  An example of this is on the `General` page in the help for the `Index` field, but there are several others.

jvoisin:
I'm not really super-duper confident with allowing html markup inside tooltips. Shall I simply edit the messages? Or are we sure that we want html there?

eharris:
Some of the help messages are necessarily complex, and without formatting they just all get glommed together and are much harder to understand.  I think being able to have simple formatting is needed, and since I believe it was also allowed in the prior tooltips, it makes sense to preserve that capability.  We do still control the help text, so we can decide that we only allow simple formatting and not things like font changes, div's, etc.

jvoisin:
It should be working correctly now :)

airsonic-main/src/main/webapp/WEB-INF/jsp/helpToolTip.jsp
```diff
@@ -8,11 +8,41 @@ PARAMETERS
...
+<script>
+    $(function () {
+        $(document).tooltip({
+            content: function () { return $(this).prop('title');}
+        });
+    });
+</script>
+
+<style>
+    div.ui-tooltip {
```
fxthomas: Can this be put in the themes' CSS?

fxthomas:
Looks good to merge now!

PR #1162, opened 2019-07-07
Search test enhancements

tesshucom:
This is an additional test case for more detailed observation of the current status of the search function.
It helps to know how the behavior changes after Lucene update.


airsonic-main/src/test/java/org/airsonic/player/service/search/AnalyzerFactoryTestCase.java
```diff
@@ -423,7 +424,7 @@ public void testUax29() {
         assertEquals("988", terms.get(4));
         assertEquals("aria", terms.get(5));
 
-        // artist
+        // Underbars around words, ascii and semicolon.
```
jvoisin:
```suggestion
        // Underscores around words, ascii and semicolon.
```

airsonic-main/src/test/java/org/airsonic/player/service/search/AnalyzerFactoryTestCase.java
```diff
@@ -412,7 +413,7 @@ public void testUax29() {
          * Case using test resource name
          */
 
-        // title
+        // Semicolon , comma and hyphen.
```
jvoisin:
```suggestion
        // Semicolon, comma and hyphen.
```

airsonic-main/src/test/java/org/airsonic/player/service/search/AnalyzerFactoryTestCase.java
```diff
@@ -433,6 +434,72 @@ public void testUax29() {
         assertEquals("cafe", terms.get(3));
         assertEquals("zimmermann", terms.get(4));
 
+        // Underbars around words and slashes
```
jvoisin:
```suggestion
        // Underscores around words and slashes.
```

airsonic-main/src/test/java/org/airsonic/player/service/search/AnalyzerFactoryTestCase.java
```diff
@@ -433,6 +434,72 @@ public void testUax29() {
...
+        query = "_ID3_ARTIST_ Sarah Walker/Nash Ensemble";
+        terms = toTermString(query);
+        assertEquals(5, terms.size());
+        assertEquals("id3_artist", terms.get(0));
+        assertEquals("sarah", terms.get(1));
+        assertEquals("walker", terms.get(2));
+        assertEquals("nash", terms.get(3));
+        assertEquals("ensemble", terms.get(4));
+        
+        // space
```
jvoisin:
```suggestion
        // Space
```

airsonic-main/src/test/java/org/airsonic/player/service/search/SearchServiceSpecialGenreTestCase.java
```diff
@@ -0,0 +1,242 @@
...
+@ContextConfiguration(
+        locations = {
+                "/applicationContext-service.xml",
+                "/applicationContext-cache.xml",
+                "/applicationContext-testdb.xml",
+                "/applicationContext-mockSonos.xml" })
+@DirtiesContext(
+        classMode = DirtiesContext.ClassMode.AFTER_EACH_TEST_METHOD)
+/*
+ * Tests to clarify strings that can be used in genre.
```
jvoisin:
```suggestion
 * Tests to prove what kind of strings/chars can be used in the genre field.
```

airsonic-main/src/test/java/org/airsonic/player/service/search/SearchServiceSpecialGenreTestCase.java
```diff
@@ -0,0 +1,242 @@
...
+            System.out.println("--- *********************** ---");
+            dataBasePopulated = true;
+        }
+    }
+
+    @Test
+    public void testQueryEscapeRequires() {
+    
+        /*
+         * Legacy is not searchable. (FILE01 - FILE16)
```
jvoisin: I don't understand this comment :/

airsonic-main/src/test/java/org/airsonic/player/service/search/SearchServiceSpecialGenreTestCase.java
```diff
@@ -0,0 +1,242 @@
...
+            musicFolders = new ArrayList<>();
+
+            File musicDir = new File(resolvePath.apply("Search/SpecialGenre"));
+            musicFolders.add(new MusicFolder(1, musicDir, "accessible", true, new Date()));
+
+        }
+        return musicFolders;
+    }
+
+    private int count = 1;
```
jvoisin: Can't this variable be in `populateDatabase` ?

tesshucom: This is the count called from ``@Before``, so it is written there.

I wrote it with reference to other tests, but it seems that Scan may not be performed correctly if it is called almost simultaneously from multiple methods.
(A painful workaround ...)

airsonic-main/src/test/java/org/airsonic/player/service/search/SearchServiceSpecialGenreTestCase.java
tesshucom: surely.

 - There are 19 files in ``src/test/resources/MEDIAS/Search/SpecialGenre/ARTIST1/ALBUM_A``.
 -  In FILE01 to FILE16, Special strings for Lucene syntax are stored as tag values ​​of Genre.

I need a post-update comparison for future searches,
but there is no case because legacy does not handle this at all.

-> I should have noted down the failure cases. I will add.

airsonic-main/src/test/java/org/airsonic/player/service/search/SearchServiceSpecialGenreTestCase.java
jvoisin: Please add a small comment explaining this then :)

airsonic-main/src/test/java/org/airsonic/player/service/search/SearchServiceSpecialGenreTestCase.java
```diff
@@ -0,0 +1,309 @@
...
+        songs = searchService.getRandomSongs(simpleStringCriteria.apply("/"));
+        Assert.assertEquals(0, songs.size());
+
+    }
+
+    @Test
+    public void testBrackets() {
+        
+        /*
+         * Jaudiotagger applies special treatment to bracket (FILE17).
```
jvoisin: Aren't comments going before the function in Java?

airsonic-main/src/test/java/org/airsonic/player/service/search/SearchServiceSpecialGenreTestCase.java
tesshucom: I revised that according to your advice.

airsonic-main/src/test/java/org/airsonic/player/service/search/SearchServiceSpecialGenreTestCase.java
tesshucom: I revised that according to your advice.

airsonic-main/src/test/java/org/airsonic/player/service/search/SearchServiceSpecialGenreTestCase.java
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    /*
+     * A count to deliberately stagger the execution of populateDatabase.
+     * If populate Database is called from multiple methods almost simultaneously,
+     * it may not execute correctly.
+     */
+    private int count = 1;
+
+    private synchronized void populateDatabase() {
```
muff1nman: Rather than doing this, doesn't it make more sense to use unique airsonic home paths for each test method?

muff1nman: Also this method seems duplicated across these test classes. Can either move it to a utility class or add some inheritance.

tesshucom:
I agree to that.
For now, create a upper class to use in the package to resolve duplicates and synchronization.

airsonic-main/src/test/java/org/airsonic/player/service/search/SearchServiceSpecialGenreTestCase.java
tesshucom: Solved by inheritance.

> Rather than doing this, doesn't it make more sense to use unique airsonic home paths for each test method?

If the meaning is not to define MusicFoldsder ​​in each test class, but to define in MusicFolderTestData collectively, modify additionally.

There are some problems when included in the current MusicFolderTestData# getTestMusicFolders, so I think it will be in the form of adding new methods.

airsonic-main/src/test/java/org/airsonic/player/service/search/SearchServiceSpecialGenreTestCase.java
tesshucom: When the index created by executing SearchServiceSpecialGenreTestCase is confirmed in the viewer, it is as follows.

image

Legacy analyzes Genre with StandardAnalyzer, so some strings are discarded.
As a result, no Genre field is created in the index.
Asserts to confirm this implementation.

Issue #1142, opened 2019-06-22
Search result quality


CTParkinson:
## Problem description

Had the same issue with Subsonic, Searching is kinda garbage.
For example searching for "For all The Cows" - by Foo Fighters.
![image](https://user-images.githubusercontent.com/46909431/59967001-49b91500-951c-11e9-8607-2e47c41a0e49.png)

Why does a full text match not appear above everything else?

But worse than that, the artist name gets matched in song and album results, and also the album titles get matched in song results;
![image](https://user-images.githubusercontent.com/46909431/59967077-46725900-951d-11e9-8754-9394c2f97211.png)

literally every song by Lamb of God is listed, in the order they appear on the album, and i assume each album too, because the search term matches the artist name but not the song titles...

Since the artist results lists them already, unless the album contains the search term also I would expect to not see any results matching the band name in albums, same goes for song results.

Unless the item title matches the search, maybe don't show that in the results set?

### Steps to reproduce

1. Search for practically anything.

## System information

 * **Airsonic version**: 10.2.1-RELEASE – February 18, 2019
 * **Operating system**: Whatever DietPi is. Debian? Unsure which version.
 * **Java version**: I don't know
 * **Proxy server**: None

## Additional notes

I would like to use the search via the API, images were taken from Airsonic ui for familiarity

tesshucom:
Unfortunately, current search queries are not optimized for scoring.

> Searching is kinda garbage.

I think you should be very careful not to break past functions, but search functions are functions that need to be improved because they are considered to contain certain problems and potential bugs.
I think so too.


> Unless the item title matches the search, maybe don't show that in the results set?

That is also a strong valid solution, but the Artist displayed in the Songs line is the tag data of the song, so the content may differ from the Album Artist.

Therefore, in general, the method of adjusting the Boost value is taken.
The Boost value defaults to 1.
(For example, designate as in Song1.3 Album1.2 Artist1.1.)
That way you can't avoid a complete search, but you can increase the match score of the song title and show it on top.
(``Sorting`` is a slightly different concept.)

The current search feature includes adjustment of the Boost value, but is a dead code.
We need to improve this.

I have created a PR on #1130.
This is itself a prerequisite PR for search engine updates.
But I think it is also the first step in tackling these various search qualities.

For example, we are looking into enabling registration of multiple Boost values ​​for the index.
[...]

Issue #1139, opened 2019-06-18
Path to search in a specific case is incorrect


tesshucom:
This issue has spun off from #1130.
The purpose is to observe and consider the current specifications rather than the defect report.

 - How to handle folder path is somewhat gray for current search specifications.
 - Lucene upgrades have some impact on search accuracy.

## Problem description

This issue reports an example of a shuffle search false search.
In addition, it also mentions how folder paths are used when searching.

## System information

 * **Airsonic version**: 10.3.1-RELEASE – May 21, 2019
 * **Operating system**: Apache Tomcat/8.5.40(embeded), Windows 10
 * **Java version**: , java 1.8.0_201
 * **Proxy server**: None
 * **Client**: Google Chrome 75.0.3770.90 (Official Build)(64bit)

### Steps to reproduce

Create a MusicFolder with the following name.

![image](https://user-images.githubusercontent.com/27724847/59682137-8cbf7380-9210-11e9-9dd9-e9d63a600cdf.png)

Set.

![image](https://user-images.githubusercontent.com/27724847/59682384-17a06e00-9211-11e9-863e-2fc7b2468571.png)

It looks like it can be read properly when scanned.

![image](https://user-images.githubusercontent.com/27724847/59682422-2e46c500-9211-11e9-96bb-9f1bf37d4fc6.png)

Select ``Home> Random``.

![image](https://user-images.githubusercontent.com/27724847/59682485-4fa7b100-9211-11e9-9909-1675bdf23e90.png)

Let's select Music1.
Somehow all DIR is displayed.

![image](https://user-images.githubusercontent.com/27724847/59682538-677f3500-9211-11e9-8749-e05357d2365a.png)

Music2 is empty.

![image](https://user-images.githubusercontent.com/27724847/59682589-7fef4f80-9211-11e9-8b06-d574ecc06fe8.png)

Music3 is empty.

![image](https://user-images.githubusercontent.com/27724847/59682635-95fd1000-9211-11e9-9e1c-b6bff05910d1.png)

The reason for some unexpected behavior is as follows.
___

The Airsonic Lucene search uses ``SpanOr`` instead of the exact path specification by ``Or``.
``SpanOr`` is similar to a phrase, divided into words and evaluated in order.

The directory names are interpreted by Analyzer as follows:

 - ``accessible`` -> ``accessible`` (as it is)
 - ``accessible︴s`` -> ``accessible`` and ``s``
[...]

PR #1166, opened 2019-07-10
Update to Lucene 8.1.1

tesshucom:
Update Lucene from 3.0.3 to the current version 8.1.1.

Influence (user point of view)

 - There is almost no change.
  You can check it with SearchServiceTestCase.

 - Word breaks are changed to UAX#29 method.
  Details can be checked with AnalyzerFactoryTestCase.

 - Improved to filter correctly for input values.
   You can check with SearchServiceStartWithStopwardsTestCase.

Influence (internal)

 - The processing of Analyzer becomes strict.
 - Genre enables search by user readable value (tag value)
   Except in some cases, Index's Genre is compatible with databases.
   You can check with SearchServiceSpecialGenreTestCase.

The index created by the update can be viewed on the latest version of [Luke 8.0.0](https://github.com/DmitryKey/luke/releases).


tesshucom:
Considering the preservation of git history before and after the update, merging is difficult.
This is because some of the current test cases are closely related to functions other than SearchService.

It is thought that the specification that was hidden because there was no SearchService test up to now has been revealed.
It's not a bad sign.

If we classify each test case and related processing blocks, it will be as follows.

Some are clearly mergeable, but some may need to be considered.
How to classify SearchServiceSpecialPathTestCase and SearchServiceSpecialGenreTestCase is bothersome.

image

airsonic-main/src/main/java/org/airsonic/player/service/search/AnalyzerFactory.java
```diff
@@ -47,75 +55,128 @@
...
+                Analyzer defaultAnalyzer = createDefaultAnalyzerBuilder().build();
+                Analyzer genreAnalyzer = createGenreAnalyzerBuilder().build();
 
-        @Override
-        public TokenStream tokenStream(String fieldName, Reader reader) {
-            TokenStream result = super.tokenStream(fieldName, reader);
-            return new ASCIIFoldingFilter(result);
-        }
+                Map<String, Analyzer> fieldAnalyzers = new HashMap<>();
+                fieldAnalyzers.put(FieldNames.GENRE, genreAnalyzer);
```
tesshucom: Commenting out this process has no effect on existing web screens or clients.
Therefore, it may be considered unnecessary processing, but it is a relatively important decision in the specification.

tesshucom: The reason for stricting the genre string is to change the domain value from the previous mutable to immutable.
This means that genre reference on DB is possible via domain value.

We can cope with difficult requirements like #105.
While multivalue processing on DB involves very complex implementations, doing this with Lucene is relatively safe and fast.

Conversely, the requirement to change domain design from immutable to mutable, should be cautious.
As in #1132.

fxthomas:
I don't have a lot of time to review the code at the moment, especially since I don't know much about Lucene, but I've pulled that on my server and am observing how it behaves. So far I haven't seen big bugs, great job!

One minor thing I've noticed (although I'm not sure if that was not the case before your PR) : the default rules appear to completely ignore common English articles (the, a, an, etc.) and auxiliaries (will, be), but it's not the case for French ones (le, la, les, une). For example, if I'm searching for John Williams, "wil" and "willi" will list Mr. Williams as the first result, but "will" returns nothing (empty page).

tesshucom:
Stop words are generally designated words that appear frequently in newspapers and emails.
Because it depends on the language to search, the one that is suitable for Analyzer is selected.

By default, ENGLISH_STOP_WORDS_SET is adopted.
The contents are as stated below:

```
"a", "an", "and", "are", "as", "at", "be", "but", "by",
"for", "if", "in", "into", "is", "it",
"no", "not", "of", "on", "or", "such",
"that", "the", "their", "then", "there", "these",
"they", "this", "to", "was", "will", "with"
```
I think there are two issues related to Stopword.

 - Although Stopword is used for parsing at index creation, it is not applied to queries at search time.
 - Is the current Stopword suitable for music search, Airsonic?

The fact that it is not applied to the search query is generally a defect.
Therefore, I fixed it when defining a new Analyzer.
The change in test results shows this.

Changing the content of the stop word is not a bug fix but a specification change.
Therefore it is not included in this PR.
It is as you pointed out.

For example, when the following is required : 

> As used in the index, stopwords are good only for typical articles from different countries. Let's define new Stopwords!
```
// There is a specification similar to the index!
DEFAULT_IGNORED_ARTICLES = "The El La Los Las Le Les"; 
```
Fixing itself is easy.

tesshucom:
#1139 bug seems to be a traditional bug ( [SearchServiceSpecialPathTestCase](https://github.com/airsonic/airsonic/blob/master/airsonic-main/src/test/java/org/airsonic/player/service/search/SearchServiceSpecialPathTestCase.java) ) .

 - Also occurs in Airsonic v10.0.0
 - Also occurs in Latest version of Subsonic 6.1.5 (build 759abe) – December 1, 2018

___

Oddly, it does not occur in my clone.
I can handle it correctly.

[SearchServiceSpecialPathTestCase](https://github.com/jpsonic/jpsonic/blob/master/jpsonic-main/src/test/java/org/airsonic/player/service/search/SearchServiceSpecialPathTestCase.java)

I checked the difference for the solution, but at the moment I didn't have no idea at all.
It takes time to solve it.

___

I think that it is not necessary to associate it with the update of this library.
As a matter of fact, the cause of this failure has nothing to do with the processing flow of SearchService.
It is due to an error in the earlier processing.
The domain value of path is already mishandled before creating the Lucene index.

fxthomas:
Thanks for the detailed explanation, I think I understand better! In any case it's not really a big deal, and I agree with you on keeping it consistent between parsing and index creation.

eharris:
I just applied this to my tree and built it, and at startup I got a tracebacks about the search service failing due to the index being from an unsupported version.

```
2019-07-20 14:20:30.555 ERROR --- o.a.p.s.search.SearchServiceImpl         : Failed to search for random albums.

org.apache.lucene.index.IndexFormatTooOldException: Format version is not supported (resource BufferedChecksumIndexInput(MMapIndexInput(path="/var/lib/airsonic/lucene2/album/segments_ak"))): -9 (needs to be between 1071082519 and 1071082519). This version of Lucene only supports indexes created with release 6.0 and later.
```

So I tried to update the index by doing a forced rescan.  That generated tons of errors similar to this:

```
2019-07-20 14:24:07.488 ERROR --- o.a.p.s.search.IndexManager              : Failed to create search index for Emmanuel

        at org.airsonic.player.service.search.IndexManager.index(IndexManager.java:107) ~[classes/:10.5.0-SNAPSHOT]
        at org.airsonic.player.service.search.SearchServiceImpl.index(SearchServiceImpl.java:65) [classes/:10.5.0-SNAPSHOT]
        at org.airsonic.player.service.MediaScannerService.scanFile(MediaScannerService.java:240) [classes/:10.5.0-SNAPSHOT]
...
```

Users should not have to take any special steps to be able to upgrade their existing installs to a version using the new Lucene version.

I think there needs to be some way of auto-updating the index when the version is incompatible, or at least better handling of the consequences when the version is incompatible.  For example, rather than generating thousands of errors due to NPE's, it should refuse to attempt the update, or maybe automatically set the old index aside when it is incompatible.  Or maybe the new format index should be in files with a different name so that you can easily move between versions that use the old and new Lucene without having to completely reindex your whole library.

eharris:
Also, all the new lines of code contain `^M` line endings, so that needs to be fixed.

tesshucom:
You're right. It is already mentioned in #1130.
After a library update, at least 2 PRs are required to reach a practical release.
Apart from improving search accuracy.

 - Reader Improvement. This halves the initial overhead time of the search. Eliminate potential bugs that are not closed.
 - Index generation change. This is an already existing implementation. You need to fix it to work properly. It is after the modification of Reader because lock processing by framework is added.

PR division is being performed because of the scale.

``^M`` noticed for the first time.
It may be related to starting to use Github's client to support LFS.
to correct.

eharris:
@tesshucom #1130 has already been applied and closed a couple of weeks ago.  Which additional PR's should be applied in order to have something practical to test?

tesshucom:
Since the test has already been performed on the master, no special patch other than this PR is required.

#1130 and #1162 have already been applied to the master.
The reason why the test is pre-added to the master is to clarify the current specification and clarify the processing that is affected by the update.

Like #1139, there is also a test case that is not directly caused by Lucene but has a strong relevance.
At the discretion of the maintainer, the current implementation may be modified prior to the update.

If you want to add a test case, please do a PR for the master.

eharris:
@tesshucom I don't understand.  You stated above:

> After a library update, at least 2 PRs are required to reach a practical release.

Then you stated:

> no special patch other than this PR is required.

Which PR's should I be testing with this one to see that the end user usability issues I raised above have been addressed?

tesshucom:
These PRs will be added after these pass approvals.
This is because of division by PR size and dependencies.

eharris:
If there are additional commits/PR's needed, I would rather see them added to this or have this done as part of a feature branch that includes them.

tesshucom:
I can not do mainstream branch operations.
Because it is a large-scale correction, I have been talking about division unit of PR etc.
I respect the views of the main maintainers.

eharris:
You can add multiple commits to this PR, and that does not lessen the ability to review them, yet lets you build on your previous work.  That acts similarly to a feature branch.  Although it would probably be good to change the PR title to be more general.

tesshucom:
I understand what you say.
I am not against adding other processes to this PR.
However, the code modifications involved in the search are large.
It has been proposed to reduce PR size by the previous conversation.

There are already many correction points in this PR.
In the changes so far, are there any places that are likely to invite discussion?
At least @muff1nman has no reaction.

eharris may be waiting.
But I am also waiting.

muff1nman:
Sorry for the delay in responding on this one. I am part of the way through reviewing it. In the meantime to add my two cents:

1. Smaller PRs are preferable to larger ones as they are easier to reason about, test and revert.
2. We should not break master. Tests or actual runtime. 

So in line with 2, I would agree with @eharris that an index format change (if any, again I'm still reviewing this) should be transparent to the user and included in the PR.

muff1nman:
@tesshucom I imagine we can add some logic to automatically reindex when an older index is in place. What do you think?

muff1nman:
@tesshucom I'm also okay with that being a separate PR if it makes the changeset clearer, but I imagine that PR would be a requirement before this one could get merged.

tesshucom:
I understand.
Thank you for the polite review.

In parallel, add two more fixes.
It will be a little longer, but they are inseparable from the library update.

 - Reader Improvement. -> Fix speedup and bug of closing fixes. Index directory lock will be applied as well as DB.
 - Index generation change.  -> To work properly. Or to work correctly with directorylock flow.

> @tesshucom I imagine we can add some logic to automatically reindex when an older index is in place. What do you think?

I think this process is essential.
And it is forgotten because the index was not updated, but it already exists.

[MediaScannerService](https://github.com/airsonic/airsonic/blob/master/airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java)#deleteOldIndexFiles() 

As it is already included in the design, I think that it is good to do the repair or operation check of this.
However, it's arguable whether to fix it a bit or change it to what Airsonic prefers.

 - It is not suitable that the product name is included in the root directory. Set lucene-> index according to the directory of db
 - Where do you keep version literals? The schema version of DB is managed by MediaFileDao. The same as this. It is probably a reasonable design that IndexManager will manage. MediaScannerService's SearchService import should be IndexManager import by nature.

etc.

Reader Improvement is just a technical replacement.
I think that Index generation change should reflect the maintainer's thought more strongly than my opinion.
I do not seek an immediate answer.
I think that it is good to be careful because it becomes a long-term specification.

muff1nman:
Ah perfect, so really we should just need to bump INDEX_VERSION to 16 and if I'm understanding the logic correctly, it should create a new index on bean creation.

tesshucom:
> it should create a new index on bean creation.

Specifically, it is easy.
Synonymous with a change of  literal directory name.

It will be resolved if the root directory is changed when IndexWriter is created.
Currently it is ``(/var/airsonic/)lucene2``.
If this is replaced with ``lucene3`` (confusing) ...  a new Index will be created naturally.

muff1nman:
Hm let me check the logic again.

muff1nman:
Ah so I would guess that the deleteOldIndexFiles is no longer needed because at some point the index storage changed from individual files to directories based on the fact that getIndexFile isn't called by anything other than deleteOldIndexFiles. Would you agree with that assesment?

If that's the case I would be for fixing up deleteOldIndexFiles to work again but with the `lucene2` directory.

muff1nman:
Maybe we can use the org.apache.lucene.util.Version class to get the major lucene version so that way the new directory will be lucene7.

tesshucom:
Even when managing directories, it is a fact that logic equivalent to deleteOldIndex is required.
Expected results have already been presented by eharris.

There are risks when there is a fix in the index implementation of the lucene library and when Airsonic's schema or data changes.

lucene loves destructive updates.
Minor version changes may force index regeneration.

 - If the directory name is only the major version, you may be troubled if deleteOldIndex can not be done intentionally
 - In the future, even if it is decided that the library will not be changed frequently, deleteOldIndex is required to change the index design as well as the function improvement.

muff1nman:
Let me see if I understand.

You are suggesting that there are other circumstances when the index might need regeneration and that deleteOldIndex should handle this logic. Furthermore, just having the lucene major version stored in the directory is inadequate to implement the necessary logic. 

So lets just store the metadata about the index generation somewhere else. Maybe as a file json inside the directory? It could hold airsonic version, lucene version, etc.

eharris:
Considering how many revisions of lucene that airsonic has missed, I'm not too worried about updates of the lucene version breaking the index support often, since I would suspect that once we get the lucene version up to something semi-current, the need to apply further updates will be much lower.

Unless there's a good reason not to, I think my current preference for the index update handling would be to have any updates of the lucene version use a unique index directory, and to not have the code make any attempt to delete/cleanup/recreate old index versions.  This would allow for easily upgrading and downgrading the airsonic version without requiring the index to be completely rebuilt, which can be quite time consuming for any reasonably large library size.  This is important to me because it is not uncommon that I switch between newer and older versions of airsonic several times a day to check for differences in the behavior between the versions while I am working on issues or features.

tesshucom:
If you want to refine, there are also various approaches.

There is also a way not to include the product version simply.
It is a method to manage whether data rebuilding should be done at release time with INDEX_VERSION only.
Root directory is permanently fixed.
Database management is based on this idea.

Lucene has a migrate mechanism for close version.
Therefore, there is a possibility that the implementation had been concerned about the product version(guess).

Migrating is a smart way if users keep up with server updates.
It's ridiculous, but if you're not sure if migration is valid then only data removal is sure.

In the case of data deletion premise, the need to struggle to manage product versions is reduced.
Rebuilding the Lucene Index is not a very long process.
It does not involve rebuilding the DB, so if the DB has already been scanned, initialization is a short time.

tesshucom:
Fixed the line feed code.
So far, there has been no change in processing since the first PR.

tesshucom:
Reader has been improved.
I do not do profiling with JMeter because I do not have much data at hand.
But this will be faster than the currently released search feature.

The framework's directory lock will work and you will not be able to delete the index directory  as before while it's running.



#### SearchServiceImpl.java 

There is no difficult correction.

 - Get a searcher using getSearcher
 - Release after using the searcher

This is important.
```
        if(isEmpty(searcher)) {
            return result;
        }
```

The case where searcher is null is included in normal processing.
If you do a search before conducting a scan at the first start, searcher is null.

From the user's point of view:
If you do a search before conducting a scan at the first start,  it is appropriate for the web screen to display a blank (or not found) instead of displaying a Java error.

#### IndexManager.java

Reader fixes are summarized in IndexManager.
Of particular note is the refresh of the sercher in stopIndexing.

Readers are not consumed for each search as before, but are kept reference by SearcherManager while the server is running.
(The file is not expanded in memory, it is a reference.)
Therefore, if index rewriting is performed, explicit updating is required.

From the user's point of view:
 - Scan after Startup
 - File Update  and Scan

 It is OK if search can be performed using updated values.

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
```diff
@@ -70,110 +72,165 @@
...
+    public @Nullable IndexSearcher getSearcher(IndexType indexType) {
+        if (!searchers.containsKey(indexType)) {
+            synchronized (searchers) {
+                File indexDirectory = util.getDirectory.apply(getIndexRootDirectoryName(), indexType);
+                try {
+                    if (indexDirectory.exists()) {
+                        SearcherManager manager = new SearcherManager(FSDirectory.open(indexDirectory.toPath()), null);
+                        searchers.put(indexType, manager);
+                    } else {
+                        LOG.warn("{} does not exist. Please run a scan.", indexDirectory.getAbsolutePath());
```
tesshucom: If you search before scan, warning.
Need to fix message level or log level?

airsonic-main/src/main/java/org/airsonic/player/service/search/AnalyzerFactory.java
```diff
@@ -1,121 +1,182 @@
...
+     * 
+     * @return analyzer for index
+     * @see DocumentFactory
+     */
+    public Analyzer getAnalyzer() {
+        if (isEmpty(analyzer)) {
+            try {
+                analyzer = createDefaultAnalyzerBuilder().build();
+            } catch (IOException e) {
+                LOG.error("Error when initializing Analyzer.", e);
```
muff1nman: Why discard the error here? Maybe wrap it with a runtime exception and let it bubble up?

airsonic-main/src/main/java/org/airsonic/player/service/search/AnalyzerFactory.java
```diff
@@ -1,121 +1,182 @@
...
+     * The processing of the operands is expressed
+     * in the AnalyzerFactory implementation.
+     * Rules for tokenizing/converting input values ​
+     * should not be described in QueryFactory.
+     * 
+     * @return analyzer for query
+     * @see QueryFactory
+     */
+    public Analyzer getQueryAnalyzer() {
+        if (isEmpty(queryAanalyzer)) {
```
muff1nman: Typo on `queryAanalyzer`.

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
```diff
@@ -1,169 +1,236 @@
...
+     * Return the IndexSearcher of the specified index.
+     * At initial startup, it may return null
+     * if the user performs any search before performing a scan.
+     * 
+     * @param indexType
+     * @return
+     */
+    public @Nullable IndexSearcher getSearcher(IndexType indexType) {
+        if (!searchers.containsKey(indexType)) {
+            synchronized (searchers) {
```
muff1nman: Instead of using synchronized directly, can you instead use `computeIfAbsent`. Then you don't have to worry about putting the synchronized in the right spot (I don't think it is here).

airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceUtilities.java
```diff
@@ -1,200 +1,187 @@
...
+                dist.add(album);
+            }
+        }
+    };
+
+    private final Function<String, File> getRootDirectory = (version) -> {
+        return new File(SettingsService.getAirsonicHome(), version);
+    };
+
+    public final BiFunction<String, IndexType, File> getDirectory = (version, indexType) -> {
```
muff1nman: Some of these functions such as this one should probably be package private (remove `public` keyword). Actually I'd make this whole class package private to really drive home the point that it shouldn't be exposed and is an implementation detail.

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
```diff
@@ -1,169 +1,236 @@
...
+        IndexWriterConfig config = new IndexWriterConfig(analyzerFactory.getAnalyzer());
+        return new IndexWriter(FSDirectory.open(dir.toPath()), config);
+    }
+
+    /**
+     * Close Writer of all indexes and update SearcherManager.
+     * Called at the end of the Scan flow.
+     */
+    public void stopIndexing() {
+        Arrays.asList(IndexType.values()).forEach(i -> stopIndexing(i));
```
muff1nman: Can use method reference here. I.e. replace `i -> stopIndexing(i)` with `this::stopIndexing`.

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
```diff
@@ -1,169 +1,236 @@
...
+
+    @Autowired
+    private DocumentFactory documentFactory;
+
+    @Autowired
+    private SearchServiceUtilities util;
+
+    private Map<IndexType, SearcherManager> searchers = new ConcurrentHashMap<>();
+
+    private Map<IndexType, IndexWriter> writers = new ConcurrentHashMap<>();
```
muff1nman: Can you explain the need for the concurrent hash map here? Can you detail which entrypoints/methods into IndexManager may be called concurrently?

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
muff1nman: Warn should be fine here.

muff1nman:
> Considering how many revisions of lucene that airsonic has missed, I'm not too worried about updates of the lucene version breaking the index support often, since I would suspect that once we get the lucene version up to something semi-current, the need to apply further updates will be much lower.

I would hope the idea would be to stay up to date with lucene and from reading about their changes, we'd have to trash the index for any major version change which has happened 5 times since lucene 3x was put in.

As far as knowing when to trash the index and start over, I'm thinking we should keep it simple for now and use a similar implementation to how `deleteOldIndexFiles` used to work. Basically use a new unique directory whenever we need to within a new version.

To address @eharris concern:

> Unless there's a good reason not to, I think my current preference for the index update handling would be to have any updates of the lucene version use a unique index directory, and to not have the code make any attempt to delete/cleanup/recreate old index versions

~~I think we can have a "hidden" configuration setting (not exposed in the UI) for disabling automatic cleanup.~~

On second thought my index is only 1.6MB. That's like nothing. Maybe we just leave out automatic deletion for now entirely. We should still use a unique directory though.

airsonic-main/pom.xml
```diff
@@ -197,6 +201,15 @@
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
         </dependency>
+		<dependency>
+			<groupId>org.checkerframework</groupId>
+			<artifactId>checker-qual</artifactId>
+			<!-- Guava uses it implicitly.
+				When PRing, it is necessary to confirm
+				whether null annotation is permitted. -->
+			<scope>provided</scope>
```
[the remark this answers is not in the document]
tesshucom: There is no problem in ``compile``.
As a result, both are the same.

In the case of ``compile``, it is explicitly included in the product.
On the other hand, in the case of an annotation library, the option of ``provided`` is available if it is used for compilation and not included in the product.
(At present, Airsonic has no view of what to use from many nullness libraries)

The result is the same because Airsonic includes the ''provide'' library in the product for embedded.

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
tesshucom: I only keep in mind add/remove for map.
(Such as multiple calls from REST)
Lucene classes such as SearcherManager are thread safe.

tesshucom:
I'll correct the point pointed out.
After a while, I'll try to fix deleteOldIndexFiles with the policy of cleanup&fixed directory.
It is very simple and considered understandable by anyone.

As muff1nman says, the index will be more smaller with the same content to improve compression.
IO/memory efficiency is also improved.
Therefore, it may result in slightly different preferable results than those guessed by heavy users' past experience.

I think that it is not bad to make a simple logic and to evaluate it from a new viewpoint.

eharris:
@tesshucom It looks like you merged in changes from master.  I think it would be better if you rebased instead.

eharris:
@tesshucom I would also strongly suggest that you separate out the reformatting (fixing of `^M`) into a separate commit from the code changes.

airsonic-main/src/main/java/org/airsonic/player/service/search/QueryFactory.java
```diff
@@ -1,231 +1,291 @@
...
+     *  XXX 3.x -> 8.x :
+     *  "SpanOr" has been changed to "Or".
+     *   - Path comparison is more appropriate with "Or".
+     *   - If "SpanOr" is maintained, the DOC design needs to be changed.
+     */
+    private final BiFunction<@NonNull Boolean, @NonNull List<MusicFolder>, @NonNull Query> toFolderQuery = (
+            isId3, folders) -> {
+        BooleanQuery.Builder mfQuery = new BooleanQuery.Builder();
+        folders.stream()
+        	.map(isId3 ? toFolderIdQuery : toFolderPathQuery)
```
eharris: There's a tab/indent problem here.

eharris:
Specific field searches don't seem to work.  I tried searching for "title:blah", "album:blah", "artist:blah", "genre:blah" and got no results.  I also tried quoted strings, and those didn't seem to work either, as I get many more results than I should.  "And" and "Or" also don't seem to work.

Can you explain what search query options are expected to be supported?

tesshucom:
@eharris 
Maybe you are misunderstanding.
Query handled in the Search implementation is an internal language for search engines.
Same as database SQL.
As SQL does, it is not meant for control directly by the user.

Test case of QueryFactory was created to clarify legacy formulas and compare them to new versions of formulas.
This is due to the fact that the query interpretation of the search engine is relatively less version-dependent than the API implementation.

It easier to detect correction is made that affects the search expression making or upgrading again also in the future .

eharris:
I understand that the internal query language is different than the external (user seen) query language.  But what is implemented in the internals defines what is available to be used by the user.  Since you seem to be very familiar with what Lucene provides and how Airsonic is using it, I'm asking what search query features are expected to be usable by the user, because I have never seen any documentation or specification on what search features are (meant to be) exposed.

For instance, I would expect there to be at a minimum phrase matching via quotes (e.g. `"adjacent words"`), but that doesn't seem to work.  I would hope that specific field matching (e.g. `title:blah` would be provided as well.  Even better would be AND and OR (e.g. `artist:"Iron Maiden" and title:"where eagles"`) but none of these seem to work.

tesshucom:
Airsonic, like Subsonic, does not support special search query input.

It is not impossible to respond.
(If we are willing to make it quite complicated)

Although we can make it as convenient as we like, the web screen is structured to be paired with the REST API.
If we want functional improvement, I think that the improvement that can be common to WEB REST UPnP Sono is the priority.

On the screen, API is performing search equivalent to Seach2.
This search specializes in returning results from many fields very quickly by OR nesting.
It's actually amazingly fast.

However, the present condition is not enough as a function.
The example given in #1142 is a dissatisfaction point that everyone feels.

There are various ways to get around this, one is the method of query control like eharris says.
We may want to narrow down the target fields as mentioned in #1142 or we may want to be able to specify when searching.

These are effective, but there is a problem that the added function is incompatible with the existing UI, or it is also difficult to use.

The impact on existing implementations is low, and the one that is effective is the adjustment of scoring.
This is a common sense approach, and can be said to be a mystery not implemented in the current search.
(Even from the perspective of consistency)
This is the safest fastest destination we can offer initially to people who find it difficult to search.
It's easy to add and can be removed immediately.


Although off the topic of this PR, the improvement of the complaint is after the update.
It will be easier for others besides me to work after the update.

airsonic-main/src/main/java/org/airsonic/player/service/search/AnalyzerFactory.java
tesshucom: I revised that according to your advice.

airsonic-main/src/main/java/org/airsonic/player/service/search/AnalyzerFactory.java
```diff
@@ -47,75 +55,128 @@
 
     private Analyzer analyzer;
 
-    private Analyzer queryAnalyzer;
+    private Analyzer queryAanalyzer;
```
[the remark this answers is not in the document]
tesshucom: I revised that according to your advice.

airsonic-main/src/main/java/org/airsonic/player/service/search/QueryFactory.java
tesshucom: I revised that according to your advice.

airsonic-main/src/main/java/org/airsonic/player/service/search/AnalyzerFactory.java
tesshucom: I revised that according to your advice.

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
tesshucom: It's exactly as you said.
Furthermore, getSearcher is beautiful for Optional.


The barrier is lambda's exception handling.
We can do more things by writing lambdas that throw exceptions.
However, in the vicinity of search there are a lot of IOExceptions.
And if we introduce it, the source code will be quite different.

After thinking for a while, I prioritized the old try catch and avoid adopting new mechanisms sharply.
If a common view of Airsonic is shown for lambda exceptions, I rewrite this PR and follow it in all subsequent code.
This feels like a policy that is included in everything else, not just this PR.

I've removed synchronized.

tesshucom:
> @tesshucomマスターからの変更をマージしたようです。代わりにリベースすればもっと良いと思います。

It is force push after rebase.
I'll do not force push after this,  let go the final merge in the way of muffinman.

I wondered if I would try again, but it seems that there is a problem with travis at the same time.

[Past Incidents](https://www.traviscistatus.com/)
After this my travis environment can not handle some POMs correctly.
Subsequent commits seem to be working fine with Airsonic's cache on Airsonic side travis .
I do not want to pull in new problems, so I will try to avoid retroactive retries.

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
tesshucom: I revised that according to your advice.

airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceUtilities.java
tesshucom: I revised that according to your advice.

tesshucom:
@muff1nman

Fixed ``deleteOldIndexFiles``.
It can be said that the change commits necessary to merge with master are complete.

I recognize that it takes quite a while for the review to take place the overall volume is high.
I accept the correction until you are convinced.

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -54,7 +53,7 @@
     @Autowired
     private SettingsService settingsService;
     @Autowired
-    private SearchService searchService;
```
tesshucom: MediaScannerService does not search.
SearchService changes to IndexManager.

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -44,7 +44,6 @@
 @Service
 public class MediaScannerService {
 
-    private static final int INDEX_VERSION = 15;
```
tesshucom: Now that IndexManager has been created, it is appropriate to handle INDEX_VERSION with IndexManager.
This variable has moved.

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -383,43 +382,10 @@ public MediaLibraryStatistics getStatistics() {
...
-     * Returns the index file for the given index version.
-     *
-     * @param version The index version.
-     * @return The index file for the given index version.
-     */
-    private File getIndexFile(int version) {
-        File home = SettingsService.getAirsonicHome();
-        return new File(home, "airsonic" + version + ".index");
-    }
-
```
tesshucom: In the previous review, file operations related to Index were encapsulated in IndexManager.
This also moves to IndexManager.

airsonic-main/src/main/java/org/airsonic/player/service/SearchService.java
```diff
@@ -33,24 +32,14 @@
...
-
-    void index(MediaFile mediaFile);
-
-    void index(Artist artist, MusicFolder musicFolder);
-
-    void index(Album album);
-
-    void stopIndexing();
-
     SearchResult search(SearchCriteria criteria, List<MusicFolder> musicFolders,
```
tesshucom: If MediaScannerService references IndexManager instead of SearchService, there is no reason to expose Index related methods in the SearchService interface.
Removing these from the interface splits the search and indexing into a design.

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
```diff
@@ -62,14 +65,40 @@
 
     private static final Logger LOG = LoggerFactory.getLogger(IndexManager.class);
 
-    // XXX 3.x -> 8.x : #1166 Holds a tentative literal. It needs to be resolved by the release.
-    private static final String INDEX_ROOT_DIR_NAME = "lucene2";
```
tesshucom: Change the root directory name and fix it.
Index-related file name definitions are defined by the functions at the beginning of IndexManager.

airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceImpl.java
```diff
@@ -53,31 +53,6 @@
...
-    @Override
-    public void index(Album album) {
-        indexManager.index(album);
-    }
-
-    @Override
-    public void stopIndexing() {
-        indexManager.stopIndexing();
-    }
-
```
tesshucom: These delegate methods are no longer needed.

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
```diff
@@ -232,4 +261,79 @@ public void release(IndexType indexType, IndexSearcher indexSearcher) {
...
 
+    /**
+     * Check the version of the index and clean it up if necessary.
+     * Legacy type indexes (files or directories starting with lucene) are deleted.
+     * If there is no index directory, initialize the directory.
+     * If the index directory exists and is not the current version,
+     * initialize the directory.
+     */
+    public void deleteOldIndexFiles() {
+
```
tesshucom: It is as follows if it is rewritten as simple as possible.

eharris:
@tesshucom Thanks for your work on this.

This is a personal preference, but I would prefer if you rebased out the later commits that fix previous commits (things like fixing indent, style changes, typo fixes) so that the commits are clean.  This makes them easier to review and results in a cleaner commit history when it is merged.  Some maintainers handle this by squashing commits when a change is merged, but this is a complicated enough change that keeping distinct commits for the different parts would be nice, so I think it is worthwhile to clean up the commits to avoid the need to squash.

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
```diff
@@ -232,4 +261,79 @@ public void release(IndexType indexType, IndexSearcher indexSearcher) {
...
+                initializeIndexDirectory();
+
+            }
+        } else {
+
+            // Create current version directory at first start without index
+            initializeIndexDirectory();
+
+        }
+
```
eharris: Lots of extra unnecessary whitespace above

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
```diff
@@ -232,4 +261,79 @@ public void release(IndexType indexType, IndexSearcher indexSearcher) {
...
+            } else {
+                LOG.warn("Failed to create index directory :  (index version {}). ", INDEX_VERSION);
+            }
+        } catch (IOException e) {
+            LOG.error("Failed to create index directory :  (index version "
+                    .concat(Integer.toString(INDEX_VERSION))
+                    .concat("). "), e);
+        }
+    }
+
```
eharris: Unneeded whitespace

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
```diff
@@ -62,14 +65,40 @@
...
+     *  - When schema definition is changed due to modification of AnalyzerFactory,
+     *    DocumentFactory or the class that they use.
+     *
+     */
+    private static final int INDEX_VERSION = 16;
+
+    /**
+     * Literal name of index top directory.
+     */
+    private static final String INDEX_ROOT_DIR_NAME = "index";
```
eharris: Thanks for the directory name change.  But could I perhaps suggest using "search" rather than index?  Index could mean lots of things other than lucene.

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -69,13 +68,13 @@
 
     @PostConstruct
     public void init() {
-        deleteOldIndexFiles();
+        indexManager.deleteOldIndexFiles();
```
eharris: Re: https://github.com/airsonic/airsonic/pull/1166#issuecomment-515270408 and my own comment regarding keeping the old index when upgrading/downgrading airsonic versions, I agree with providing and correcting the `indexManager.deleteOldIndexFiles()` method, but I do not want it to be used automatically.

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -69,13 +68,13 @@
...
         statistics = settingsService.getMediaLibraryStatistics();
         schedule();
     }
 
     public void initNoSchedule() {
-        deleteOldIndexFiles();
+        indexManager.deleteOldIndexFiles();
```
eharris: Same here, would prefer not to call `indexManager.deleteOldIndexFiles()` automatically.

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
eharris: It might also be noted in a comment here that the directory name should be changed (like with a version suffix) whenever a change that causes index incompatibility is made.  In fact, maybe it should be made `search3` now so that is clear since the last one was `lucene2`?

tesshucom: Indexes seem to be appropriate from past sources, but let me ask muff1nman to decide.

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
tesshucom: This is probably inevitable.
As a system specification, regardless of the disk size, there is a debate as to whether to adopt rotate or not.
Some people hate rotation and some hate deletion.

I personally do not mind either, but I think that it may be taken into consideration that index schema changes are not frequent, and that some people prefer automation.

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
tesshucom: That?
I think it was to be a literal, but my English ability may have a misunderstanding.
Ask muff1nman for confirmation.

tesshucom:
>  but this is a complicated enough change that keeping distinct commits for the different parts would be nice, so I think it is worthwhile to clean up the commits to avoid the need to squash.

As it is quite a long-term battle, there are various ways to deal with it.
Rebase is relatively easy, as there is no deletion/add of files.
Therefore, at this stage we will keep a history to avoid misunderstanding.

It is recommended to deal with that time, because unnecessary comments may be deleted, test cases may be merged, etc.

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
eharris: Yes, it is probably inevitable, and I wouldn't have a problem with automatic cleanup once we are a few more release versions (of airsonic) ahead.  I just don't want it automatic when the change is so recent, since I often change between the last few versions to test things.

eharris:
> It is recommended to deal with that time, because unnecessary comments may be deleted, test cases may be merged, etc.

I just prefer to try to keep the needed work for the maintainer to merge to a minimum.  They don't seem to have a lot of time, and I don't want that to delay this getting merged.

tesshucom:
At the request of muff1nman, I will accept any cooperation.
Until then I'll avoid irreversible update operations.

I understand that you want to rush, but the load on me and muff1nman for this PR is inherently high.
It is because it is trying to solve various problems that can not be achieved even with Subsonic and to give approval to it.
Merge is not a bottleneck.

It is a problem that has been neglected for decades because of its difficulty.
If it takes weeks or months to validate, I will not complain.

If we want to accelerate the whole movement, we need a higher quality reviewer between me and the muffin man.
Or parallel validation to measure performance problems is also useful.
If a third party does a code break and adds a new test to the master, I will incorporate it into this PR.

Constructive cooperation that is highly useful in the future is highly welcome.

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
tesshucom: I have made relevant comments on this matter.
It is necessary to strategically consider practical issues in technical issues and rather long-term maintenance, not emotional theory.

Unless there is anyone who can make revolutionary proposals, policy choices by the elimination method may be necessary.
What we can do is to present new information that will be used to make muffmanman decisions.

airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceUtilities.java
```diff
@@ -1,200 +1,176 @@
...
-        return new Term(FieldNames.ID, Integer.toString(artist.getId()));
-    };
-
-    public final Term createPrimarykey(MediaFile mediaFile) {
-        return new Term(FieldNames.ID, Integer.toString(mediaFile.getId()));
-    };
-
-    public final boolean addIgnoreNull(Collection<?> collection, Object object) {
-        return CollectionUtils.addIgnoreNull(collection, object);
-    }
```
tesshucom: This is somewhat weird code by design.

It does the same (compare with equals) as with legacy, but does not have duplicate input values ​​when this code is called.
It suggests that there was some anxiety at the time of implementation, there were uninvestigated problems, or there was a simple misunderstanding.

- If we want to compare strictly, we should compare id.
- Even if you compare ids, this code will not work on the data flow.

If this code is removed and duplicate results are displayed, it indicates that the defect exists prior to the processing of the search.
There is no problem even if it is left as it is, but it is ultimately unnecessary code.

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
```diff
@@ -1,169 +1,339 @@
...
+
+    public void index(Artist artist, MusicFolder musicFolder) {
+        Term primarykey = documentFactory.createPrimarykey(artist);
+        Document document = documentFactory.createArtistId3Document(artist, musicFolder);
+        try {
+            writers.get(IndexType.ARTIST_ID3).updateDocument(primarykey, document);
+        } catch (Exception x) {
+            LOG.error("Failed to create search index for " + artist, x);
+        }
+    }
```
tesshucom: Mysterious duplicate checks exist because legacy implementations have not paid attention to the uniqueness of index documents.
Currently, documents with the same id do not exist in the search index, so the search search will not contain duplicate ids.

Unlike in the past, this point is guaranteed, so there is no need to check for duplicates after a search.
(There is no problem if you leave it, but if you are concerned that it may cause misunderstanding, it is better to delete it.)

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
```diff
@@ -1,169 +1,339 @@
...
+        Document document = documentFactory.createAlbumId3Document(album);
+        try {
+            writers.get(IndexType.ALBUM_ID3).updateDocument(primarykey, document);
+        } catch (Exception x) {
+            LOG.error("Failed to create search index for " + album, x);
+        }
+    }
+
+    public void index(Artist artist, MusicFolder musicFolder) {
+        Term primarykey = documentFactory.createPrimarykey(artist);
```
tesshucom: Mysterious duplicate checks exist at search result creation because legacy implementations have not paid attention to the uniqueness of index documents.
Currently, documents with the same id do not exist in the search index, so the search result will not contain duplicate ids.

Unlike in the past, this point is guaranteed, so there is no need to check for duplicates after a search.
(There is no problem if you leave it, but if you are concerned that it may cause misunderstanding, it is better to delete it.)

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
tesshucom: Rebased the latest master and changed it to pass all CheckStyle.

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
tesshucom: Rebased the latest master and changed it to pass all CheckStyle.

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
tesshucom: Same as above. Underlying is the issue of Airsonic's management policy.

airsonic-main/src/main/java/org/airsonic/player/service/search/DocumentFactory.java
```diff
@@ -57,7 +57,7 @@
     static {
 
         TYPE_ID = new FieldType();
-        TYPE_ID.setIndexOptions(IndexOptions.NONE);
+        TYPE_ID.setIndexOptions(IndexOptions.DOCS);
```
tesshucom: The document design may be different from the legacy, but it is a mandatory modification.
An Index is required to successfully retrieve the primary key.
In other words, this modification prevents double registration.

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
```diff
@@ -187,7 +187,7 @@ private void stopIndexing(IndexType type) {
         boolean isUpdate = false;
         // close
         try {
-            isUpdate = -1 == writers.get(type).commit();
+            isUpdate = -1 != writers.get(type).commit();
```
tesshucom: It's a simple code mistake.
The judgment as to whether or not a commit has been made is incorrect.
This fix allows Sercher refresh to work correctly.
Corrects the operation when multiple scans are executed, which is not in a single case.

tesshucom:
Since there is no code difference, update from 8.1.1 to 8.2.0.

In the fix so far, data flow consistency is expected to be guaranteed.
Judging that it was practical, I released it on my clone.
(However, Japanese language processing and further functions have been added.)

Some people are running it.
No problems have been reported by the user.

____

The only problem I know is that JavaScript causes bugs in genres that contain double quotes.

There are several ways to deal with this.
Ultimately, JavaScript correction is desirable, but if it cannot be corrected immediately, double quotes can be excluded from the search index as a response to the transition period.

In general, genres rarely use double quotes.
It is not a serious problem because it is an expected side effect and there is a way to deal with it.
I don't know how much impact JavaScript has since I didn't hack the code.

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
muff1nman: `index` is fine, but like @eharris said I think it'd be good to have a version suffix. So `index3`?

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
muff1nman: Let's leave the automatic index deletion commented out for now. As per my previous comment, these index directories are not large in size. Therefore I see no penalty to keeping around the old index for a release or two and then adding automatic deletion in once the chance of a revert is much less.

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
muff1nman: Let's comment this out for now.

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
muff1nman: Or using the existing `INDEX_VERSION` constant would be fine too. So in that case the directory would be named `index16`

muff1nman:
So to summarize, I think we should keep this as simple as possible.

- Use `INDEX_VERSION` to keep each index directory unique. `INDEX_VERSION` is not tracked in a POM or anything like that. 
- Don't cleanup old index directories for now. We can incorporate that in a later PR.

So @tesshucom, if you find the above workable, please make the following changes:

- Comment out index directory deletion on `init`.
- Make the index directory name `"index"` + `INDEX_VERSION`
- Remove `indexVersionFile` - it shouldn't be needed with the above

tesshucom:
Noted with thanks.
I will correct it as you indicated after rebase.

If requested, I can delete comments.
(If it is not elegant. Written redundant for review.)

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
tesshucom: We have approved your request.

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
tesshucom: We have approved your request.

PR #1173, opened 2019-07-22
Fix for paths stored in Lucene index

tesshucom:
The fix is ​​for the Lucene index.
It does not affect the record of Database.

There is a place to switch Path by recursive processing in Scan, and it is correctly registered in Database.

The same is true for Lucene indexes, but the processing order is incorrect.
Incorrect path may be registered in Lucene index.

This fix allows Lucene indexes to handle the correct path perfectly matched with the database,.

The bug of #1139 is resolved.
This bug seems to be a traditional bug .

 - Also occurs in Airsonic v10.0.0
 - Also occurs in Latest version of Subsonic 6.1.5 (build 759abe) – December 1, 2018

___

If this is merged before #1166, test case modifications should be incorporated into #1166.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
@@ -237,13 +237,13 @@ public class MediaScannerService {
 
         LOG.trace("Scanning file {}", file.getPath());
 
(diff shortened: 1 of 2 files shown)


tesshucom:
I created a test case because I was suspicious of the search behavior, but the direct cause was in Scan.
I created a PR.

PR #1174, opened 2019-07-23
Fixed null exception when attempt to stream non-file

eharris:
There are MediaFile entities (like directories) that would cause a null pointer exception if they were attempted to stream.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/StreamController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/StreamController.java
@@ -128,7 +128,7 @@ public class StreamController {
             // In that case, create a separate playlist (in order to support multiple parallel streams).
             // Also, enable partial download (HTTP byte range).
             MediaFile file = getSingleFile(request);
-            boolean isSingleFile = file != null;
+            boolean isSingleFile = file != null && file.isFile();
             HttpRange range = null;
 
             if (isSingleFile) {


tesshucom:
Can you suggest a concrete operation example as a case where a directory is passed as an argument?

eharris:
@tesshucom It is easily done via the REST API `id=` parameter.

tesshucom:
The specification is defined as follows:
```
A string which uniquely identifies the file to stream. Obtained by calls to getMusicDirectory.
```
A string which uniquely identifies the file.
By design, the directory is not included in the premise of the input value.

There is no place to reproduce on the screen.
Correct?

eharris:
Yes, as far as I know there isn't a way to reproduce this problem using the built-in web client.  However, clients using the REST API are not necessarily well-behaved, and the REST API should be robust enough not to have unhandled exceptions even when a client attempts to do something outside the spec.

I would actually prefer that `handleRequest()` return an error when either an improper `MediaFile` or `Playlist` id is provided, but since I do not know all the use cases of this function and it is already written in a permissive manner, it seemed safer in the short term to just avoid the exception.

tesshucom:
(Generally thinking)

As defined in the parameter specification, there is no problem even if it is as it is.
Of course error with message are better.
In this case, NullPointerException or IllegalArgumentException with a message is preferred.

(If you look at the history of Subsonic's implementation of Rest)

It has been arepeated modifications to wrap with the Rest message instead of the general exception.
As you can see by comparing the exception design of the entire Rest, the places where general exceptions are thrown are limited.
For example, when an access error to an external resource, or when a permission error occurs in the scan.

If you want to be consistent with other implementations, it will return the error of Rest in the method specified in the ``Error handling`` specification.

jvoisin:
People are doing stupid things all the time, so it's highly likely that, as @eharris said, a misbehaving client will throw folders at this endpoint, and I'm convinced that gracefully explicitly handling this error, instead of yielding a null-pointer-exception (NPE) is the way to go.

Unhandled exceptions should be considered as a bug, and shouldn't be yielded, no matter what users are feeding the REST API.

muff1nman:
We should be failing early as possible.

@eharris What was the behavior you observed client side when the NPE occurred? I would hope that  the fallback error handler for the REST controller kicked in and returned a 500 error.  Also can you post the stack trace as well?

tesshucom:
I might have overlooked a bit.
There is processing required after a very long block.
It is the same opinion that should shift to exception control.

eharris:
@muff1nman here are the requested sample traceback and error response.

Error response:
```
<?xml version="1.0" encoding="UTF-8"?>
<subsonic-response xmlns="http://subsonic.org/restapi" status="failed" version="1.15.0">
   <error code="0" message="NullPointerException"/>
</subsonic-response>
```

Traceback:
[catalina.out.log](https://github.com/airsonic/airsonic/files/3429965/catalina.out.log)

fxthomas:
If we're changing existing behavior, shouldn't we explicitly return a correct HTTP error code (e.g. 400 Bad Request) along with a message?

muff1nman:
Might make sense to add a simple reproducing test for this. Either in the docker based integration tests or if its possible in the unit tests.

fxthomas:
Does something like this sound good?

```diff
diff --git a/airsonic-main/src/test/java/org/airsonic/player/api/AirsonicRestApiIntTest.java b/airsonic-main/src/test/java/org/airsonic/player/api/AirsonicRestApiIntTest.java
index 723b3c42..9834c3d3 100644
--- a/airsonic-main/src/test/java/org/airsonic/player/api/AirsonicRestApiIntTest.java
+++ b/airsonic-main/src/test/java/org/airsonic/player/api/AirsonicRestApiIntTest.java
@@ -1,7 +1,14 @@
 package org.airsonic.player.api;
 
 import org.airsonic.player.TestCaseUtils;
+import org.airsonic.player.dao.MediaFileDao;
+import org.airsonic.player.dao.MusicFolderDao;
+import org.airsonic.player.domain.MediaFile;
+import org.airsonic.player.domain.MusicFolder;
+import org.airsonic.player.service.MediaScannerService;
+import org.airsonic.player.service.SettingsService;
 import org.airsonic.player.util.HomeRule;
+import org.airsonic.player.util.MusicFolderTestData;
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
 import org.junit.Test;
@@ -13,6 +20,8 @@ import org.springframework.http.MediaType;
 import org.springframework.test.context.junit4.SpringRunner;
 import org.springframework.test.web.servlet.MockMvc;
 
+import java.util.Date;
+
 import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
 import static org.springframework.test.web.servlet.result.MockMvcResultHandlers.print;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
@@ -34,6 +43,18 @@ public class AirsonicRestApiIntTest {
     @Autowired
     private MockMvc mvc;
 
+    @Autowired
+    private MediaFileDao mediaFileDao;
+
+    @Autowired
+    private MusicFolderDao musicFolderDao;
+
+    @Autowired
+    private SettingsService settingsService;
+
+    @Autowired
+    private MediaScannerService mediaScannerService;
+
     @ClassRule
     public static final HomeRule classRule = new HomeRule(); // sets airsonic.home to a temporary dir
 
@@ -42,6 +63,36 @@ public class AirsonicRestApiIntTest {
         AIRSONIC_API_VERSION = TestCaseUtils.restApiVersion();
     }
 
+    @Test
+    public void directoryStreamErrorTest() throws Exception {
+
+        MusicFolderTestData.getTestMusicFolders().forEach(musicFolderDao::createMusicFolder);
+        settingsService.clearMusicFolderCache();
+        TestCaseUtils.execScan(mediaScannerService);
+
+        MediaFile testMediaFile = null;
+        for (MusicFolder musicFolder : musicFolderDao.getAllMusicFolders()) {
+            for (MediaFile mediaFile : mediaFileDao.getChildrenOf(musicFolder.getPath().toString())) {
+                testMediaFile = mediaFile;
+            }
+        }
+
+        mvc.perform(get("/rest/stream")
+                .param("id", String.format("%d", testMediaFile.getId()))
+                .param("v", AIRSONIC_API_VERSION)
+                .param("c", CLIENT_NAME)
+                .param("u", AIRSONIC_USER)
+                .param("p", AIRSONIC_PASSWORD)
+                .param("f", EXPECTED_FORMAT)
+                .contentType(MediaType.APPLICATION_JSON))
+                .andExpect(status().isOk())
+                .andExpect(jsonPath("$.subsonic-response.status").value("failed"))
+                .andExpect(jsonPath("$.subsonic-response.version").value(AIRSONIC_API_VERSION))
+                .andExpect(jsonPath("$.subsonic-response.error.code").value("0"))
+                .andExpect(jsonPath("$.subsonic-response.error.message").value("NullPointerException"))
+                .andDo(print());
+    }
+
     @Test
     public void pingTest() throws Exception {
         mvc.perform(get("/rest/ping")
```

fxthomas:
Right now instead of an exception the whole stream request hangs. Is the case `!file.isFile()` used somewhere? Would it be better to return an error directly?

PR #1186, opened 2019-07-31
Replace a now-defunct repo with a working one

jvoisin:
The http://teleal.org/m2 maven repository doesn't exist anymore (the domain is now a parking), so we should use http://4thline.org/m2 (which is the upstream for cling anyway) instead. But since the later uses some modern CA for its TLS cert, we need to modify where the OpenJDK9 is getting its CA from, since the bundled ones are outdated. This idea was stolen from https://www.deps.co/guides/travis-ci-latest-java/

This commit also remove `https://download.java.net/maven/2/`, since it's also dead: 404 all the way.

PR #1192, opened 2019-08-06
Add a test skip temporarily

tesshucom:
In connection with #1186.
Skip testing the search package.

While this is effective for temporarily avoiding the problem at hand, it is not a fundamental solution.
In my environment, the first build failed with Java8 and Java9.

Clipboard02

Achieve green after retrying several times.

Clipboard03

Even with this support, the build time seems too long.

___

Structurally, It seems that docker-test and verification (or Integration-test) for each JVM  is sufficient.
Most unit tests should only be done on a representative JVM, and in some cases it may be better to divide them into layers and run them in parallel.
(View/Controller/Model?)

Or define a test category.
(REST/Service/Database/Scan/Stream... etc?)
In this case, annotation management is generally preferred.
It is desirable to have a test utility project that holds annotations.

The travis build may be particularly slow from the end of last month.
We might want to hear muff1nman's opinion on test management.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -322,6 +322,10 @@
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-surefire-plugin</artifactId>
                     <version>2.22.0</version>
+                    <configuration>
+                        <!-- #1186 Add a test skip temporarily -->
+                        <excludes>org.airsonic.player.service.search.*TestCase.java</excludes>
+                    </configuration>
                 </plugin>
                 <plugin>
                     <groupId>net.nicoulaj.maven.plugins</groupId>


jvoisin:
The tests are slow because the caches are cold: travis-ci is doing some caching of the maven packages, I purged them because of #1186

jvoisin:
I don't understand why it's failing with Java9 :/

tesshucom:
Oh, it may be a little different from my environmental condition.

> jackson-databind-2.9.9.jar: CVE-2019-14379

I remember #1187 mentioning CVE-2019-12086, but in any case, patch #1187 seems to have a higher priority.
As far as you can see the log, it seems reasonable to rerun after the jackson update

tesshucom:
> liquibase-core-3.6.3.jar: bootstrap.min.js (pkg:javascript/bootstrap@3.1.1) : CVE-2018-14040, CVE-2018-14041, CVE-2018-14042, CVE-2019-8331
> liquibase-core-3.6.3.jar: bootstrap.js (pkg:javascript/bootstrap@3.1.1) : CVE-2018-14040, CVE-2018-14041, CVE-2018-14042, CVE-2019-8331
> liquibase-core-3.6.3.jar: jquery-1.11.0.min.js (pkg:javascript/jquery@1.11.0.min) : CVE-2015-9251, CVE-2019-11358

jvoisin:
It looks like travis is broken, and high for java8:

```
Caused by: com.fasterxml.jackson.core.exc.InputCoercionException: Numeric value (1565126694914) out of range of long (-9223372036854775808 - 9223372036854775807)
```
1565126694914 is in range…

tesshucom:
I don't think it's a serious problem, but it looks like it needs to be investigated.
Jackson's mapping is often a miracle.

It seems that CVE was updated at the same time as PR, but there seems to be a little troublesome problem for applicable library.
All of these need to be resolved in the near future, but immediate resolution is difficult.
I thought that if it was a minor problem, it could be handled at the same time as this PR.
But it ’s not the right amount to take that approach.

I think it's not bad to add temporary restraint.
The green build has been blocked for too long.

Github warnings are provided quite a bit ahead of CVE for severe problems.
Jackson seems crazy, but there is probably a workaround.
I agree with the idea of ​​thinking about how to solve #1187 separately in parallel with the temporary workaround.

jvoisin:
I fixed the build with 49c6e89f, 8260fed1, b650ac70 and 9a5f3f94

file-struct:
Thank you.

PR is already solved, but I will add information.

Information was added to [Past Incidents](https://www.traviscistatus.com/).


> Resolved - This incident has been resolved. 
> Aug 7, 14:40 UTC
> Investigating - We are receiving reports of Linux and Windows builds where network connectivity is unreliable, intermittently working or slow. We are opening this incident to acknowledge the issue and also keep you posted on our work to solve it. Thank you for your patience. 
> Aug 6, 17:55 UTC

A series of mysterious behaviors was likely affected by this.

tesshucom:
Thank you.

PR is already solved, but I will add information.

Information was added to [Past Incidents](https://www.traviscistatus.com/).


> Resolved - This incident has been resolved. 
> Aug 7, 14:40 UTC
> Investigating - We are receiving reports of Linux and Windows builds where network connectivity is unreliable, intermittently working or slow. We are opening this incident to acknowledge the issue and also keep you posted on our work to solve it. Thank you for your patience. 
> Aug 6, 17:55 UTC

A series of mysterious behaviors was likely affected by this.

PR #1195, opened 2019-08-07
Remove the useless Zoom link

jvoisin:
POUET

- People can just click on the picture, having a dedicated link for this is obnoxious
- Clicking on the picture provides a nice zoom effect, while keeping airsonic in the background, instead of displaying only the picture, interrupting the playback.

### DIFF
--- a/airsonic-main/src/main/webapp/WEB-INF/jsp/coverArt.jsp
+++ b/airsonic-main/src/main/webapp/WEB-INF/jsp/coverArt.jsp
@@ -103,21 +103,13 @@ PARAMETERS
     </c:if>
 </div>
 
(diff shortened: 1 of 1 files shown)


eharris:
The problem is that the image-click zoom doesn't show the image in full size, the displayed size is limited to an area smaller than the current frame.  Try it with a very large cover-art image and you'll see.

I agree that I don't like the link navigating away from the player though.  Perhaps instead you could change the Zoom link to open the image in a new tab/window, which shouldn't interrupt the playback?

eharris:
Hmm, I just tested this, and in Chrome, the Zoom link doesn't cause playback to stop, as it is loaded within the same frame.

PR #1197, opened 2019-08-08
Enable additional checkstyles for whitespace

eharris:


tesshucom:
@eharris

I like this approach.
I think eharris knows well and has worked hard.

Speaking dare, I think that checkstyle discussions and immediate code correction are not necessary in the individual issues review before such rules are added in advance.
Because the issus conversation will be unnecessarily long.

The intention of the update can be grasped by such a history that the code is naturally corrected with checkstyle merge.
I think the silent checkstyle update process contributes greatly to the quality of reviews.

PR #1200, opened 2019-08-08
Enable checkstyle for simple boolean expressions

eharris:
There may need to be some discussion about this one, as it is more subjective.


jp-searching:
Can you summarize the history related to this?
It is self-evident because it is written in a file.

PR #1204, opened 2019-08-11
Improve error handling for jaudiotaggerParser artwork fetch

eharris:
I saw a NPE error from my airsonic instance today due to `jaudiotaggerParser.getArtwork()` returning null, so this moves the later access attempts within the existing try/catch block.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/CoverArtController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/CoverArtController.java
@@ -292,12 +292,12 @@ public class CoverArtController implements LastModified {
             try {
                 LOG.trace("Reading artwork from file {}", mediaFile);
                 artwork = jaudiotaggerParser.getArtwork(mediaFile);
+                is = new ByteArrayInputStream(artwork.getBinaryData());
+                mimeType = artwork.getMimeType();
             } catch (Exception e) {
                 LOG.debug("Could not read artwork from file {}", mediaFile);
                 throw new RuntimeException(e);
             }
-            is = new ByteArrayInputStream(artwork.getBinaryData());
-            mimeType = artwork.getMimeType();
         } else {
             is =  new FileInputStream(file);
             mimeType = StringUtil.getMimeType(FilenameUtils.getExtension(file.getName()));


tesshucom:
It is reasonable to use isImageAvailable.
Unnecessarily NPE avoidance often complicates cause tracking.

eharris:
`isImageAvailable` only checks if the tag is not present, which was already handled.  This change now handles if the tag is present, but the contents are empty/malformed.

tesshucom:
OK. I think what you think is right.

> if the tag is present, but the contents are empty/malformed

What I thought was whether this NPE was a deserving exception thrown.
If exception handling of getArtwork is appropriate, it will be easier to understand.

```
    public Optional<Artwork> getArtwork(MediaFile file) {

        AudioFile audioFile = null;

        try {
            LOG.trace("Reading artwork from file {}", file);
            audioFile = AudioFileIO.read(file.getFile());
        } catch (CannotReadException | IOException | TagException | ReadOnlyFileException
                | InvalidAudioFrameException e) {
            LOG.debug("Could not read artwork from file {}", file);
            throw new RuntimeException(e);
        }

        Tag tag = audioFile.getTag();
        if (isEmpty(tag)
                || isEmpty(tag.getFirstArtwork())
                || isEmpty(tag.getFirstArtwork().getBinaryData())) {
            return Optional.empty();
        }
        return Optional.of(tag.getFirstArtwork());

    }
```
I don't object to PR because it works as it is.

tesshucom:
OK. I think what you think is right.

> if the tag is present, but the contents are empty/malformed

What I thought was whether this NPE was a deserving exception thrown.
For example, it may be written as:

```
    public Optional<Artwork> getArtwork(MediaFile file) throws IOException {

        AudioFile audioFile = null;

        try {
            LOG.trace("Reading artwork from file {}", file);
            audioFile = AudioFileIO.read(file.getFile());
        } catch (CannotReadException | IOException | TagException | ReadOnlyFileException
                | InvalidAudioFrameException e) {
            throw new IOException("Could not read artwork from file ".concat(file.getPath()), e);
        }

        Tag tag = audioFile.getTag();
        if (isEmpty(tag)
                || isEmpty(tag.getFirstArtwork())
                || isEmpty(tag.getFirstArtwork().getBinaryData())) {
            return Optional.empty();
        }
        return Optional.of(tag.getFirstArtwork());

    }
```

I don't object to PR because it works as it is.
I don't know (original source code) if there is a strong reason to wrap with ``RuntimeException``.

eharris:
I'm not sure why github is showing this as unmerged, but it was.

PR #1223, opened 2019-09-03
Added checkstyle to prohibit windows line endings

eharris:
Suggestion taken from jvoisin in #1121.


muff1nman:
Let's leave this till after #1166 to prevent unnecessary conflicts.

eharris:
Why would we want to wait?  I'd much rather not have whitespace problems be introduced than to have to clean them up later, touching more lines in the process.  It's easy enough for #1166 to be rebased on master (after this is merged) so that any checkstyle violations can be fixed.

eharris:
Ah nevermind, I understand now.  You mean because he's actually changing many of the lines in the files this would touch.

Issue #1218, opened 2019-08-29
FLAC : web player jumps to next song after wrong track length


kiv57:
## Problem description

Since 10.4.0 upgrade, the web player shows wrong track length for the flac track currently played (always randomly shorter).
After that wrong duration, it switches to the next track.
On that event the log shows :

<code>2019-08-29 18:57:07.198 INFO --- o.a.p.c.StreamController : x.x.x.x : Client unexpectedly closed connection while loading http://domain.tld/stream?player=1&id=23683 (java.io.IOException: Connexion ré-initialisée par le correspondant)
2019-08-29 18:57:07.209 INFO --- o.a.p.s.LoggingExceptionResolver : x.x.x.x : Client unexpectedly closed connection while loading http://domain.tld/stream?player=1&id=23683 (java.io.IOException: Relais brisé (pipe))</code>

### Steps to reproduce

- launch a flac album
- look at the track length increasing up to a wrong, shorter duration
- after that exact duration, player switches to next track

## System information

 * **Airsonic version**: 10.4.0
 * **Operating system**: Debian Stretch server
 * **Java version**: 1.8.0_201
 * **Proxy server**: None
 * **Client**: Firefox Version 68.0.2 (64 bits)

aeleeds:
I have the same problem with transcoding from Vorbis. It picks some arbitrary track length less than the actual song length. And then it keeps using that arbitrary track length for all subsequent tracks. (Stretch server; Macos 10.14.6; Chrome 76.0; Airsonic 10.4.0)

eharris:
It's a known issue with the MEJS player (which is used for the web-based media player) that if the reported Content-length is incorrect, the player will not continue playing past the given media size.

However, if transcoding is involved, then this report doesn't really make sense, since as of the 10.4 release, transcoded files should no longer report a length to the player.  Since there is no duration to (wrongly) stop at, the player should continue playing until the stream completes.

If transcoding is not involved, then the root cause of this issue is probably something causing Airsonic to believe the file is a different duration than it is, which probably means something related to the specific files (or tags) involved.  Can you provide some example media files that exhibit this problem?

kiv57:
Thank you @eharris for your reply.

> It's a known issue with the MEJS player (which is used for the web-based media player) that if the reported Content-length is incorrect, the player will not continue playing past the given media size.
> 
[...]

PR #1224, opened 2019-09-04
Add support for ListenBrainz Audio Scrobbler Service

ibmibmibm:
- Add textbox in user settings page for ListenBrainz token
- Add changes to db
    - Add db colume to store MusicBrainz Recording ID (improves #164)
    - Add db colume to store ListenBrainz token
- Add test for reading id
    - Add tag on testing file
- Add localization entry


ibmibmibm:
Try to make changes for Java8 now...

airsonic-main/src/main/java/org/airsonic/player/service/AudioScrobblerService.java
```diff
@@ -285,39 +381,66 @@ private RegistrationThread() {
...
                 RegistrationData registrationData = null;
+                int succeedCount = 0;
                 try {
                     registrationData = queue.take();
-                    scrobble(registrationData);
+                } catch (Exception x) {
+                    LOG.error("Failed to deque registration.", x);
+                }
+
+                LOG.info("Begin registration for " + registrationData.title + ".");
```
eharris: I would suggest all these `registrationData.title` values in log messages be (double-)quoted.

airsonic-main/src/main/java/org/airsonic/player/service/AudioScrobblerService.java
```diff
@@ -285,39 +381,66 @@ private RegistrationThread() {
...
+
+                try {
+                    scrobbleLastFm(registrationData);
+                    ++succeedCount;
+                    LOG.info("Succeed in Last.fm registration for " + registrationData.title + ".");
                 } catch (IOException x) {
-                    handleNetworkError(registrationData, x);
+                    LOG.warn("Network error in Last.fm registration for " + registrationData.title + ".", x);
```
eharris: The logging in these try..catch blocks probably doesn't need a full stack trace of the exception, which is what you get when you pass the exception to the log message.  I would suggest using `x.toString()` instead to get just the error details.

airsonic-main/src/main/resources/org/airsonic/player/i18n/ResourceBundle_en.properties
```diff
@@ -367,6 +367,8 @@ personalsettings.paginationsize=Number of related albums/directories to show ini
 personalsettings.lastfmenabled=Register what I'm playing at <a href="https://www.last.fm/" target="_blank">Last.fm</a>
 personalsettings.lastfmusername=Last.fm username
 personalsettings.lastfmpassword=Last.fm password
+personalsettings.listenbrainzenabled=Register what I'm playing at <a href="https://listenbrainz.org/" target="_blank">ListenBrainz</a>
```
eharris: There's no point in putting web links in the help messages, because they won't be clickable anyway (if you move the mouse away from the help icon, the tooltip will disappear).  Also, the help should only contain simple formatting tags, and only when necessary.

ibmibmibm: It's not inside a tooltip, but on the personal settings web page.

eharris: Ah, gotcha.  In that case, if an external link is needed, please keep in mind #712

ibmibmibm:
looks like #1218 ?

eharris:
@ibmibmibm What does that ticket have to do with this?

ibmibmibm:
> @ibmibmibm What does that ticket have to do with this?

Looks like that occurs in CI runs, too.

eharris:
The CI runs do not use a media player, they just pull the stream directly from the API.  We don't know yet why Travis is having issues, but there's no way that MEJS player issues are involved.

airsonic-main/src/main/java/org/airsonic/player/service/AudioScrobblerService.java
```diff
@@ -259,18 +272,101 @@ private String calculateAuthenticationToken(String password, long timestamp) {
...
             ResponseHandler<String> responseHandler = new BasicResponseHandler();
             String response = client.execute(request, responseHandler);
-            return response.split("\\n");
+            return response;
+        }
+    }
 
+    /**
+     * Scrobbles the given song data at listenbrainz.org, using the protocol defined at https://listenbrainz.readthedocs.io/en/latest/dev/api.html.
+     * ListenBrainz is not support NowPlaying notification.
```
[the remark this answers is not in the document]
ibmibmibm: I wasn't realized that ListenBrainz actually support NowPlaying notification. I add code for it, but forgot to remove this comment.

airsonic-main/src/main/java/org/airsonic/player/service/AudioScrobblerService.java
```diff
@@ -285,39 +381,66 @@ private RegistrationThread() {
...
+                } catch (Exception x) {
+                    LOG.error("Failed to deque registration.", x);
+                }
+
+                LOG.debug("Begin registration for '" + registrationData.title + "'.");
+
+                try {
+                    scrobbleLastFm(registrationData);
+                    ++succeedCount;
+                    LOG.info("Succeed in Last.fm registration for '" + registrationData.title + "'.");
```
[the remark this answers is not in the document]
ibmibmibm: Done. ListenBrainz's strings is also changed.

airsonic-main/src/main/java/org/airsonic/player/service/AudioScrobblerService.java
```diff
@@ -285,39 +381,66 @@ private RegistrationThread() {
...
                 } catch (IOException x) {
-                    handleNetworkError(registrationData, x);
+                    LOG.warn("Network error in Last.fm registration for '" + registrationData.title + "':" + x.toString());
                 } catch (Exception x) {
-                    LOG.warn("Error in Last.fm registration.", x);
+                    LOG.warn("Error in Last.fm registration for '" + registrationData.title + "'.", x);
```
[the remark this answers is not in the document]
ibmibmibm: Done.

airsonic-main/src/main/java/org/airsonic/player/service/AudioScrobblerService.java
```diff
@@ -259,18 +272,101 @@ private String calculateAuthenticationToken(String password, long timestamp) {
...
+                     " at ListenBrainz: " + registrationData.time);
         }
     }
+    /**
+     * Returns if submission succeeds.
+     */
+    private boolean submitListenBrainz(RegistrationData registrationData) throws Exception {
+        Map<String, Object> additional_info = new HashMap<String, Object>();
+        if (registrationData.musicBrainzReleaseId != null) {
+            additional_info.put("release_mbid", registrationData.musicBrainzReleaseId);
```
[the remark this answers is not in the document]
ibmibmibm: Done.

airsonic-main/src/main/java/org/airsonic/player/service/AudioScrobblerService.java
```diff
@@ -285,39 +370,62 @@ private RegistrationThread() {
...
+
+                try {
+                    scrobbleListenBrainz(registrationData);
+                    ++succeedCount;
+                    LOG.info("ListenBrainz registration successful for '" + registrationData.title + "'.");
+                } catch (Exception x) {
+                    LOG.warn("Error in ListenBrainz registration for '" + registrationData.title + "': " + x.toString());
+                }
+
+                if (succeedCount != 2) {
```
fxthomas: Can you make that number a documented constant?

airsonic-main/src/main/java/org/airsonic/player/service/AudioScrobblerService.java
```diff
@@ -285,39 +370,62 @@ private RegistrationThread() {
         public void run() {
             while (true) {
```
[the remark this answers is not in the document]
fxthomas: What happens if we stop retrying? Is the scrobble removed from the queue at some point?

airsonic-main/src/main/java/org/airsonic/player/service/AudioScrobblerService.java
```diff
@@ -285,39 +370,62 @@ private RegistrationThread() {
...
+                    LOG.error("Failed to deque registration: " + x.toString());
+                }
+
+                LOG.debug("Begin registration for '" + registrationData.title + "'.");
+
+                try {
+                    scrobbleLastFm(registrationData);
+                    ++succeedCount;
+                    LOG.info("Last.fm registration successful for '" + registrationData.title + "'.");
+                } catch (Exception x) {
```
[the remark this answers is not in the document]
fxthomas: Something could also happen on ListenBrainz' side (rate-limiting returning HTTP 502, for example). This probably won't raise `IOException`?

airsonic-main/src/main/java/org/airsonic/player/service/AudioScrobblerService.java
```diff
@@ -97,7 +99,7 @@ public synchronized void register(MediaFile mediaFile, String username, boolean
         try {
             queue.put(registrationData);
         } catch (InterruptedException x) {
-            LOG.warn("Interrupted while queuing Last.fm scrobble.", x);
+            LOG.warn("Interrupted while queuing Last.fm scrobble: " + x.toString());
```
fxthomas: Is that still uniquely used by Last.fm? Shouldn't that text be changed to remove that?

airsonic-main/src/main/java/org/airsonic/player/service/AudioScrobblerService.java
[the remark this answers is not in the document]
ibmibmibm: `queue.take` will make this thread wait until queue is not empty.
If we stop the worker thread, we have to resume it again whenever queue is not empty, and just like `queue.take` do.

ibmibmibm:
I've separated AudioScrobblerService into two helper class, to handle different audio scrobbler target.

airsonic-main/src/main/java/org/airsonic/player/service/AudioScrobblerService.java
```diff
@@ -18,309 +18,60 @@
...
 
 /**
  * Provides services for "audioscrobbling", which is the process of
- * registering what songs are played at www.last.fm.
- * <p/>
- * See http://www.last.fm/api/submissions
+ * registering what songs are played at website.
  *
- * @author Sindre Mehus
+ * @author Shen-Ta Hsieh
```
fxthomas: Seeing this I'm not sure about adding names to the source code. Down the road a file may be modified by a lot of people, so it's definitely not going to be one person ; plus it's available from the Git history.

airsonic-main/src/main/java/org/airsonic/player/service/AudioScrobblerService.java
```diff
@@ -18,309 +18,60 @@
...
-                    handleNetworkError(registrationData, x);
-                } catch (Exception x) {
-                    LOG.warn("Error in Last.fm registration.", x);
-                }
+        UserSettings userSettings = settingsService.getUserSettings(username);
+        if (userSettings.isLastFmEnabled() && userSettings.getLastFmUsername() != null && userSettings.getLastFmPassword() != null) {
+            if (lastFMScrobbler == null) {
+                lastFMScrobbler = new LastFMScrobbler();
             }
+            lastFMScrobbler.register(mediaFile, userSettings.getLastFmUsername(), userSettings.getLastFmPassword(), submission, time);
```
fxthomas: It would be better to pass uername and password through the constructor.

airsonic-main/src/main/java/org/airsonic/player/service/AudioScrobblerService.java
```diff
@@ -18,309 +18,60 @@
...
-            }
-            try {
-                sleep(60L * 1000L);  // Wait 60 seconds.
-            } catch (InterruptedException e) {
-                LOG.error("Failed to sleep after Last.fm registration failure for " + registrationData.title, e);
+        if (userSettings.isListenBrainzEnabled() && userSettings.getListenBrainzToken() != null) {
+            if (listenBrainzScrobbler == null) {
+                listenBrainzScrobbler = new ListenBrainzScrobbler();
             }
+            listenBrainzScrobbler.register(mediaFile, userSettings.getListenBrainzToken(), submission, time);
```
fxthomas: Same comment about token in the ctor. Make an interface with a `register(mediaFile, submission, time)` method?

airsonic-main/src/main/java/org/airsonic/player/service/scrobbler/ListenBrainzScrobbler.java
```diff
@@ -0,0 +1,232 @@
...
+import java.io.IOException;
+import java.util.*;
+import java.util.concurrent.LinkedBlockingQueue;
+
+/**
+ * Provides services for "audioscrobbling" at listenbrainz.org.
+ * <p/>
+ * See https://listenbrainz.readthedocs.io/
+ *
+ * @author Shen-Ta Hsieh
```
fxthomas: Same comment as before with the author name.

eharris:
I'd really like to see this get merged soon.  @ibmibmibm can you address @fxthomas concerns?  I will also try to allocate time to review in the next day or two.

eharris:
> I've separated AudioScrobblerService into two helper class, to handle different audio scrobbler target.

We really appreciate your work on this, it's really shaping up well!

That being said, I'd really like to see the code rearrangement (moving Last.fm to its own helper class) done as a separate commit (or even its own PR) before the changes to add ListenBrainz.

airsonic-main/src/main/java/org/airsonic/player/service/AudioScrobblerService.java
ibmibmibm: Is this service construct one for every users? I was guessing that this service object shared between all users, so I passed this data through register function.

airsonic-main/src/main/java/org/airsonic/player/service/AudioScrobblerService.java
ibmibmibm: Same comment above.

ibmibmibm:
> > I've separated AudioScrobblerService into two helper class, to handle different audio scrobbler target.
> 
> We really appreciate your work on this, it's really shaping up well!
> 
> That being said, I'd really like to see the code rearrangement (moving Last.fm to its own helper class) done as a separate commit (or even its own PR) before the changes to add ListenBrainz.

Now in separated commits.

airsonic-main/src/main/java/org/airsonic/player/service/AudioScrobblerService.java
fxthomas: Ah, I see, if you see it as a long-running service that's shared for everyone it makes sense.

What do you think of turning it into a real Spring service then? IMO this makes it more clear that it's intended this way and avoids the `if (lastFMScrobbler == null)` dance a few lines above.

fxthomas:
Thank you, great job on building this! :smile: Aside from my last comment we should be good to go.

eharris:
This has a merge conflict that needs to be resolved.

muff1nman:
Merged in 84df4e3b94cb60b7b710e8856c89d07a107047ec

PR #1229, opened 2019-09-11
Fixed bug where new users were given access to all media folders

eharris:
Newly created users were previously given access to all media folders before the user was updated to only allow access for the folders specified in the submitted form.
This change now defaults to no access to any folders, in case a crash or exception occurs between when the user is created and the later update of the allowed folders.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/SecurityService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/SecurityService.java
@@ -41,6 +41,7 @@ import javax.servlet.http.HttpServletRequest;
 
 import java.io.File;
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


jvoisin:
Nice catch :)

eharris:
Maybe not that great a catch after all.  I just realized that it was probably originally this way because the `SubsonicRESTController` also uses this method, and the default defined in the API spec is to grant all folders unless specific folders are provided.

So I guess this is another case of deciding if we want to break API compatibility or leave things as they are.

muff1nman:
Hm yeah I dont think this is worth breaking API over.

eharris:
I'm closing this for now, since it results in a breaking change to the REST API.

PR #1230, opened 2019-09-11
Admin users now always get access to all media folders

eharris:
It seems to be somewhat common that users will add new media folders and then
wonder why they can't see the contents, because they haven't enabled their
user to have access to the new folder(s).  This tries to improve that
situation by automatically granting all admin users access to all media
folders.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/SettingsService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/SettingsService.java
@@ -241,6 +241,8 @@ public class SettingsService {
     private AvatarDao avatarDao;
     @Autowired
     private ApacheCommonsConfigurationService configurationService;
+    @Autowired
+    private SecurityService securityService;
 
     private String[] cachedCoverArtFileTypesArray;
     private String[] cachedMusicFileTypesArray;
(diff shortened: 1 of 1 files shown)


jooola:
I'm not sure this is a good idea, admin user might not want to listen all of the media folders, it is my case anyway.

eharris:
Well that's a perfect reason why you shouldn't be using an admin account for normal listening.  :-)

This change does use the Admin Role to determine if the user is an admin.  There is still a separate PR to completely remove the hardcoded admin account (#1131), but this doesn't depend on that, so I made it a separate PR.

tesshucom:
I don't think anyone will refuse simple enhancements, obviously unnecessary feature removal, simple updates, or bug fixes.
If you lose a legacy feature working without problems and replace it with another feature, you might be more careful.

Even if you think it's good, others are wary.
In the worst case, it can be confusing and discouraging for the user, which is counterproductive.
Wasn't such a feature a policy to consider in advance in Issues?

And the definition of ``usability`` is greatly expanded in this project.
I want proper labeling.
If classification is appropriate, it is easy to eliminate with clones.

Or you can do it freely with your clone.

tesshucom:
I don't think anyone will refuse simple enhancements, obviously unnecessary feature removal, simple updates, or bug fixes.
If you lose a legacy feature working without problems and replace it with another feature, you might be more careful.
In the worst case, it can be confusing and discouraging for the user, which is counterproductive.
Wasn't such a feature a policy to consider in advance in Issues?

And the definition of ``usability`` is greatly expanded in this project.
If classification is appropriate, it is easy to eliminate with clones.
Or you can do it freely with your clone.

The solution against the problem seems somewhat intentional and extreme.

muff1nman:
I'm not sure that I'm fond of this either - seems to be masking the underlying issue (that new folders don't have admin rights setup when you add them)

eharris:
To me it seems logical that admin users should always be able to see everything.  But if the takeaway of the comments so far is that the preferred method of accomplishing this is instead to automatically add permissions for any new media folders for all admin users, then it would seem that this PR now on hold pending #1131 being merged.

PR #1232, opened 2019-09-11
Media file scan will now heed configured music/video file types

### LINKED ISSUE #1227: Music file types (extensions?) list is not honored?
## Problem description

I would like to exclude all lossless formats from getting indexed (added to the database) at all.
I'm trying to achieve this logged in as admin, Settings => Music Files is currently filled with  
`mp3 ogg oga aac m4a wma ape mpc shn mka opus` which I thought to mean that only files with this extension are considered (case insensitive).

### Steps to reproduce

With the settings as described, I do both "Scan Media Folders" and "Clean-up database". Then I open my Android app DSub, swipe down to get the latest database from the server.
Whatever I do, FLAC files are still being shown, and are playable too.

## System information

 * **Airsonic version**:
Still on 10.3.1-RELEASE – May 21, 2019 :blush: but this problem has been with me for a long time.

 * **Operating system**:
Debian oldstable (stretch) no-GUI server install.

 * **Java version**: 
Apache Tomcat/8.5.40, java 1.8.0_222

 * **Proxy server**:
nginx

 * **Client**:
Admin in Firefox, actual listening on Android client (DSub)

eharris:
It will also now heed the Media Folders -> Exclude Pattern.  Fixes #1227.

This will need some testing, as there could be unintended side effects.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/MediaFileService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/MediaFileService.java
@@ -193,10 +193,10 @@ public class MediaFileService {
         List<MediaFile> result = new ArrayList<MediaFile>();
         for (MediaFile child : mediaFileDao.getChildrenOf(parent.getPath())) {
             child = checkLastModified(child, useFastCache);
-            if (child.isDirectory() && includeDirectories) {
+            if (child.isDirectory() && includeDirectories && includeMediaFile(child)) {
                 result.add(child);
             }
-            if (child.isFile() && includeFiles) {
+            if (child.isFile() && includeFiles && includeMediaFile(child)) {
                 result.add(child);
             }
         }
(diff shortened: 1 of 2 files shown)


airsonic-main/src/main/java/org/airsonic/player/service/MediaFileService.java
```diff
@@ -412,11 +412,22 @@ private void updateChildren(MediaFile parent) {
         mediaFileDao.createOrUpdateMediaFile(parent);
     }
 
+    public boolean includeMediaFile(MediaFile candidate) {
+        return includeMediaFile(candidate.getFile());
+    }
+
+    public boolean includeMediaFile(File candidate) {
+        String suffix = FilenameUtils.getExtension(candidate.getName()).toLowerCase();
+        if (!isExcluded(candidate) && (FileUtil.isDirectory(candidate) || isAudioFile(suffix) || isVideoFile(suffix))) {
```
jvoisin: Nit: Splitting the condition into two if improves a bit the readability. Feel free to ignore if you disagree :)

```suggestion
        if (!isExcluded(candidate) {
           if (FileUtil.isDirectory(candidate) || isAudioFile(suffix) || isVideoFile(suffix)) {
```

eharris: I just moved those two lines from the other function without changing them. I agree splitting would improve readability, I just usually tend to separate functional vs cleanup changes (to minimize changes and simplify reviewing).

PR #1235, opened 2019-09-15
Improve search accuracy

tesshucom:
Related to #1142.
This PR proposes the following two improvements.

 - Setting boost value. Legacy code was conscious of introduction, but it was dead code.
 - Airsonic-specific Stopward for music search.

In a search system, these are usually considered at design time.
The current search does n’t take into account the convenience of users who type short phrases.

Here is a simple example.
___

### Stopward change example

``Will`` cannot be used in legacy searches.

image

image

___

## Example of boost value adjustment

It is reasonable to give priority to the leftmost item in the search results.
Since the boost value is assigned a very small value, the priority will be reversed if the cost of the Artist name is high.

### before

image

### after

image

___

Brainstorm is required for the value set in Stopward.
(Particularly the opinions of English speaking people are necessary. Because I am Japanese.)

 - Once this is done, maintenance is easy and anyone can do it. For example, if you add a composer, you can give it a new priority.
 - Stopwords are assigned differently for Artist and other fields. This is because the words you want to ignore, such as `` feat``,  `` with``, are a little different.

@muff1nman 
 - Although not required, there are some considerations. This changes affect the index version. So if you release this PR at the same time as the Lucene update, users will be satisfied (If the release is divided, the index must be reconfigured.). This PR does not include increments.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/search/AnalyzerFactory.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/search/AnalyzerFactory.java
@@ -51,6 +51,10 @@ import static org.springframework.util.ObjectUtils.isEmpty;
 @Component
 public final class AnalyzerFactory {
 
+    private static final String STOP_WORDS = "org/airsonic/player/service/search/analysis/stopwords.txt";
+
+    private static final String STOP_WORDS_ARTIST = "org/airsonic/player/service/search/analysis/stopwords_artist.txt";
+
     private Analyzer analyzer;
 
     private Analyzer queryAnalyzer;
@@ -105,7 +109,19 @@ public final class AnalyzerFactory {
                 .addTokenFilter(CJKWidthFilterFactory.class)
                 .addTokenFilter(ASCIIFoldingFilterFactory.class, "preserveOriginal", "false")
                 .addTokenFilter(LowerCaseFilterFactory.class)
-                .addTokenFilter(StopFilterFactory.class)
+                .addTokenFilter(StopFilterFactory.class, "words", STOP_WORDS)
+                .addTokenFilter(EnglishPossessiveFilterFactory.class);
+        addTokenFilterForUnderscoreRemovalAroundToken(builder);
+        return builder;
+    }
+
+    private Builder createArtistAnalyzerBuilder() throws IOException {
+        Builder builder = CustomAnalyzer.builder()
+                .withTokenizer(StandardTokenizerFactory.class)
+                .addTokenFilter(CJKWidthFilterFactory.class)
+                .addTokenFilter(ASCIIFoldingFilterFactory.class, "preserveOriginal", "false")
+                .addTokenFilter(LowerCaseFilterFactory.class)
+                .addTokenFilter(StopFilterFactory.class, "words", STOP_WORDS_ARTIST)
                 .addTokenFilter(EnglishPossessiveFilterFactory.class);
         addTokenFilterForUnderscoreRemovalAroundToken(builder);
         return builder;
@@ -134,7 +150,15 @@ public final class AnalyzerFactory {
     public Analyzer getAnalyzer() throws IOException {
         if (isEmpty(analyzer)) {
             try {
-                analyzer = createDefaultAnalyzerBuilder().build();
+
+                Analyzer defaultAnalyzer = createDefaultAnalyzerBuilder().build();
+                Analyzer artistAnalyzer = createArtistAnalyzerBuilder().build();
+
+                Map<String, Analyzer> fieldAnalyzers = new HashMap<>();
+                fieldAnalyzers.put(FieldNames.ARTIST, artistAnalyzer);
+
+                analyzer = new PerFieldAnalyzerWrapper(defaultAnalyzer, fieldAnalyzers);
+
             } catch (IOException e) {
                 throw new IOException("Error when initializing Analyzer.", e);
             }
@@ -161,9 +185,11 @@ public final class AnalyzerFactory {
             try {
 
                 Analyzer defaultAnalyzer = createDefaultAnalyzerBuilder().build();
+                Analyzer artistAnalyzer = createArtistAnalyzerBuilder().build();
                 Analyzer genreAnalyzer = createGenreAnalyzerBuilder().build();
 
                 Map<String, Analyzer> fieldAnalyzers = new HashMap<>();
+                fieldAnalyzers.put(FieldNames.ARTIST, artistAnalyzer);
                 fieldAnalyzers.put(FieldNames.GENRE, genreAnalyzer);
 
                 queryAnalyzer = new PerFieldAnalyzerWrapper(defaultAnalyzer, fieldAnalyzers);
--- a/airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
@@ -74,7 +74,7 @@ public class IndexManager {
      *    DocumentFactory or the class that they use.
      *
      */
-    private static final int INDEX_VERSION = 16;
+    private static final int INDEX_VERSION = 17;
 
     /**
      * Literal name of index top directory.
--- a/airsonic-main/src/main/java/org/airsonic/player/service/search/IndexType.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/search/IndexType.java
@@ -39,7 +39,7 @@ public enum IndexType {
             FieldNames.TITLE,
             FieldNames.ARTIST),
         boosts(
-            entry(FieldNames.TITLE, 2F))),
+            entry(FieldNames.TITLE, 1.1F))),
 
     ALBUM(
         fieldNames(
@@ -47,7 +47,7 @@ public enum IndexType {
             FieldNames.ARTIST), 
             // FieldNames.FOLDER), // XXX 3.x -> 8.x : Remove folder from multi-field search condition
         boosts(
-            entry(FieldNames.ALBUM, 2F))),
+            entry(FieldNames.ALBUM, 1.1F))),
 
     ALBUM_ID3(
         fieldNames(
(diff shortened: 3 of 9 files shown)


fxthomas:
A few questions about how this works:

* Is it possible to not ignore stopwords but leave them at a low priority? This way one can still search for a stop word if we really want it. How do you think this would interfere with the accuracy?
* How consistent is your PR with the "articles to ignore" field in Settings > General that specifically mentions that these articles will be ignored in the index? Is that setting still relevant?
* Edge case related to the setting : depending on the language, some stop words might be different. For example [los](https://musicbrainz.org/search?query=los&type=release&method=indexed) in German means either "go!" or "without" and should probably not ignored. That's the reason why I like the idea of keeping it configurable.

Otherwise I like what your screenshots show, it sounds reasonable to boost the priority of the artist field in the Artist column. Great job!

tesshucom:
This PR shows very effective steps that are technically easy to understand.
It is also a very basic technique for search engine and should be tackled first.
It is desirable to adjust as necessary.

> Is it possible to not ignore stopwords but leave them at a low priority?

Stopwords and Boost commits are separated.
These are separate mechanisms and can be adjusted individually.
It is possible not to set Stopward at all or you can remove the filter with Analyzer.

I think you can create logic that lowers the boost for a specific word, but I have never seen it.

> How do you think this would interfere with the accuracy?

This just depends on the data used.
If Stopwords is not set at all, when a song with a relatively long title is input, a considerable number of erroneous searches increase.
If there are more Stopwords than keywords that are really important, it will be detrimental to the cost calculation.
As a result It becomes that becomes difficult to display in order that user wanted.

In that sense, stopwords and Boost are related topics.

> How consistent is your PR with the "articles to ignore" field in Settings > General that specifically mentions that these articles will be ignored in the index? Is that setting still relevant?

Since the contents are defined below, I think it is a good idea to consider this.
(You can easily rewrite and test the effect)

stopwords.txt 
stopwords_artist.txt


Although dynamically change is technically possible, it may not be common to make Stopword changes externally during operation.
Because operation becomes complicated (Full scan is required, etc.).


> depending on the language, some stop words might be different.

I understand.

This is a personal idea, but language-specific localization is difficult to support and should be done individually by the people in that country.
Usually stopwords.txt is prepared for each country.
(If it is general, it can be found by google search)

I think that it is possible to give an idea to lower the tuning burden step by step.

 - First, we will commit to the scope of Standard(English). As a basic mechanism, the contents of this PR are reflected in the master. The first step should be as simple as possible, it will be easier for people to see later.
 - For example, if you have a dedicated directory like trancecode or db and there is an additional stopwords.txt, load it. Consider such a mechanism. I think this can be done without me.

tesshucom:
This PR shows very effective steps that are technically easy to understand.
It is also a very basic technique for search engine and should be tackled first.
It is desirable to adjust as necessary.

> Is it possible to not ignore stopwords but leave them at a low priority?

Stopwords and Boost commits are separated.
These are separate mechanisms and can be adjusted individually.
It is possible not to set Stopward at all or you can remove the filter with Analyzer.

I think you can create logic that lowers the boost for a specific word, but I have never seen it.

> How do you think this would interfere with the accuracy?

This just depends on the data used.
If Stopwords is not set at all, when a song with a relatively long title is input, a considerable number of erroneous searches increase.
If there are more Stopwords than keywords that are really important, it will be detrimental to the cost calculation.
As a result It becomes that becomes difficult to display in order that user wanted.

In that sense, stopwords and Boost are related topics.

> How consistent is your PR with the "articles to ignore" field in Settings > General that specifically mentions that these articles will be ignored in the index? Is that setting still relevant?

Since the contents are defined below, I think it is a good idea to consider this.
(You can easily rewrite and test the effect)

stopwords.txt 
stopwords_artist.txt


Although dynamically change is technically possible, it may not be common to make Stopword changes externally during operation.
Because operation becomes complicated (Full scan is required, etc.).


> depending on the language, some stop words might be different.

I understand.

This is a personal idea, but language-specific localization is difficult to support and should be done individually by the people in that country.
Usually stopwords.txt is prepared for each country.
(If it is general, it can be found by google search)

I think that it is possible to give an idea to lower the tuning burden step by step.

 - First, we will commit to the scope of Standard(English). As a basic mechanism, the contents of this PR are reflected in the master. The first step should be as simple as possible, it will be easier for people to see later.
 - For example, if you have a dedicated directory like trancecode or db and there is an additional stopwords.txt, load it. Consider such a mechanism. I think this can be done without me. All current test codes are written assuming English.

tesshucom:
Examples specializing in the native language :

In Japan, it is possible to handle most if English and Japanese are supported.
So create a mixed file.

https://github.com/jpsonic/jpsonic/blob/master/jpsonic-main/src/main/resources/com/tesshu/jpsonic/service/stopwords.txt

I'm worried that if it's configurable from the screen, it's convenient for some people, but it can annoy beginners.
So I think it's a good idea to consider adding expert options as another additional feature addition.
If you can create a representative example for each country, you can also link from an Airsonic document.

tesshucom:
Examples specializing in the native language :

In Japan, it is possible to handle most if English and Japanese are supported.
So create a mixed file.

https://github.com/jpsonic/jpsonic/blob/master/jpsonic-main/src/main/resources/com/tesshu/jpsonic/service/stopwords.txt

I'm worried that if it's configurable from the screen, it's convenient for some people, but it can annoy beginners.

So If necessary I think it's a good idea to consider adding expert options as another additional feature addition.
If you can create a representative example for each country, you can also link from an Airsonic document.

There is no good example of introducing such customization in music search.
So I think everyone needs to think about good ideas together.

tesshucom:
#### Candidates to remove from default Stopwards

los  -> to be delete because German is not good

la 　-> can't find an artist called la la or a song called la la la

 #### Candidates to add

None for now

tesshucom:
> * Is it possible to not ignore stopwords but leave them at a low priority? This way one can still search for a stop word if we really want it. How do you think this would interfere with the accuracy?

We can create logic that lowers the boost for specific words.
That is technically possible.

I have never seen it.
Probably not common.
This is to complicate cost calculations.

Even if the boost value is lowered...
If it hits , false search increases when a song with a long title is input.
If there are more noisy words than important keywords, this has a negative impact on cost.

As a result, it will be difficult to display in the order the user wants.

> * How consistent is your PR with the "articles to ignore" field in Settings > General that specifically mentions that these articles will be ignored in the index? Is that setting still relevant?

It is statically defined in the following file:

stopwords.txt 
stopwords_artist.txt

It can be changed dynamically, but the following points need to be considered.

 - Complex operation (full scan required)
 - Stopwords is part of the Analyzer mechanism. Especially strongly dependent on Tokenizer.
    (In other words, multilingual support is not possible completely only with this setting value.)

Generally, it is handled by the file unit of stopwords.txt as is done in this PR.


> * Edge case related to the setting : depending on the language, some stop words might be different. For example [los](https://musicbrainz.org/search?query=los&type=release&method=indexed) in German means either "go!" or "without" and should probably not ignored if your collection consists mostly of German artists/albums. That's the reason why I like the idea of keeping it configurable.

i'll put in.

_____

Happy if you can accumulate feedback as follows:


 #### Candidates to remove from default Stopwards

los  -> to be delete because German is not good.

la 　-> can't find an artist called la la or a song called la la la.

 #### Candidates to add

None for now

 #### I tried to add but didn't

x 　-> X means collaboration. I can't find an artist named X.

fxthomas:
> Complex operation (full scan required)

That's correct, although we could automatically start a rescan if that happens.

> Stopwords is part of the Analyzer mechanism. Especially strongly dependent on Tokenizer (In other words, multilingual support is not possible completely only with this setting value.)

You're right on this. If we want to implement this we should probably add a Tokenizer setting in addition to the stop words, but not in this PR.

I will pull & test, will let you know if everything looks good for me.

tesshucom:
The currently "articles to ignore" is a setting item corresponding to index of domain.
(When creating the index on the left side of the screen, to remove the first article and properly classify it)

"articles to ignore" is included in the domain design and is also an output item of the REST API.
So this may not be deleted.

Search Stopward is not directly related to domain design.
Legacy implementations use defaults of Lucene and are not visible on Airsonic code.
The previous Stopward is defined in EnglishAnalyzer.ENGLISH_STOP_WORDS_SET.

fxthomas:
Ah, I missed that it was never affecting Lucene at all (the "index" the field was mentioning was the left sidebar index, not the Lucene index). I don't have an issue with the changes if that is the case.

I found no issues so far on my server (was expecting none, but still it's nice to know).

tesshucom:
Thank you very much.
Removed los and la from Stopward.

Language switching can be achieved by replacing Analyzer and Stopword if the language is similar to English.
I think fxthomas can probably do that.
A little annoying, but Legacy is moving forward because it couldn't.

Implementing a simple switch may be possible but impractical
because additional analyzers and data are required.

I saw in the previous issues that "plugins have little merit to hard work".
For this reason, there are currently no plans for language switching as a successor to this PR.

(Depending on the language, additional logic and dictionaries can increase the war size by 30Mb.)

airsonic-main/src/main/java/org/airsonic/player/service/search/AnalyzerFactory.java
```diff
@@ -99,13 +103,13 @@ private void addTokenFilterForTokenToDomainValue(Builder builder) throws IOExcep
                     "pattern", "\\[\\]", "replacement", "\\[ \\]", "replace", "all");
     }
 
-    private Builder createDefaultAnalyzerBuilder() throws IOException {
+    private Builder createDefaultAnalyzerBuilder(boolean isArtist) throws IOException {
```
muff1nman: It seems better to leave `createDefaultAnalyzerBuilder` alone and create a new `createArtistAnalyzerBuilder` method to be consistent with how you've handled the genere scenario.

tesshucom: I revised that according to your advice.

tesshucom:
For example, if you have an additional Stopward file in a specific directory, you might find it useful to load it.
(There is also a method to make it unofficial function for a while without releasing it to the user.)

Adding a setting to the screen may be a little strange.
 -  Beneficial to some core users while beginners get confused.
 -  Generally handled in files

There are few in English alone, but it is huge depending on the language.
Multilinguals tend to be huge.
Assuming that the content is complex, it will be easier to manage the files.

In any case, I agree with the idea that Stopward can be changed in some way.
It is very useful to have a way to test what Stopward is suitable for.

tesshucom:
You can combine two commits.
I will leave it to you.

In terms of content, I think this is a slightly different modification.
The boost value is small this time, but you can increase it a little more.
This value is adjusted according to the purpose.

fxthomas:
I do believe that this one is ready to merge as soon as conflicts are resolved?

tesshucom:
Plese note only the following points:

>  This changes affect the index version. If the (Update of Lucene)release is divided, the index must be reconfigured.

At the time this PR was created, it was impossible to predict when the release this PR would take place.

fxthomas:
Good point. If we release both Lucene upgrades and this PR at the same time, things should be fine without additional upgrades, am I right?

muff1nman:
Ah shucks this should have gotten pulled into the previous release. I'm sorry I forgot about this one. Well this will need to bump the index version now.

fxthomas:
We just need to update `INDEX_VERSION`, right? If you are okay I think I can merge this one :wink:

tesshucom:
After rebase, INDEX_VERSION update merged with Stopword update.

fxthomas:
Thank you so much for the help, merging now!

Issue #339, opened 2017-04-06
postgresql  ERROR while scanning media folder


issuemover631:
<a href="https://github.com/cascooscuro"><img src="https://avatars2.githubusercontent.com/u/5746813?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [cascooscuro](https://github.com/cascooscuro)**
_Thursday Apr 06, 2017 at 10:57 GMT_
_Originally opened as https://github.com/Libresonic/libresonic/issues/337_

----

Hi,

wile scanning media folder, Libresonic throws an error and stops the scan:

[2017-04-06 12:20:12,027] INFO MediaScannerService - Scanned media library with 10750 entries.
[2017-04-06 12:20:13,776] ERROR MediaScannerService - Failed to scan media library.
org.springframework.dao.DataIntegrityViolationException: PreparedStatementCallback; SQL [update media_file set folder=?,type=?,format=?,title=?,album=?,artist=?,album_artist=?,disc_number=?,track_number=?,year=?,genre=?,bit_rate=?,variable_bit_rate=?,duration_seconds=?,file_size=?,width=?,height=?,cover_art_path=?,parent_path=?,play_count=?,last_played=?,comment=?,changed=?,last_scanned=?,children_last_updated=?,present=?, version=? where path=?]; ERROR: invalid byte sequence for encoding "UTF8": 0x00; nested exception is org.postgresql.util.PSQLException: ERROR: invalid byte sequence for encoding "UTF8": 0x00
        at org.springframework.jdbc.support.SQLStateSQLExceptionTranslator.doTranslate(SQLStateSQLExceptionTranslator.java:102)
        at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:73)
        at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:81)
        at org.springframework.jdbc.support.AbstractFallbackSQLExceptionTranslator.translate(AbstractFallbackSQLExceptionTranslator.java:81)
        at org.springframework.jdbc.core.JdbcTemplate.execute(JdbcTemplate.java:649)
        at org.springframework.jdbc.core.JdbcTemplate.update(JdbcTemplate.java:870)
        at org.springframework.jdbc.core.JdbcTemplate.update(JdbcTemplate.java:931)
[...]

PR #1236, opened 2019-09-15
filter out \0 characters from metadata, fixes 'ERROR: invalid byte se…

Yetangitu:
filter out \0 characters from metadata, fixes 'ERROR: invalid byte se…quence for encoding "UTF8": 0x00;' (#339)

Postgresql uses C-strings to represent string types in the database and as such these can not contain `\0` characters. While _Airsonic_ itself does not produce strings containing such data, external sources (e.g. metadata from files or off the 'net) can and sometimes do contain such 'garbage' which leads to problems for those who use Postgresql and nonsensical stored data for those who do not.

This patch makes sure these `\0` characters are filtered out before they can cause problems.


Yetangitu:
...another problem with Travis leading to false alarms about failed checks. Time to find a CI-system which works, Travis has been very spotty lately...

tesshucom:
I was going to do a survey related to this.
I haven't examined fully the JAudioparser specification yet. ( ``\0`` is used in id3v2.4)
JDBC or data layer mapper support may be good... but can't answer right away:'(

Yetangitu:
> I was going to do a survey related to this.
> I haven't examined fully the JAudioparser specification yet. ( `\0` is used in id3v2.4)
> JDBC or data layer mapper support may be good... but can't answer right away:'(

There is a difference between `null` (i.e. the absence of input) and `\0` characters in string data. The former fulfills a need, the latter is garbage in normal strings.

Jaudiotagger is not the only metadata provider, there is also `ffmpeg` for those file formats which are not supported by _Jaudiotagger_. Better to fix this there where all metadata providers meet, in `MetaData`.

tesshucom:
OK.
To tell the truth, I found a bad code in parser with my fix earlier.
(Code that does not malfunction but is unnecessary.)
I'm preparing test data to fix this now, so i'll check together with id3v2.4 delimiter.

Yetangitu:
Random malfunctions? Sounds... odd. Do keep in mind that the intention of this change is not so much to get the metadata parser to reproduce the contents of tags 1 on 1 but to make sure that the data returned does not trigger the `\0` character problem in Postgresql. Better make sure the test represents that fact... _be liberal in what data you accept, conservative in what data you produce_ ...

eharris:
As far as nul (`\0`) characters in metadata, id3v2.3 does not "support" multi-value fields (this was only added in id3v2.4 AFAIK), but v2.3 also seems to be the most commonly used (and best supported) tagging mechanism.

However, while it is not officially allowed, I can tell you that I have somewhat often seen files with metadata that uses `\0` as multi-value separators as a way to shoehorn some v2.4 functionality into v2.3 (and also other file formats that are not mp3).  Since that is the case, depending on the field, rather than simply stripping `\0` characters, there are almost certainly instances where they should instead be replaced with something else that still acts as a separator.  The most common tags I have seen this in are `GENRE`-type fields.  

Personally I prefer semicolon since that virtually never occurs "normally" in artist or song names or other metadata except for very verbose descriptions.  Maybe newline could also work?

Yetangitu:
Sure, it can be replaced with anything other than an empty space, however the old dictum _garbage in, garbage out_ still holds.. Where I encountered `\0` (indeed, in the `GENRE` field) it contained nothing but `Text="\u0000";`. This would not be made much better or clearer by replacing the `\0` with a colon...

In other words, this problem can not really be solved entirely satisfactorily as it is unknown why those `\0` characters are there in the first place. Maybe it makes sense to replace them with a separator for strings which contain more than just that `\0` character. In my case that would still give garbage but for the cases you mentioned it would be reasonable.

I'll update the patch.

tesshucom:
MusicBee is an example of a software that has an unfavorable delimiter specification.
This case may be deliberately generated.

Semi-colon delimiters are common, but ID3v2.2, so strictly speaking, there are cases outside the specification.
After all, it needs to be able to handle either \0, semicolon or slash.
However, since it is a sensitive problem, it must be backed up.

Resolution by string manipulation and conversion is always subtle.
I greatly agree with the necessary case, but it is directly linked to overall performance degradation.

Yetangitu:
BTW & OT: this is why Travis bøtched the last test run:

```
[WARNING] unable to get access token for Google Container Registry, configuration for building image will not contain RegistryAuth for GCR

java.io.IOException: Error code 404 trying to get security access token from Compute Engine metadata for the default service account. This may be because the virtual machine instance does not have permission scopes specified.

    at com.google.auth.oauth2.ComputeEngineCredentials.refreshAccessToken (ComputeEngineCredentials.java:111)

    at com.google.auth.oauth2.OAuth2Credentials.refresh (OAuth2Credentials.java:149)

    at com.spotify.docker.client.auth.gcr.ContainerRegistryAuthSupplier$DefaultCredentialRefresher.refresh (ContainerRegistryAuthSupplier.java:195)

    at com.spotify.docker.client.auth.gcr.ContainerRegistryAuthSupplier.getAccessToken (ContainerRegistryAuthSupplier.java:225)

    at com.spotify.docker.client.auth.gcr.ContainerRegistryAuthSupplier.authForBuild (ContainerRegistryAuthSupplier.java:294)

    at com.spotify.docker.client.auth.MultiRegistryAuthSupplier.authForBuild (MultiRegistryAuthSupplier.java:77)

    at com.spotify.docker.client.DefaultDockerClient.build (DefaultDockerClient.java:1468)

    at com.spotify.docker.client.DefaultDockerClient.build (DefaultDockerClient.java:1445)

    at com.spotify.plugin.dockerfile.BuildMojo.buildImage (BuildMojo.java:201)

    at com.spotify.plugin.dockerfile.BuildMojo.execute (BuildMojo.java:110)

    at com.spotify.plugin.dockerfile.AbstractDockerMojo.tryExecute (AbstractDockerMojo.java:259)

    at com.spotify.plugin.dockerfile.AbstractDockerMojo.execute (AbstractDockerMojo.java:248)

    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:137)

    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:210)

    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:156)

    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:148)

    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)

    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:81)

    at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build (SingleThreadedBuilder.java:56)

    at org.apache.maven.lifecycle.internal.LifecycleStarter.execute (LifecycleStarter.java:128)

    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:305)

    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:192)

    at org.apache.maven.DefaultMaven.execute (DefaultMaven.java:105)

    at org.apache.maven.cli.MavenCli.execute (MavenCli.java:956)

    at org.apache.maven.cli.MavenCli.doMain (MavenCli.java:288)

    at org.apache.maven.cli.MavenCli.main (MavenCli.java:192)

    at sun.reflect.NativeMethodAccessorImpl.invoke0 (Native Method)

    at sun.reflect.NativeMethodAccessorImpl.invoke (NativeMethodAccessorImpl.java:62)

    at sun.reflect.DelegatingMethodAccessorImpl.invoke (DelegatingMethodAccessorImpl.java:43)

    at java.lang.reflect.Method.invoke (Method.java:498)

    at org.codehaus.plexus.classworlds.launcher.Launcher.launchEnhanced (Launcher.java:289)

    at org.codehaus.plexus.classworlds.launcher.Launcher.launch (Launcher.java:229)

    at org.codehaus.plexus.classworlds.launcher.Launcher.mainWithExitCode (Launcher.java:415)

    at org.codehaus.plexus.classworlds.launcher.Launcher.main (Launcher.java:356)
```

eharris:
I created an issue for the Cucumber/build issue, #1240

airsonic-main/src/main/java/org/airsonic/player/service/metadata/MetaData.java
```diff
@@ -152,12 +152,11 @@ public void setMusicBrainzReleaseId(String musicBrainzReleaseId) {
...
 
-    // filter out any \0 characters
     private String sanitize(String meta) {
-        if (meta != null) {
-            return meta.replace("\u0000", "");
-        } else {
+        if (meta == null || meta.replace("\u0000", "").length() == 0) {
             return null;
+        } else {
+            return meta.replace("\u0000", ":");
```
eharris: Colon is pretty commonly used in conventional titles, especially for classical music.  I would strongly suggest using semi-colon as that is more effective as a multi-value separator.

eharris:
Since this is a Postgres specific problem, I wonder if it wouldn't be better to find a way to do the filtering only for Postgres and in the database interface layer?

Yetangitu:
> Since this is a Postgres specific problem, I wonder if it wouldn't be better to find a way to do the filtering only for Postgres and in the database interface layer?

I don't thinks so given that `\0` characters don't serve any real purpose, can not be rendered nor edited in the tag editor and as such _should_ be filtered out or changed into something which can be rendered and edited. Moving this to the database layer would only complicate things for no reason and to no gain.

airsonic-main/src/main/java/org/airsonic/player/service/metadata/MetaData.java
Yetangitu: Or a dash, or a double colon, or semicolon+space,... make it configurable in the _advanced_ section. 

A quick survey over a set of 200.000 files (ranging from music to lectures) gives the following:

character(s) | count
-----|----
:|3457
/|1716
;|369
+|79
\\ |4
::|0

Double-colon to the rescue

Yetangitu:
`:` -> `::` as replacement character for `\0`

eharris:
Except that double colon isn't used by any software that I know of as a delimiter.  Part of the point (IMO) is to use something that other software will still recognize as a delimiter in case the metadata is saved back to the files (or used via the API).  Luckily Picard/Musicbrainz makes that configurable, but lots of software doesn't, and my experience the most commonly used delimiter for multi value fields that is ALSO minimally used as parts of real names or titles is semicolon.  Colon (:) and comma (,) are very commonly used, especially in titles that contain an expansion (e.g. "Waltz: Learn to Dance") or in things like classical music ("Movement I: Andante"), bible verses ("John 3:16") or law subjects, (+) is commonly used as another version of ampersand (&), slashes are not uncommonly used in names (e.g. AC/DC and If/Then (a broadway show)).

eharris:
Also it should be noted that software that does a string split on colon will end up with an empty field between the double colon, which is often thrown away when saving info back to the files, so the double colon would be removed and we are back where we started.

tesshucom:
(1) Not a postgres limitation (at least in the documentation).
(2) pgjdbc has raised issues in the past by users and expressed refusal to respond because of performance concerns.

This is a problem specific to pgjdbc.

For a program that has a parser that handles Id3, it is assumed that the input contains ``\0``.
It is doubtful whether unnatural metadata conversion is an official application requirement for Airsonic.
(if there was no problem with pgjdbc, there should have been no problem.)

Since the conditions are clear, users can avoid problems without changing the program.
At the moment, as ehrris says, it is common for id3 users to consider the versatility of the data.

It sounds unkind, but it's not a good idea for Airsonic to have problems that pgjdbc developers are concerned about.
Even if we take measures, can be careful.

It is not pgjdbc or us that has committed the deepest sins.
id3 specification :-@

Yetangitu:
Well, as said, there are many possibilities since there is no 'standard'. As to whether there is other software out there which uses double-colon as a separator, maybe, who can tell. There is no 'good' choice here since the most likely candidates - semicolon, colon, comma, ampersand (&) - are in relatively common use in filenames. I say make it configurable and use semicolon as a default, those who want to change it are free to do so. Alternatively just replace `\0` with a space character or a newline (which won't get rendered in _airsonic_). Leaving those `\0` characters in place is not really an option given that they shouldn't have been there in the first place. 

Postgresql does have a limitation on `\0` characters in string types (`varchar/character varying`) which is not likely to be solved '_because of the widespread use of C strings (nul-terminated strings) inside the backend.  Making \0 a legal member of strings would break all those internal APIs, requiring touching far more code than anyone would want to do.  It'd likely break a great deal of client-side code as well._ (https://www.postgresql.org/message-id/8239.1500489054%40sss.pgh.pa.us)

The next version of the patch will include a configurable separator, default set to `;` (semicolon). The configuration option will be on the `database` page, under the `User table Quote` parameter.

tesshucom:
I understand the technical background.
But the location where the conversion is processed is not appropriate by design.
Deadweight for non-Postgres users.
(Don't get me wrong. I love Postgres.)

Appropriate method:

#### I. Add setting items to the screen

image

Of course, it is clearly stated in Help that is is Postgres-JDBC-aware features and that it will cause deterioration in scan performance.


#### II. Creating a compatible parser

(1) Create an interface to the meta parser
(2) Create a wrapper for an existing parser and pass the conversion method for the required method.
    Or inherit existing class and override.
(3) Add UT that can clearly show the difference between existing class and pgsql class.

Certainly troublesome.
However, this proof method was effective even for large-scale changes such as SearchService.

#### III. Switching between parser instances

Get the flag from SettingService and switch the parse.
As a result, if the function is Off, the existing processing speed is not affected.

____


Such a process would convince  other than Postgres users and engineers and Postgres users.
Furthermore, if a similar request occurs in the future, it can be used as an example of a coping method that is effective as a use case in which the user can arbitrarily change the parser.

tesshucom:
The conversion method should be considered in the design.
Deadweight for non-Postgres users.
(Don't get me wrong. I love Postgres.)

Examples of reducing risk:

#### I. Add setting items to the screen

image

Of course, it is clearly stated in Help that is is Postgres-JDBC-aware features and that it will cause deterioration in scan performance.


#### II. Creating a compatible parser

(1) Create an interface to the meta parser
(2) Create a wrapper for an existing parser and pass the conversion method for the required method.
    Or inherit existing class and override.
(3) Add UT that can clearly show the difference between existing class and pgsql class.

Certainly troublesome.
However, an effective proof method.

#### III. Switching between parser instances

Get the flag from SettingService and switch the parse.
As a result, if the function is Off, the existing processing speed is not affected.

____


Such a process would convince  other than Postgres users and engineers and Postgres users.
This is because it is not a change of existing processing, but some implementations are paralleled.

Furthermore, if a similar request occurs in the future, it can be used as an example of a coping method that is effective as a use case in which the user can arbitrarily change the parser.

For example, #1132 can be solved with this.
If the original process is preserved and the user understands the meaning of behavior switching, the specification is not destroyed.

____

https://en.wikipedia.org/wiki/Strategy_pattern


However, it's not a universal magic.
Keep in mind that they require a forced scan to ensure data integrity.

If it is a processing change, not a Strategy, requires an incrementing data version.
Even in the case of Strategy, scanning is necessary, so it is an operation method for experts.

Basically, adding processing to DTO is a risky act.
When doing this in a live running system, appropriate attention and measures are required.

Yetangitu:
I prefer KISS over _pattern soup_ to be honest. Also, _null_ characters in textual data do not make sense [1] no matter whether the database supports them or not given that they can not be rendered nor edited. They end up in there anyway for one reason or another, from 'creative' attempts to get around technical limitations of the `ID3` specification through erroneous conversion of empty `C` strings to UTF8 (which, as noted, _does_ accept `\0` as a valid character) to [_Jaudiotagger_ including them](https://bitbucket.org/ijabz/jaudiotagger/commits/c808c1b4cfed7bd3d19e9323da190ff652c78856) when maybe it shouldn't [2]. Browsing the _Jaudiotagger_ source reveals that _iTunes_ may be another culprit here [3].

Given these facts and the fact that the performance aspect of a simple (non-regex) `String.replace()` is negligible I think it is best to just get rid of those `\0` characters, either by replacing or just simply removing them. Everyone will benefit, both by having cleaner data as well as (for those who use Postgresql) by being able to use _airsonic_ in the first place.

The last version of this patch adds a configuration option (`Settings -> Advanced -> Metadata \0 replacement`) for the replacement character to use, default is `;` (semicolon). When the field is cleared `\0` characters are dropped without replacing them (which is the way _Jaudiotagger_ handles this situation when writing tags).

 [1] notwithstanding the fact that `\u0000` _is_ a legal character in UTF8 (which gives lie to _Postgresql_ claiming to fully support this encoding)

 [2] see _//Wav ID3 tags needs trailing null for each value (not usually required) to work in Windows_ in one of the linked changes.

 [3]  see the `removeTrailingTerminatorOnWrite` parameter, explained as follows: _iTunes needlessly writes null terminators at the end for TextEncodedStringSizeTerminated values, if this option is enabled (which it is by default) these characters are removed_.

Yetangitu:
build failures due to [CVE-2019-16335](https://nvd.nist.gov/vuln/detail/CVE-2019-16335), [CVE-2019-14540](https://nvd.nist.gov/vuln/detail/CVE-2019-14540) - upgrading jackson-databind to 2.10.0.pr3 as a temporary fix. If this CVE does not affect _airsonic_ it should be added to `airsonic-main/cve-suppressed.xml`

tesshucom:
I have no doubts about the policy that Yetangitu is thinking.

What I care about is more , more  tiny.
Additional processing can be clearly separated from existing processing.
Avoid mixing with legacy parse.

Strategy, Filter, Wrapper, Transfer Utility ... anything is fine.
It is not desirable to include logic in the DTO generally.

If ambiguous function division, it will be difficult to understand and no one can touch it.

SearchService is a good example.
Despite many problems and having only a few hundred lines,
no one has touched for a long time.
We have a terrible experience.

Parser is definitely one of the hardest to understand and important.
Cumulative correction adds complexity.
So I think it would be best if it was written in a new class.
pgjdbc should not be a requirement for a base class.

If a new discussion takes place in the future, 
we validate for ``the new class``, not the ``JaudiotaggerParser``.
Should be so.

Sorry for being noisy, there is no objection to the content.

tesshucom:
I have no doubts about the policy that Yetangitu is thinking.

What I care about is more , more  tiny.
Additional processing can be clearly separated from existing processing.
Avoid mixing with legacy parse.

Strategy, Filter, Wrapper, Transfer Utility ... anything is fine.
It is not desirable to include logic in the DTO generally.

If ambiguous function division, it will be difficult to understand and no one can touch it.

SearchService is a good example.
Despite many problems and having only a few hundred lines,
no one has touched for a long time.
We have a terrible experience.
KISS principle and creating untouchable code is not equivalent.

Parser is definitely one of the hardest to understand and important.
Cumulative correction adds complexity.
So I think it would be best if it was written in a new class.
pgjdbc should not be a requirement for a base class.

If a new discussion takes place in the future, 
we validate for ``the new class``, not the ``JaudiotaggerParser``.
Should be so.

Sorry for being noisy, there is no objection to the content.

Yetangitu:
> If a new discussion takes place in the future,
> we validate for `the new class`, not the `JaudiotaggerParser`.
> Should be so.

The filtering is not being done in _JaudiotaggerParser_ but in _MetaData_, this being the most logical spot for it to take place so that _all_ metadata (no matter the source) goes through the mill. Realistically speaking it is highly unlikely for any other source than _JaudiotaggerParser_ to produce `\0` characters given that _FFmpegParser_ (with my patch) produces metadata through _ffprobe_ (which doesn't produce `\0` characters), _DefaultParser_ takes its clues from file and directory names (which by necessity can not contain `\0` characters) and it would take almost malicious intent (or just a clumsy cut-and-paste...?) to feed `\u0000` into the tag editor.

tesshucom:
OK.
If there is a necessity and logic is encapsulated in MetaData, I think the following changes are necessary.
 - Calls settingsService.getNullCharacterSubstitute() only once during scanning.
 - The new arguments added to MetaData are expressed as immutable booleans.

Parsing is done asynchronously to the main thread of Sever.

tesshucom:
OK.
If there is a necessity and logic is encapsulated in MetaData, I think the following changes are necessary.
 - Calls settingsService.getNullCharacterSubstitute() only once during scanning.
 - The new arguments added to MetaData are expressed as immutable booleans.

Parsing is done asynchronously to the main thread of Sever.
Ensures consistent parsing.

tesshucom:
OK.
If there is a necessity and logic is encapsulated in MetaData, I think the following changes are necessary.
 - Calls settingsService.getNullCharacterSubstitute() only once during scanning.
 - The new arguments added to MetaData are expressed as immutable booleans.

Parsing is done asynchronously to the main thread of Sever.
Ensures consistent parsing.
(In other words, the range that must be taken into consideration is expanding.)

Yetangitu:
> OK.
> If there is a necessity and logic is encapsulated in MetaData, I think the following changes are necessary.
> 
>     * Calls settingsService.getNullCharacterSubstitute() only once during scanning.

settingsService.getNullCharacterSubstitute() is called once in the constructor, the result is stored in `nullCharacterSubstitute`. One call per scanned file, consistency guaranteed there but not between files in a scan job. The parameter can be injected through `MediaScannerService` to ensure consistency within a single scan run.

>     * The new arguments added to MetaData are expressed as immutable booleans.

But they're not, unless the code is changed back into what it was before which either dropped `\0` characters or replaced them with a (non-configurable) substitute.

I think you're overcomplicating this issue. Look at the way _Jaudiotagger_ handles a related issue:
 - `TagOptionSingleton` contains boolean to either strip or not strip trailing `\0`
 - `TextEncodedStringSizeTerminated` checks settings object (on every call...) and removes `\0` if needed and, in a somewhat clumsy twist, adds some of those back if they original data ended with a `\0`.

_Jaudiotagger_ does a lot more (replaces certain keywords, adds parenthesis, sometimes changes numeric genre codes to their textual representation, etc) so use made a bunch of special data type objects (`TextEncodedStringSizeTerminated` being an example of such). If _airsonic_ has the intention of becoming a full-fleged tag editor (_should it?_) it might need to go to such lengths to make sure all idiosyncrasies are handled (like '`All TextInformation frames support multiple strings, stored as a null separated list, where null is represented by the termination code for the character encoding. This functionality is only officially support in ID3v24`' - it would also need a way to represent and edit such multi-field tags).

I think a decision needs to be made on how to handle the problem with `\0` characters and, related to that, the way `ID3` tags are handles:

 - the simple solution (which is what this patch represents) is to do `\0` character substitution/removal in `MetaData` or in the `dao` layer (`MediaFileDao`, `AlbumDao` and `ArtistDao`, essentially anything which handles metadata, either by using a simple filter or by defining a new string type which handles the filtering. It works, has good performance and solves the problem but it is not entirely _correct_ in that it changes multi-string tags into single-but-longer-string tags. This is also true for the existing tag editor which doesn't support such multi-string tags.
 - the more complicated solution is to overhaul the tag code to actually support those `\0` characters when used as string separators, i.e. allow multiple values for a single tag. Note that this would also solve the `\0` character problem as the database would now contain interpreted versions of the tags instead of raw tag data.

tesshucom:
At this stage, I think that it is only necessary to solve one point of suppressing the error of pgjdbc.

As Yetangitu is concerned, it is closely related to the id3v2.4 specification.
But I don't think we should write code that is not needed now(Although related to the KISS principle).
Therefore, a design that allows recursive maintenance and less affected by existing code would be happy.

There is the idea that if you want to avoid the complexity of thread synchronization, it should be a startup option rather than a screen option.

Yetangitu:
Given that the most likely - and probably the only - source of `\0` characters in strings is `JaudiotaggerParser` and given that this is most likely related to the use of [`getRawContent()`](http://www.jthink.net/jaudiotagger/javadoc/org/jaudiotagger/tag/TagField.html#getRawContent()) (which can return binary data) instead of [`getFirst()`](http://www.jthink.net/jaudiotagger/javadoc/org/jaudiotagger/tag/Tag.html#getFirst(java.lang.String)) or [`getFields()`](http://www.jthink.net/jaudiotagger/javadoc/org/jaudiotagger/tag/Tag.html#getFields()) it might be possible to solve the problem by modifying `JaudiotaggerParser` instead of implementing a `\0` filter. I'll look into this as it would negate the need for data filtering and lead to a more correct implementation.

Yetangitu:
This is a radically simplified version which filters out trailing `\0` characters in `JaudiotaggerParser`. Thus far I have only see it return _trailing_ `\0` characters so this patch should take care of that problem and with that the problem with _Postgresql_ choking on some files. As to why `Jaudiotagger` returns some fields with trailing `\0` characters while it explicitly strips them when moving data from _WAV info_ tags to _ID3_ is a good question, I consider this to be a bug.

Yetangitu:
Hopefully the last and certainly the simplest version, this patch removed embedded and trailing `\0` characters from string types returned by `Jaudiotagger`.

Yetangitu:
I did some digging in the raw metadata as returned by `Jaudiotagger` for a mixed ~200.000 file collection of recorded lectures, podcasts, radiio archives, audio books and music to find out in which way multi-field metadata is stored. While `ID3v24` explicitly allows `\0` as separator I did not encounter _any_ instance of such use. All the `\0` characters I found were used as terminators in `ID3v24` frames:

```
Description="Script"; Text="Latn\0";
Description="ASIN"; Text="B109854DC6\0";
Description="R128_ALBUM_GAIN"; Text="0\0";
```
Stripping these `\0` characters appears to be the right approach.

Terms where mostly separated by `/`, `;` or `,` but a sizeable number used different characters: `&`, `-`, ` ` (space), `.`, `+`, `:` and possibly some others. There is no method to the madness making an attempt at parsing such multi-term frames fraught with peril as it is sure to be incomplete (as in '_will not recognise a multi-term frame_') as well as error-prone (as in '_will split a frame into multiple terms where no such splitting is warranted_').

_Genre_ frames contain everything from nonsense to monstrosities like the following:

`Guitar Virtuoso, Rock, Blues-Rock/Hard Rock/Fusion/British Blues/Album Rock/Pop-Rock`

Those who like to listen to Jazz and Funk need to cope with `Jazz & Funk`, `Jazz + Funk`, `Jazz/Funk`, `Jazz-Funk`, `Jazz:Funk` and `Jazz Funk`. This is remarkable as I hardly had any such music in the test collection. Conclusion: tags are a mess.

tesshucom:
Certainly it is a thorny problem.

But, It is inefficient to do this in legacy MetaParser.
Depending on server requirements.

The perth implementation is much cooler with Lucene.
Fast, logical code, and more maintainable.

So I don't feel need to do work than necessary with the original MetaParser and I don't think it's modern method.
MetaParser is overprocessed, and some people think that SimpleMetaReader is sufficient.

For the time being, I think the postgresql problem should be solved within a very small scope.
Just because there is a problem with postgres doesn't mean expanding the confusion into a bigger problem.

 - Provides an option for postgres users to convert `\ 0` to`; `.
 - Does not work unless turned on arbitrarily

The initial purpose may be completed with it.

Trial and error with unclear policy is dangerous in core processing.

Yetangitu:
>     * Provides an option for postgres users to convert `\ 0` to`; `.

Judging by the above results it does not make sense to _replace_ `\0` given that the only `\0` characters encountered are _terminating_ nulls. Simple removal (like this patch now does) is all that is needed.

>     * Does not work unless turned on arbitrarily

Given the negligible performance impact - a single call to `String.replace()` performed once per field when scanning a file into the database - and the fact that those `\0` characters don't serve any purpose it doesn't make any sense to make a special case for _Postgresql_.

The data in the database and/or _Lucene_ is not meant for consumption by an `ID3` parser, that task is performed by `Jaudiotagger` or `ffmpeg` or any other future parser implementation. The parsers form the interface between the domain of raw `ID3` data (with or without embedded `\0` characters or other encoding-dependent terminator characters) and the rest of the application.

tesshucom:
If it is Airsonic's policy, that's fine.
I am not complaining about the Yetangitu's content and you are not responsible for everything.

However... it is a change in the Parser specification.
Issues that may be categorized clearly separately from topics such as simple Secure, Modernize, and external integrations.
Because, Input conversion is and a prerequisite for later development and current implementation.

This is an little important issue and a suggestion that it may be labeled special.
Projects where important specification changes are merged without awareness are no longer easy for outsiders to collaborate with.
Outsiders just cooperate and leave, with upper design management in mind.

muff1nman:
What I'd like to see personally is more focus on this from the original issue (emphasis mine):

> wile scanning media folder, Libresonic throws an error and **stops the scan**:

While I think we can all agree that `\0` is an edge case, I don't think its acceptable to have one file prevent a scan from completing. Maybe a better solution is to alert the user files that had issues like these (a simple LOG for now), and move on with scanning the rest of the media collection. This would give the user the opportunity to fix the metadata themselves and then we all win. We don't have to do anything on our end, the data is kept as is and the user has cleaner tags.

PR #1237, opened 2019-09-15
Update media types supported by Jaudiotagger

Yetangitu:
_Jaudiotagger_ does *not* support `ape` or `mp+` while it *does* support `aif/aiff/aifc`, `mp4`, `m4p` and `dsf`. This patch changes the supported types accordingly (and gets rid of a load of error messages caused by _Jaudiotagger_ being fed formats it doesn't understand)

 [edit]

While it does indeed support `mp4` it does not seem to do so in a meaningful way; I removed this type from the list, use ffmpeg for this purpose


Yetangitu:
Another random Travis failure it seems?

eharris:
Nit: Your checkin comment from the second commit would be nicer if it was split into a subject line and then the longer description.  See https://chris.beams.io/posts/git-commit/ for a good commit comment style guide.

FYI: I also kicked off the travis run that failed, and will do the same in the future whenever I see one.

Yetangitu:
I made a new PR which includes these changes as well as enables ffmpegParser to parse metadata in audio files not covered by other parsers, see #1243

PR #1239, opened 2019-09-16
Change to new shuffle generator

tesshucom:
The algorithm will be updated so that the shuffle can be performed with more consideration for "variation".
For user, the phenomenon that the same album continues even though there are many songs is reduced.

The algorithm is determined at startup by the environment.
Linux is almost the strongest because it uses ``/dev/random``.
Displays the algorithm to use at startup.

### Linux

image

### Windows

image

In environments where the new algorithm is not available, the same algorithm is used as before.

Generally, this is a random algorithm used when generating a cipher.
The speed does not change much.

For source code, the subclassing of SearchServiceImpl's utility functions is complete by this PR.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceUtilities.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceUtilities.java
@@ -31,13 +31,21 @@ import org.airsonic.player.service.MediaFileService;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.lucene.document.Document;
 import org.checkerframework.checker.nullness.qual.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+import java.security.Security;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
+import java.util.Random;
 import java.util.function.BiConsumer;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import static org.springframework.util.ObjectUtils.isEmpty;
 
@@ -70,6 +78,45 @@ public class SearchServiceUtilities {
      */
     @Autowired
     private MediaFileService mediaFileService;
+
+    private static final Logger LOG = LoggerFactory.getLogger(SearchServiceUtilities.class);
+
+    private Random random;
+    
+    private static Random createSecureRandom(String algorithm) {
+        SecureRandom secureRandom = null;
+        try {
+            LOG.debug("{} is used to create a random list of songs.", algorithm);
+            secureRandom = SecureRandom.getInstance(algorithm);
+        } catch (NoSuchAlgorithmException e) {
+            LOG.error("Usually unreachable.", e);
+        }
+        return secureRandom;
+    }
+
+    {
+
+        String algorithmNative = "NativePRNG";
+        String algorithmSHA = "SHA1PRNG";
+
+        List<String> algorithms = Arrays.asList(Security.getProviders()).stream()
+                .flatMap(p -> p.getServices().stream().filter(s -> "SecureRandom".equals(s.getType())))
+                .map(s -> s.getAlgorithm()).collect(Collectors.toList());
+
+        if (algorithms.contains(algorithmNative)) {
+            random = createSecureRandom(algorithmNative);
+        } else if (algorithms.contains(algorithmSHA)) {
+            random = createSecureRandom(algorithmSHA);
+        }
+
+        if (isEmpty(random)) {
+            random = new Random(System.currentTimeMillis());
+            LOG.debug("NativePRNG and SHA1PRNG cannot be used on this platform.");
+        }
+
+    }
+
+    public Function<Integer, Integer> nextInt = (range) -> random.nextInt(range);
 
     public final Function<Long, Integer> round = (i) -> {
         // return
--- a/airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceImpl.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceImpl.java
@@ -50,9 +50,6 @@ public class SearchServiceImpl implements SearchService {
     @Autowired
     private SearchServiceUtilities util;
 
-    // TODO Should be changed to SecureRandom?
-    private final Random random = new Random(System.currentTimeMillis());
-
     @Override
     public SearchResult search(SearchCriteria criteria, List<MusicFolder> musicFolders,
             IndexType indexType) {
@@ -108,7 +105,7 @@ public class SearchServiceImpl implements SearchService {
 
         List<D> result = new ArrayList<>();
         while (!docs.isEmpty() && result.size() < count) {
-            int randomPos = random.nextInt(docs.size());
+            int randomPos = util.nextInt.apply(docs.size());
             Document document = searcher.doc(docs.get(randomPos));
             id2ListCallBack.accept(result, util.getId.apply(document));
             docs.remove(randomPos);
--- a/airsonic-main/src/test/java/org/airsonic/player/service/search/SearchServiceUtilitiesTestCase.java
+++ b/airsonic-main/src/test/java/org/airsonic/player/service/search/SearchServiceUtilitiesTestCase.java
@@ -1,0 +1,16 @@
+package org.airsonic.player.service.search;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertNotNull;
+
+public class SearchServiceUtilitiesTestCase {
+
+    SearchServiceUtilities utilities = new SearchServiceUtilities();
+
+    @Test
+    public void testNextInt() {
+        assertNotNull(utilities.nextInt.apply(100));
+    }
+
+}


airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceUtilities.java
```diff
@@ -71,6 +79,47 @@
...
+            } catch (NoSuchAlgorithmException e) {
+                LOG.error("Usually unreachable.", e);
+            }
+        } else if (algorithms.contains(algorithmSHA)) {
+            /* win */
+            try {
+                random = SecureRandom.getInstance(algorithmSHA);
+                LOG.info("SHA1PRNG is used to create a random list of songs.");
+            } catch (NoSuchAlgorithmException e) {
+                LOG.error("Usually unreachable.", e);
```
eharris: Since these two `try .. catch` blocks duplicate the same caught exception and LOG message, I'd probably just use a single `try .. catch` and move the conditional inside it.  That would remove the duplicated code.

eharris: Since these two try .. catch blocks duplicate the same caught exception and LOG message, I'd probably just use a single try .. catch and move the conditional inside it. That would remove the duplicated code.

airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceUtilities.java
```diff
@@ -71,6 +79,47 @@
...
+
+        List<String> algorithms = Arrays.asList(Security.getProviders()).stream()
+                .flatMap(p -> p.getServices().stream().filter(s -> "SecureRandom".equals(s.getType())))
+                .map(s -> s.getAlgorithm()).collect(Collectors.toList());
+
+        if (algorithms.contains(algorithmNative)) {
+            /* linux */
+            try {
+                random = SecureRandom.getInstance(algorithmNative);
+                LOG.info("NativePRNG is used to create a random list of songs.", algorithmNative);
```
eharris: I would probably make all these `LOG.info` messages be `LOG.debug`, as most users won't care about having log messages about what PRNG is in use.

airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceUtilities.java
```diff
@@ -71,6 +79,47 @@
...
+            } catch (NoSuchAlgorithmException e) {
+                LOG.error("Usually unreachable.", e);
+            }
+        } else if (algorithms.contains(algorithmSHA)) {
+            /* win */
```
eharris: Minor nit: I would spell out windows, makes it easier to grep for without false positives
Also for these single line comments, I suggest using single line `//` for comments that don't span lines

airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceUtilities.java
```diff
@@ -71,6 +79,47 @@
...
+            try {
+                random = SecureRandom.getInstance(algorithmSHA);
+                LOG.info("SHA1PRNG is used to create a random list of songs.");
+            } catch (NoSuchAlgorithmException e) {
+                LOG.error("Usually unreachable.", e);
+            }
+        }
+
+        /* others */
+        if (isEmpty(random)) {
```
eharris: I don't think that testing for the list of algorithms being empty is safe here.
The list of algorithms could contain items that are not one of the two methods handled above, which would result in `random` being left un-initialized.
This should probably just be the fallthrough `else` on the above `if`.

tesshucom: Exception cases are included.
It is different from ``else``.

fxthomas:
Can we use `/dev/urandom` on Linux instead? It's probably just as strong as `/dev/random` without the nasty effect of blocking if there is not enough entropy, which _does_ happen, especially on headless servers.

eharris:
I agree with @fxthomas, we shouldn't be exhausting the entropy source for high security needs for something as non-critical as shuffle.

airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceUtilities.java
eharris: I know it is different.  My point is that it isn't safe to test for the list being empty as a fallthrough, as there can be other items that do **not** match the tests above.

tesshucom:
For Solaris/Linux/MacOS, The default is to obtain seed from /dev/random and obtain random numbers from /dev/urandom.

I think the following article is easy to understand and nice..

https://metebalci.com/blog/everything-about-javas-securerandom

tesshucom:
My explanation may have been short of words.
For Solaris/Linux/MacOS, The default is to obtain seed from /dev/random and obtain random numbers from /dev/urandom.
(/dev/random is only once)
Therefore, resources are not exhausted by shuffling songs.

jvoisin:
Using `SecureRandom.getInstance(NativePRNG);` is [fine]( https://docs.oracle.com/javase/10/docs/api/java/security/SecureRandom.html#%3Cinit%3E() ), java doesn't do stupid things wrt. PRNG.

airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceUtilities.java
tesshucom: It is not a specification that you absolutely want to use something other than legacy.

Only one of the two popular algorithms are explicitly used.
Most environments other than the retro JVM are covered one or more of two algorithms.
For encryption requirements.

Even if advanced hardware has a special random number generator, it doesn't use it.
Taking the same method as legacy in an unexpectedly irregular environment is nothing dangerous or anything.

tesshucom: It is not a specification that you absolutely want to use something other than legacy.

Only one of the two popular algorithms are explicitly used.
Most environments other than the retro JVM are covered one or more of two algorithms.
For encryption requirements.

tesshucom:
Functionalization gives good results when the code feels redundant.
This time it was corrected in that direction.

Before, I've seen a try-catch that is bigger than my display on Airsonic.
Incredible...
It is better to refrain from expanding try-catch.

fxthomas:
> The default is to obtain seed from /dev/random and obtain random numbers from /dev/urandom.

Great, I didn't know that! Then it all looks good to me, +1 for merging.

eharris:
Rather than change to a function, it seems to me like it would have been simpler to just iterate over a list of the algorithms to try to apply and left the `try..catch` where it was.  That would have eliminated the duplicated code and also gotten rid of the multiple uses of the same algorithm names.

tesshucom:
Do you want to write more deeply nested code?

tesshucom:
Do you want to write more deeply nested code?

If you want to clarify your intention to recommend less maintainable coding, you will need a coding convention.
You can also control CheckStyle.

fxthomas:
I would be okay with merging this anyway, the code is fine and the difference between the functional and imperative version is minimal in this case.

I am with @eharris re:maintainability though. While I like functional, it sometimes baffles people. More potential contributors are more used to imperative style, which makes functional sometimes _less_ maintainable. In code I _personally_ write I would rather avoid using things like `Function`/`stream`, unless the code is _geological layers_ deep, say at least 5-6 nested levels, and/or the functional version is markedly easier to understand.

tesshucom:
I think it's also related to whether you're used to it.
It is useful to get used to it.

I write the try block as small as possible.
Also replaces loops that can be replaced with lambdas.
If it is written in such a rule, debugging at the time of failure will be easy after all.

One thing that worries me is that I feel that the stance towards the new grammar is different from that of muff1nman.
I've already made quite a lot of commits, but have been often　suggested to replace the new grammar.
Therefore, I use it actively unless it becomes excessive curry.

Before discussing with outsiders, first of all, it is necessary to unify among the maintainer at the center of Airsonic.
If reviewer preferences are disjoint, it is a burden for volunteer committer.
If a simple direction is announced in advance, outsiders will naturally follow it.

tesshucom:
Since there is no plan to use it in lambda, Function was changed to a method.


Lambda can be rewritten as

```
for (Provider provider : Security.getProviders()) {
    for (Service service : provider.getServices()) {
        if ("SecureRandom".equals(service.getType())) {
            if (algorithms.contains(algorithmNative)) {
                random = createSecureRandom(algorithmNative);
                break;
            } else if (algorithms.contains(algorithmSHA)) {
                random = createSecureRandom(algorithmSHA);
            }
        }
    }
}
```
If this is easier to understand, rewrite it?
I have experienced this case many times, so don't mind either.
'for' statements of Java change every few years.

Writing a try catch to this nesting would be awful, but if you recommend it, I'll follow it.

tesshucom:
Temporary hold because of a mysterious error in travis.

almet:
I believe this is due to an error in the build system. Can you please re-trigger the build? (I can't, it seems).

jvoisin:
Nope, it's all checkstyle violations.

airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceUtilities.java
```diff
@@ -104,9 +104,9 @@
                 .map(s -> s.getAlgorithm()).collect(Collectors.toList());
 
         if (algorithms.contains(algorithmNative)) {
-            random = secureRandom.apply(algorithmNative);
+            random = createSecureRandom(algorithmNative);
         } else if (algorithms.contains(algorithmSHA)) {
-            random = secureRandom.apply(algorithmSHA);
+            random = createSecureRandom(algorithmSHA);
```
eharris: This still has multiple uses of the `algorithmXXX` values, so I still think my prior comment suggesting iterating over the list of algorithms to use and applying the first one that matches would make this cleaner.

fxthomas: How about doing something like this?

```diff
diff --git a/airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceUtilities.java b/airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceUtilities.java
index ebc4dba9..abd74f5b 100644
--- a/airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceUtilities.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceUtilities.java
@@ -96,17 +96,14 @@ public class SearchServiceUtilities {
 
     {
 
-        String algorithmNative = "NativePRNG";
-        String algorithmSHA = "SHA1PRNG";
-
         List<String> algorithms = Arrays.asList(Security.getProviders()).stream()
                 .flatMap(p -> p.getServices().stream().filter(s -> "SecureRandom".equals(s.getType())))
                 .map(s -> s.getAlgorithm()).collect(Collectors.toList());
 
-        if (algorithms.contains(algorithmNative)) {
-            random = createSecureRandom(algorithmNative);
-        } else if (algorithms.contains(algorithmSHA)) {
-            random = createSecureRandom(algorithmSHA);
+        for (String algorithmName : new String[]{"NativePRNG", "SHA1PRNG"}) {
+            if (algorithms.contains(algorithmName)) {
+                random = createSecureRandom(algorithmName);
+            }
         }
 
         if (isEmpty(random)) {
```

fxthomas:
The failing checkstyle rules are about using the correct line endings:

```
[ERROR] src/test/java/org/airsonic/player/service/search/SearchServiceUtilitiesTestCase.java:[3] (regexp) RegexpMultiline: Do not use Windows line endings
```

Aside from this and eharris's comment, I think we will soon be able to merge this?

tesshucom:
Rebase and merge were done.

Personally, I don't think Lambda is doing many difficult things.
If you feel strange when you first see it, you probably don't know the SecurityProviders class structure.
That is probably not the essence.

If it improves, you will probably add a filter.
But you can browse the shuffle algorithm owned by the system simply by specifying one breakpoint.
On all VM (including the future) and OS.
From the standpoint of maintainability, it is simple enough and NullPointer does not occur.

The reason for making a decision after obtaining all the algorithms is because of their priority.
Whether NativePRNG or SHA1PRNG is obtained first depends on the VM implementation.

This is very short code and is executed only once at startup.
Therefore, I had wrote with that idea.
I don't think there's enough coding variation to require long discussions.

fxthomas:
As I said earlier, for me this looks good enough to be merged.

But I don't want to brush away opinions without checking first : @eharris, do you think this is good? If not, could you maybe offer some code examples to help us see more easily what we can improve here?

airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceUtilities.java
```diff
@@ -71,6 +79,45 @@
...
 
+    private static final Logger LOG = LoggerFactory.getLogger(SearchServiceUtilities.class);
+
+    private Random random;
+    
+    private static Random createSecureRandom(String algorithm) {
+        SecureRandom secureRandom = null;
+        try {
+            LOG.debug("{} is used to create a random list of songs.", algorithm);
+            secureRandom = SecureRandom.getInstance(algorithm);
```
muff1nman: Why can't we just do new `SecureRandom()` and call it a day?

tesshucom: Regardless of the VM implementation, it is clearly stated that only these two algorithms will be used.
SecureRandom() gives almost the same result, but it depends on the environment setup and VM implementation.

Every environment is likely to hold one of the two algorithms listed here.
And these are algorithms that do not block the system.

fxthomas:
One additional comment about the use of `NativePRNG` : I read [here](https://docs.oracle.com/javase/10/docs/specs/security/standard-names.html#securerandom-number-generation-algorithms) that the non-blocking behavior is not guaranteed. Shouldn't `NativePRNGNonBlocking` be used instead?

tesshucom:
That might be better.
But in case of Win, NoSuchAlgorithmException occurs.
In other words, when adopting that plan, it will be used together with OS judgment for now.

Also can consider SecureRandom.getInstanceStrong () for simple code As muff1nman said.
However, since the specifications differ depending on the version, we may prefer to explicitly specify the algorithm.

 - ...osName.startsWith("linux") && NativePRNGNonBlocking
 - SHA1PRNG
 - Standard Random

If this is okay?

tesshucom:
That might be better.

But in fact, the nature of NativePRNG blocks is platform dependent.
In other words, NativePRNG≒ /dev/urandom continues processing by reusing the random seed used in the past when depleted.
So NativePRNG is used in various VM as high-standard algorithms .
(Virtually no depletion. To be careful old VM before JDK1.7.)

NativePRNGNonBlocking is a latecomer and its availability depends on the VM.
I don't know about the VM implementation status of various Linux distributions.

Also can consider for simple code As muff1nman said.
There is also a method that depends on SecureRandom.getInstanceStrong() without being aware of the algorithm.
However, we do not know what will be returned, including unimplemented methods.

airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceUtilities.java
eharris: Yes, something like this is what I had in mind.  Seems a lot cleaner.  I'd like to see this cleanup done and get this merged.

fxthomas:
@eharris @muff1nman you had objections on merging this, do you still have them?

eharris:
@fxthomas I still haven't seen the simplification/cleanup from your comment above (https://github.com/airsonic/airsonic/pull/1239#discussion_r352832041) applied.  As far as I'm concerned, that's all that I'm waiting on.

tesshucom:
This is going nowhere. Let's try again some other time.

I do not specifically oppose Airsonic's policy,
and are free to develop new features with my own clones.
This is the good thing about open source.


It doesn't seem so important, so close this and OK.
This saves wasted time on proposers and reviewers.
Benefit for each other.

eharris:
Since the submitter doesn't seem to want to apply the requested changes, I've created a new PR with those changes (#1585) to supercede this one.

fxthomas:
@tesshucom Sorry about that. We'll try to get a decision done faster next time.
@eharris Thanks!

tesshucom:
It's fine, don't worry about it.
There's a lot more important than the discussion here, and I just feel that stagnation is undesirable.

Issue #1240, opened 2019-09-16
Integration tests: cucumber-jvm-parallel-plugin not supported


eharris:
The integration tests depend on `cucumber-jvm-parallel-plugin`, but according to the readme at https://github.com/temyers/cucumber-jvm-parallel-plugin that plugin is no longer supported and probably shouldn't be used, and instead upgrading to Cucumber 4.0.0 is recommended.

I wonder if this might be related to whatever has been causing the Travis/CI build unreliability lately.

eharris:
The Cucumber release announcement for 4.0.0 is here: https://cucumber.io/blog/announcing-cucumber-jvm-4-0-0/

randomnicode:
@jvoisin @fxthomas @tesshucom @eharris Here is a PR that gets rid of Cucumber completely and just relies on pure maven failsafe for integration tests: https://github.com/airsonic-advanced/airsonic-advanced/pull/113

My integration test times went from around ~3 mins to ~1 min (factor of 3).

Note that the PR is longer than it needs to be because [airsonic-advanced](https://github.com/airsonic-advanced/airsonic-advanced) does additional testing against external DBs (and thus needs to coordinate and remove some inconsistencies that are in airsonic), but you should be able to just take everything in the `integration-test` folder and make a PR out of it.

eharris:
I would be onboard with doing as @randomnicode suggests and removing cucumber entirely.  But there also seems to be a decent amount of unrelated (or at least not clearly related) stuff that is being touched in the PR he points to.  I vastly prefer well-focused change sets targeted at a single clearly defined objective, as they are much easier for more people to understand and review and verify.

randomnicode:
As mentioned in the comment above:
> you should be able to just take everything in the integration-test folder and make a PR out of it.

That was my suggestion to @fxthomas in another PR also (to just just at the repo folder directly instead of the PR).

The bulk of the touch points in that PR are due to the fact that you don't use external DBs for any testing, which airsonic-advanced does. There are implicit assumptions in the test codebase that don't  work when external dbs are involved and thus require changing. Your integration tests DON'T run against ANY external DB, and thus several underlying issues can be brushed under the rug because they haven't surfaced. airsonic-advanced does and thus it needs to be cognizant of that when setting up its container for the integration tests.

[...]

PR #1242, opened 2019-09-17
Change integration tests from parallel to single forked mode

eharris:
This is an attempt to make the Travis CI integration tests more reliable, and changes from parallel to single forking mode.
This is a temporary measure until cucumber-jvm-parallel-plugin (which is no longer supported) is ripped out (see #1240)

In my limited tests, no longer using parallel forking and 2 threads and instead going to single forking mode makes the integration tests about 2 minutes (12.5 vs 10.5m) slower.  But if that is the cost of fixing the unreliability of the tests, that's not that bad of a trade-off.

### DIFF
--- a/integration-test/pom.xml
+++ b/integration-test/pom.xml
@@ -196,8 +196,8 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
-                    <parallel>all</parallel>
-                    <threadCount>2</threadCount>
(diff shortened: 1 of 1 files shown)


tesshucom:
Good job. Stable operation is the top priority.
Cucumber-parallel is not unnecessary, but it depends on the server configuration.
If this PR is valid, it means that scenario division by yaml side is also valid.

muff1nman:
This is fine, but we should still probably dig into why the tests fail in parallel. There might be something interesting going on there.

Issue #1238, opened 2019-09-16
Travis CI crumbling, time for another option?


Yetangitu:
I'm seeing more and more random build failures which are totally unrelated to submitted changes. By now it is more or less expected for Travis to complain and actually a surprise when it doesn't. Give it a kick and Travis might pop up with another random build failure, if in luck it succeeds. This problem is not limited to _airsonic_, I see it everywhere Travis is used. As to whether this is related to the fact that the company was [bought by Idera](https://news.ycombinator.com/item?id=18978251) about half a year ago I do not know but the frequency of these random failures has made Travis more of a hinder than an asset by now.

I'd say it is time to switch to another CI provider, that HN discussion linked to above points out a few possible candidates:

Circle CI: https://circleci.com/

Buildkite: https://buildkite.com/

Azure Pipelines: https://azure.microsoft.com/en-us/services/devops/pipelines
 - https://docs.microsoft.com/azure/devops/pipelines/migrate/from-travis

Github Actions: https://github.com/features/actions

As far as I know all of these offer free tiers for free software projects.

jvoisin:
Don't forget (the amazing) [gitlab ci]( https://gitlab.com/airsonic/airsonic )

tesshucom:
I learned by #1240 that  Airsonic is trying to run integration tests in parallel.
But seems  to be difficult.
integration tests have many barrier factors to parallelization.
(LFS or flash IO of DB and Lucene etc)

Other CI with very fast IO may work it temporarily, but after a while the same result will be achieved.
(Because of transition competition between users.)

Execution of parallel processing by software is not so versatile.

eharris:
I'm hoping that #1242 will help fix the Travis CI reliability issues by switching from parallel threaded to single forked mode.

Yetangitu:
> Don't forget (the amazing) [gitlab ci](https://gitlab.com/airsonic/airsonic)

Sure, and here I thought that'd imply a move to Gitlab (fine by me but that is another question for another day) and was proven wrong: 

https://docs.gitlab.com/ee/ci/ci_cd_for_external_repos/github_integration.html

jvoisin:
We already have a gitlab mirror anyway :)

eharris:
The change to not do tests in parallel (#1242) seems to have fixed the Travis flakiness, so I'm going to close this as resolved.  If someone wants to pick up moving off Travis to something else, this can either be reopened or a new issue can be created with the specific plan.

PR #1243, opened 2019-09-17
Ffmpegparser audiosupport

Yetangitu:
[This PR supersedes #1237, it contains all changes from that PR]

This patch extends ffmpegParser to playable formats (i.e. those which are configured as `musicFileTypes` and `videoFileTypes`) and enables it to extract metadata from these files. This makes it possible to extract metadata from less common file formats like APE etc (#866).


Yetangitu:
[ignore the Travis failure, see #1238]

BTW, I used MusicBrainz' Picard [list of tag mappings](https://picard.musicbrainz.org/docs/mappings/) in combination with what 'exotic' files I could find to define tag mappings. As I do not have much in the way of 'exotic' files it would be beneficial for those who do to execute `ffprobe -v quiet -print_format json -show_format -show_streams <filename>` on those files to see which - if any - tags they provide and adjust the mappings accordingly for any missing data.

Yetangitu:
Travis, be good or be gone. Also, add a few scope declarations which have been missing since the beginning of time.

airsonic-main/src/main/java/org/airsonic/player/service/metadata/FFmpegParser.java
```diff
@@ -99,6 +101,76 @@ public MetaData getRawMetaData(File file) {
...
+             * tag names are forced to lower case
+             */
+            final JsonNode tags = result.at("/format/tags");
+            if (tags != null) {
+                Iterator<String> it = tags.fieldNames();
+                String key = null;
+                while (it.hasNext()) {
+                    key = it.next();
+                    // skip empty tags
+                    if (StringUtils.isBlank(tags.get(key).asText())) {
```
jvoisin: `tags.get(key).asText()` should be stored in a local variable instead of being repeated everywhere

Yetangitu: That is certainly possible but it doesn't make things faster as it only gets called once per switch cycle and it takes an extra allocation of a local. Anyway, will do. The same can be done in Jaudiotagger (f.toString())

jvoisin: It' more about simplifying/reducing the amount of code than about performances :)

eharris:
I heartily agree and support moving the mapping and parser functionality from `JaudiotaggerParser` to `MetaDataParser`, but I would prefer that that refactoring be its own commit with no other changes.

You probably also want to squash the commits that modify your own changes, and please follow the general style guidelines for commit messages (short simple subject, with later expansion following after a blank line).

And to avoid further Travis failures, you'll want to rebase in the changes from `master`.

Yetangitu:
See PR #1259 for the refactoring of `JaudiotaggerParser` and `MetaDataParser`. This also contains a corrected list of file types for `JaudiotaggerParser` in preparation of a possible merging of this PR and because the current list is just plain wrong.

I'll refactor this patch to exclude these changes but it should be clear that the result will be dependent on #1259 as `FFmpegParser` calls functions which are not present in unmodified `MetaDataParser`.

PR #1247, opened 2019-09-19
WIP: Get rid of jackson2

jvoisin:
Jackson2 has a new critical vulnerability every
two weeks, it's awful. This commits replaces
it with javax.json instead.

### DIFF
--- a/airsonic-main/cve-suppressed.xml
+++ b/airsonic-main/cve-suppressed.xml
@@ -174,11 +174,6 @@
         <cve>CVE-2018-14948</cve>
     </suppress>
     <suppress>
-        <notes>We do not enable default typing for jackson</notes>
-        <gav regex="true">.*jackson-databind.*</gav>
-        <cve>CVE-2019-12814</cve>
-    </suppress>
-    <suppress>
         <notes>We do not use the liquibase sdk</notes>
         <filePath regex="true">.*liquibase/sdk/.*</filePath>
         <cvssBelow>9.0</cvssBelow>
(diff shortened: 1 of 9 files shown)


muff1nman:
Why? As long as we don't use polymorphic typing (we don't last I checked), we can ignore the gadget vulnerabilities.

eharris:
I think the issue he's trying to address is that the build keeps breaking (seems like every couple weeks) due to new Jackson CVE's that we then have to manually exclude to get things working again.

jvoisin:
Exactly what @eharris said

fxthomas:
Isn't `javax.json` proposed for inclusion in the standard library modules? If we don't use the features Jackson provides, then we might as well use a more standard version. +1 from me!

jvoisin:
Spring is depending on Jackson:

```
[INFO] +- org.springframework.boot:spring-boot-starter-web:jar:1.5.21.RELEASE:compile
[INFO] |  +- org.springframework.boot:spring-boot-starter:jar:1.5.21.RELEASE:compile
[INFO] |  |  +- org.springframework.boot:spring-boot:jar:1.5.21.RELEASE:compile
[INFO] |  |  +- org.springframework.boot:spring-boot-autoconfigure:jar:1.5.21.RELEASE:compile
[INFO] |  |  \- org.springframework.boot:spring-boot-starter-logging:jar:1.5.21.RELEASE:compile
[INFO] |  |     +- org.slf4j:jcl-over-slf4j:jar:1.7.26:compile
[INFO] |  |     +- org.slf4j:jul-to-slf4j:jar:1.7.26:compile
[INFO] |  |     \- org.slf4j:log4j-over-slf4j:jar:1.7.26:compile
[INFO] |  +- org.hibernate:hibernate-validator:jar:5.3.6.Final:compile
[INFO] |  |  +- org.jboss.logging:jboss-logging:jar:3.3.2.Final:compile
[INFO] |  |  \- com.fasterxml:classmate:jar:1.3.4:compile
[INFO] |  +- com.fasterxml.jackson.core:jackson-databind:jar:2.8.11.3:compile
[INFO] |  |  +- com.fasterxml.jackson.core:jackson-annotations:jar:2.8.0:compile
[INFO] |  |  \- com.fasterxml.jackson.core:jackson-core:jar:2.8.11:compile
```

Meh.

jvoisin:
But, on the other hand, we can get rid of Jackson2 in our codebase as much as possible anyway.

muff1nman:
But removing jackson in our code base then doesn't really solve the issue then. As we will still get project builds broken when jackson has a vuln.

jvoisin:
Sure, but I would sleep better at night knowing that we're exposing less input to jackson.

muff1nman:
Hm I dunno - jackson is pretty battle tested and very widespread. I sleep well at night knowing we're using the most proliferated json parsing library :smile:

jvoisin:
But it has a [terrible security history]( https://www.cvedetails.com/vendor/15866/Fasterxml.html ), and they're using a blacklist approach to fix the deserialization issues :'(
As @fxthomas said, javax.json [seems to be]( https://mvnrepository.com/artifact/javax.json/javax.json-api ) in the stdlib.

muff1nman:
And how many of those are gadget related? All of the ones I looked at..

PR #1248, opened 2019-09-20
Upgrade jackson-* to 2.10.0.pr3 to fix CVE-2019-16335, CVE-2019-14540

Yetangitu:
build failures due to [CVE-2019-16335](https://nvd.nist.gov/vuln/detail/CVE-2019-16335), [CVE-2019-14540](https://nvd.nist.gov/vuln/detail/CVE-2019-14540) - upgrading _jackson-*_ to 2.10.0.pr3 as a temporary fix. If this CVE does not affect _airsonic_ it should be added to `airsonic-main/cve-suppressed.xml`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -113,17 +113,17 @@
             <dependency>
                 <groupId>com.fasterxml.jackson.core</groupId>
                 <artifactId>jackson-core</artifactId>
-                <version>2.9.9</version>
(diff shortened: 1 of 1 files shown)


jvoisin:
They don't affect airsonic, but since you've updated jackson, no need to add them to the cve blacklist.

Issue #923, opened 2019-03-26
Self-contained JAR broken on master


fxthomas:
The self-contained JAR doesn't work on the latest master (9413f1f4).

The UI shows, but JS error messages show up, no file can be played, and the logs contain tons of errors like this one:

```
2019-03-26 21:55:32.114 ERROR --- o.d.servlet.DwrServlet                   : init failed
org.directwebremoting.extend.ContainerConfigurationException: java.lang.IllegalArgumentException: name
        at org.directwebremoting.impl.StartupUtil.createAndSetupDefaultContainer(StartupUtil.java:194) ~[dwr-3.0.2-RELEASE.jar!/:na]
        at org.directwebremoting.servlet.DwrServlet.createContainer(DwrServlet.java:101) ~[dwr-3.0.2-RELEASE.jar!/:na]
        at org.directwebremoting.servlet.DwrServlet.init(DwrServlet.java:64) ~[dwr-3.0.2-RELEASE.jar!/:na]
        at org.apache.catalina.core.StandardWrapper.initServlet(StandardWrapper.java:1144) [tomcat-embed-core-8.5.35.jar!/:8.5.35]
        at org.apache.catalina.core.StandardWrapper.allocate(StandardWrapper.java:789) [tomcat-embed-core-8.5.35.jar!/:8.5.35]
        at org.apache.catalina.core.StandardWrapperValve.invoke(StandardWrapperValve.java:133) [tomcat-embed-core-8.5.35.jar!/:8.5.35]
        at org.apache.catalina.core.StandardContextValve.invoke(StandardContextValve.java:96) [tomcat-embed-core-8.5.35.jar!/:8.5.35]
        at org.apache.catalina.authenticator.AuthenticatorBase.invoke(AuthenticatorBase.java:493) [tomcat-embed-core-8.5.35.jar!/:8.5.35]
        at org.apache.catalina.core.StandardHostValve.invoke(StandardHostValve.java:140) [tomcat-embed-core-8.5.35.jar!/:8.5.35]
        at org.apache.catalina.valves.ErrorReportValve.invoke(ErrorReportValve.java:81) [tomcat-embed-core-8.5.35.jar!/:8.5.35]
        at org.apache.catalina.core.StandardEngineValve.invoke(StandardEngineValve.java:87) [tomcat-embed-core-8.5.35.jar!/:8.5.35]
[...]

PR #1249, opened 2019-09-21
Bump dwr's version

jvoisin:


fxthomas:
I think we had issues with upgrading DWR last time (https://github.com/airsonic/airsonic/issues/923), are there still present?

PR #1250, opened 2019-09-21
Remove radeox

jvoisin:
Radeox is a rendering engine for an unspecified markup.
Its [website](http://radeox.org/) is dead, the website
of its [authors](http://www.codehaus.org/) is dead too,
its [last commit](https://github.com/codehaus/radeox) was 13 years ago.

It's only used for the welcome and login messages,
as well as comments from users. If we want to have some markup parsing,
we should use something maintained with autoescaping guarantees,
instead of a piece of zombie code prone to $DEITY knows what injections.


eharris:
It should be noted that this is a breaking change (I have added the label for that) because installations that use the existing markup syntax already will no longer be rendered correctly and their text will need to be updated.

muff1nman:
Does it make sense to include a different engine? I have no clue how many people use this feature..

jvoisin:
Worse case, people can always write HTML. We only have a handful of places with user-controled text, I don't think it's worth having a markup-procesing engine for them.

PR #1252, opened 2019-09-21
Use lambdas instead of ghetto nested types

jvoisin:
This change was done in an automated fashion
via IntelliJ IDEA


eharris:
@jvoisin you have checkstyle violations to fix.

PR #1254, opened 2019-09-21
Fix remaining race condition in media playback

### LINKED ISSUE #1160: Last song in playlist repeating.
On current master branch, the last song in the playlist seems to repeat without end.

fxthomas:
**TL;DR:** This commit is hopefully the final fix on Airsonic's side for #685. It removes temporary workarounds introduced in #1080, and as such fixes #1160.

---

The root cause of the issue is the fact that, when we go to the next track in an Airsonic play queue, we change the media source in the `ended` event, as described in more detail in https://github.com/mediaelement/mediaelement/issues/2650.

In MEJS, this translates as the following two things:

* In Airsonic's `ended` event handler, we change the media source (set the `src` attribute) and call the `load()` method, followed by the `play()` method.

* Another `ended` event handler was also used internally by the MEJS player, and one of these internal uses called the `pause()` method (presumably in order to make sure that playback was stopped on some media renderers).

Unfortunately, the order in which these event handlers are called depends (in all modern browsers) on the order in which they are registered.

In our case, the first one is registered inside the `<body>` tag, but the second one is registered with `$(document).ready(...)`. This means that the first event handler is called before the second.

This means that, in some cases (when we're unlucky, hence the seemingly random nature of the bug), `pause()` is called after `load()` but before the media has finished loading.

Apparently, this causes the `AbortError: The fetching process for the media resource was aborted by the user agent at the user's request.` message to appear (which indicates exactly what's described in the last paragraph), and the playback of the next song is aborted.

### DIFF
--- a/airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
+++ b/airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
@@ -197,20 +197,16 @@
     }
 
     function createMediaElementPlayer() {
-
-        var player = $('#audioPlayer').get(0);
+        // Manually run MediaElement.js initialization.
+        //
+        // Warning: Bugs will happen if MediaElement.js is not initialized when
+        // we modify the media elements (e.g. adding event handlers). Running
+        // MediaElement.js's automatic initialization does not guarantee that
+        // (it depends on when we call createMediaElementPlayer at load time).
+        $('#audioPlayer').mediaelementplayer();
 
         // Once playback reaches the end, go to the next song, if any.
-        player.addEventListener("ended", onEnded);
-
-        // FIXME: Remove once https://github.com/mediaelement/mediaelement/issues/2650 is fixed.
-        //
-        // Once a media is loaded, we can start playback, but not before.
-        //
-        // Trying to start playback after a song has endred but before the
-        // 'canplay' event for the next song currently causes a race condition
-        // in the MediaElement.js player (4.2.10).
-        player.addEventListener("canplay", function() { player.play(); });
+        $('#audioPlayer').on("ended", onEnded);
     }
 
     function getPlayQueue() {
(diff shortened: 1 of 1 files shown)


airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
```diff
@@ -197,20 +197,11 @@
     }
 
     function createMediaElementPlayer() {
-
-        var player = $('#audioPlayer').get(0);
+        // Manually run MediaElement.js initialization.
```
jvoisin: It might be worth providing a link to this pullrequest in the comment, to explain why we're manually initializing MEjs :)

fxthomas: Done!

fxthomas: Done!

Adnn:
@fxthomas I built your branch and ran the resulting container on our "live" server, and I don't have the last song repeating anymore on my Chrome browser, whereas it was **always** repeating with the latest release. Thank you, it seems you fixed it!

fxthomas:
Thanks for testing! I assume didn't find other issues with playback (such as the original bug where the next song isn't played)?

Since this is a core feature that had a lot of difficult to debug issues lately, I want to leave it open until this weekend for either myself or other testers to find issues. If nothing happens, I'll merge it.

Adnn:
> I assume didn't find other issues with playback (such as the original bug where the next song isn't played)?

We did not observe any regression in playback behaviour  : )

fxthomas:
I did not notice any issues for a few days now, time to merge this!

fxthomas:
@muff1nman I was thinking we should probably backport this and make a 10.4.1 release. This affects a significant portion of people and the patch applies cleanly on top of the `release-10.4` branch. What do you think?

jvoisin:
Or, a new major release :3

muff1nman:
@fxthomas sure. TBH I was hoping that we could get a 10.5 released soon, but I think it'd be smarter to do a 10.4.1. Can you submit a PR for this to the 10.4 branch? We should probably also pull in the mariadb change as well.

PR #1259, opened 2019-09-22
Refactor metadata, move common functions to MetaDataParser

Yetangitu:
Refactor metadata, move common functions out of `JaudiotaggerParser` to `MetaDataParser`. This change also includes a corrected list of file types supported by `JaudiotaggerParser` in preparation for #1243.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
@@ -32,7 +32,6 @@ import org.jaudiotagger.tag.id3.AbstractID3Tag;
 import org.jaudiotagger.tag.id3.ID3v24Frames;
 import org.jaudiotagger.tag.id3.ID3v24Tag;
 import org.jaudiotagger.tag.images.Artwork;
-import org.jaudiotagger.tag.reference.GenreTypes;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -40,11 +39,7 @@ import org.springframework.stereotype.Service;
 import org.springframework.util.ObjectUtils;
 
 import java.io.File;
-import java.util.SortedSet;
-import java.util.TreeSet;
 import java.util.logging.LogManager;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 /**
  * Parses meta data from audio files using the Jaudiotagger library
@@ -56,9 +51,6 @@ import java.util.regex.Pattern;
 public class JaudiotaggerParser extends MetaDataParser {
 
     private static final Logger LOG = LoggerFactory.getLogger(JaudiotaggerParser.class);
-    private static final Pattern GENRE_PATTERN = Pattern.compile("\\((\\d+)\\).*");
-    private static final Pattern TRACK_NUMBER_PATTERN = Pattern.compile("(\\d+)/\\d+");
-    private static final Pattern YEAR_NUMBER_PATTERN = Pattern.compile("(\\d{4}).*");
     @Autowired
     private final SettingsService settingsService;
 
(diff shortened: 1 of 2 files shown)


tesshucom:
- MetaDataParser -> Domain specific more abstract meta definitions.
 - JaudiotaggerParser -> Additional parsing depending on Jaudiotagger implementation.

Jaudiotagger is especially unique around Genre.

The import statement shows the proper separation of logic.
It may break the proper separation of logic.

eharris:
@tesshucom raises a good point, and `GenreTypes` and `mapGenre` should probably stay in `JaudiotaggerParser`.  Since that will require changes anyway, please also remove the changes to the supported formats list and leave that in your original PR.

Yetangitu:
Other parsers can also return malformed genre tags, the domain here is _genre metadata_ (which in practice mostly comes down to `ID3`) a which is covered by more than just `JaudiotaggerParser`. This is why I placed those two functions and the related import in `MetaDataParser`. Placing them in `JaudiotaggerParser` will lead to duplication of effort and inconsistent results when other parsers implement their own version of a _genre_ filter.

tesshucom:
In such cases, instead of generalizing the implementation of a specific product, define an abstruct class.
It seems strange that FFmpegParser inherits Jaudiotagger, but if want to do.
```
MetaDataParser --┬--  DefaultMetaDataParser
                 └--  AbstructJaudiotaggerParser --┬--  FFmpegParser
                                                   ├--  JaudiotaggerParser
                                                   ├--  ????Parser
                                                   └--  ????Parser
```

This is a boring general theory in order not to affect the existing implementation. An aside.
___

The subtlety is that the MetaData constructor argument is variable.
It seems that you want to encapsulate your logic in MetaData and treat it as an active object.

(Assuming that is necessary)

As mentioned in #1236, all arguments that can change the logic of MetaData should be immutable.
If you use some settingsService's property frequently, you should define one immutable object class that holds those conditions.

Create an instance of  ``immutable`` object only once and share it in all MetaData.
It may be not good to have a reference to ``mutable`` settingsService.

tesshucom:
In such cases, instead of generalizing the implementation of a specific product, define an abstruct class.
It seems strange that FFmpegParser inherits AbstructJaudiotagger, but if want to do.
This is a boring general theory in order not to affect the existing implementation. An aside.
___

The subtlety is that the MetaData constructor argument is variable.
It seems that you want to encapsulate your logic in MetaData and treat it as an active object.

(Assuming that is necessary)

As mentioned in #1236, all arguments that can change the logic of MetaData should be immutable.
If you use some settingsService's property frequently, you should define one immutable object class that holds those conditions.
Create an instance of  ``immutable`` object only once and share it in all MetaData.

(Assuming that is unnecessary)

Since there is no necessity to encapsulate in MetaData, it is defined as a utility externally.
Normally you don't override the Date class for date comparisons.
The same way of thinking.

Yetangitu:
All changes to the `MetaData`object are gone, the logic lives in `MetaDataParser` now.

eharris:
@Yetangitu You have a good point, that Genre ID to Genre string mapping data is generally useful to any implementation of support for Id3v1 and id3V2 tags.  The fact we are choosing to use the list in Jaudiotagger (I like that it is very complete and includes the extended list), shouldn't preclude it being useful in other parsing implementations.  It just means we are choosing to use that as our canonical list.

Because this is the case, I am ok with having the higher level `MetaDataParser` pull in the `GenreTypes` from Jaudiotagger.

tesshucom:
Refactoring usually refers to implementation improvements that do not involve external functional changes.
What is being done is a reorganization of class functions without UT.

I do not oppose it.
But you can consider changing the class name at the same time if you make a corresponding change.
Remove product names from JaudiotaggerParser and FFmpegParser to make them more abstract etc.
If you feel resistance, need to rethink the meaning .

 - Changing the visibility of MetaDataParser method ``guess*`` is a bad sign. 
   If you adjust this visibility to work without change, can assume that it is the correct design.

Yetangitu:
The product names in `JaudiotaggerParser` and `FFmpegParser` should stay as those classes _are_ tightly linked to the related products and so help to identify them as such. The abstraction which they're built on is `MetaDataParser` which is product-neutral, it contains helper functions to be used by all implementations. The `guessArtist()`, `guessAlbum()` and `guessTitle()` functions (which may have originated in `DefaultMetaDataParser`, they seem to form the core functionality of that specific parser) are used as backup to 'fill in the blanks' in `MetaDataParser.getMetaData()` and in `DefaultMetaDataParser` and don't need to be visible outside of the `metadata` package, i.e. the modifier can be removed.

tesshucom:
In other words there is a contradiction(it is a matter of degree).
Everything that originally existed in JaudiotaggerParser is based on the product specifications of JaudiotaggerParser.
That's it for class design.

Changing method visibility is a bigger problem to be involved in package design.
If a new feature requires it, it suggests that the feature should be present in the parser package.

As I mentioned elsewhere, I hope these changes can be explicitly labeled "design changes".
At least it is evidence that the reviewer understands the basic object design and approves the change.

Note that changes of lower layer such as parsers and domains have a wide impact.

Yetangitu:
I think this has been mentioned before, but...

`JaudiotaggerParser` is an `ID3` parser first and foremost and as such contains parts of the `ID3` specification (like _genre mappings_). Those parts are not so much specific to `JaudiotaggerParser` as they are to the `ID3` specification. Since that specification is needed by other parser implementations as well it makes sense to separate those parts from `JaudiotaggerParser` into either `MetaDataParser` (where they live now) or into a separate object (like the way these are implemented in `Jaudiotagger`). The latter is possible but it would lead to a duplication of effort since `Jaudiotagger` already provides the needed code/data in a separate import (`org.jaudiotagger.tag.reference.GenreTypes`). Seen from a _purity_ perspective the latter would be more _clean_ but if that is the objective the whole of _airsonic_ is due for an overhaul (see: _second system effect_).

I changed the reference in `EditTagsController` to remove a (now superflous) reference to and import of `JaudiotaggerParser` so that it accesses these functions directly from `MetaDataParser` (which it already imports), this change is not in any PR but I can submit one if needed or add it to this patch. The call to `guessTitle()` in `EditTagsController` already goes through the generic parser (which has access to this function since it lives in `MetaDataParser` now) so in this respect `EditTagsController` is one step closer to being parser-implementation independent. The way _airsonic_ currently handles tags is a bit of a mixed bag and seems to be geared towards `ID3v1` with single fields per tag. Given that there have been requests to support a wider variety of tags (_composer_, multiple fields per tag, etc) it probably makes more sense to overhaul the tagging code as a whole to make this possible.

tesshucom:
I do not deny Yetangitu's thoughts.
If you need it, you should.
No matter what method you choose, after all, Airsonic's view.

However, it is recommended that Airsonic project declare core part design changes clearly and confidently.

Useless discussion can be avoided as long as it is clearly labeled that the design policy is different from the legacy.
Whatever it is.

This is beneficial not only for the parties but also for outsiders.

muff1nman:
Well if we're gonna argue semantics (which is good for long term airsonic) I'd like to see MetaDataParser become an interface and then move any common functionality to a package protected utility class where static functions can be used al a carte by specific implementation classes. If the state/logic is too intertwined for pulling out static functions - it is fine to use an additional abstract class but it should also be package protected so that it is clear that it is merely an implementation detail that can be changed on a whim.

eharris:
@Yetangitu are you maintaining this PR?  It has merge conflicts to be resolved, as well as requested changes.

Yetangitu:
I'll look into this one of these days...

eharris:
I'd like to see this get applied sometime soon.  Can someone else review please?

Yetangitu:
I've been running this patch for a long time now, it works as intended. I see Travis complains about something, will look in to this 'one of these days'.

PR #1263, opened 2019-09-24
Update translation

jooola:
Manually updated translations from weblate. Documentation is coming in few minutes.

Something is bothering me with the Asian characters. Not sure if they work as expected. The script could not convert them to unicode, not sure why yet.


eharris:
It should be noted that the fixes in #1300 will probably need to be added to this PR so as not to undo the fixes there.

jooola:
@eharris I will rebase once this is done, but I have some troubles with the translations since they are not all formatted and rendered the same way. Anyway, I'll investigate a bit more how we could fix this.

PR #1264, opened 2019-09-27
Expunge search index

tesshucom:
Unnecessary data in Lucene is deleted when "Clean-up database" is executed.
Subsonic doesn't have this feature.
Index data is infinitely added.

While effective in reducing disk size, it also has the effect of preventing degradation of search functionality due to long-term operation. 

 - The increase in useless data has almost no effect on Lucene search speed.
 - However, since Lucene does not manage the ``present`` flag, perform null-if-select to the database.
 - When useless data increases, meaningless SELECT increases.

If there is too much useless data after long-term operation,
for example, the display speed of the random album immediately
after the login screen will be affected.


fxthomas:
I've been running this for a few days without issues. While I'm not too familiar with Lucene, the code/tests look good and ready for merging?

tesshucom:
I think there is no problem with the structure.

It can be confirmed that the target unnecessary data has also been deleted from the test data discharged after the test.
In addition, it is confirmed that all Lucene data is deleted by deleting all MusicFolders on the actual server and performing "Scan"-> "Clean-up database".
(Lecene data directory size will not be zero)

I publish it as a clone and some people use it.

PR #1265, opened 2019-10-01
Add discogs search link to main album page

### LINKED ISSUE #586: FR: Discogs.com
Hi

It would be great if discogs.com is also offered as a context provider.

thanks

benferreri:
Small change to close #586. 
I just added a link to search discogs for the album, under type "release." Discogs can be helpful for searching for physical copies of the album or additional info not included on other sites, as discussed in the issue.

### DIFF
--- a/airsonic-main/src/main/webapp/WEB-INF/jsp/albumMain.jsp
+++ b/airsonic-main/src/main/webapp/WEB-INF/jsp/albumMain.jsp
@@ -245,21 +245,21 @@
 
     <c:if test="${not empty model.artist and not empty model.album}">
         <sub:url value="https://www.google.com/search" var="googleUrl" encoding="UTF-8">
-            <sub:param name="q" value="\"${model.artist}\" \"${model.album}\""/>
+            <sub:param name="q" value="\"${fn:escapeXml(model.artist)}\" \"${fn:escapeXml(model.album)}\""/>
         </sub:url>
         <sub:url value="https://en.wikipedia.org/wiki/Special:Search" var="wikipediaUrl" encoding="UTF-8">
-            <sub:param name="search" value="\"${model.album}\""/>
+            <sub:param name="search" value="\"${fn:escapeXml(model.album)}\""/>
             <sub:param name="go" value="Go"/>
         </sub:url>
-        <sub:url value="allmusic.view" var="allmusicUrl">
-            <sub:param name="album" value="${model.album}"/>
+        <sub:url value="https://www.allmusic.com/search/albums/%22${fn:escapeXml(model.artist)}%22+%22${fn:escapeXml(model.album)}%22" var="allmusicUrl">
         </sub:url>
+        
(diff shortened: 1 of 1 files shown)


airsonic-main/src/main/webapp/WEB-INF/jsp/albumMain.jsp
```diff
@@ -258,10 +258,15 @@
             <sub:param name="q" value="\"${model.artist}\" \"${model.album}\""/>
             <sub:param name="type" value="album"/>
         </sub:url>
+        <sub:url value="https://www.discogs.com/search/" var="discogsUrl" encoding="UTF-8">
+            <sub:param name="q" value="\"${model.artist}\" \"${model.album}\""/>
```
jvoisin: Please escape those values: if there is a `"` in an artist or album, things will go wrong™.

benferreri: Not sure what you mean; how would I better escape them? I tried to just model it off of how the last.fm link was formatted. I just tested the David Bowie album "Heroes" (has quotes in the title) and it seemed to work fine. Also seemed to work when I changed the album metadata so that there is only one quotation mark in the title.

jvoisin: Escape via `${fn:escapeXml(model.artist)}` and `${fn:escapeXml(model.album)}`, otherwise, a malicious album/artist name could result in an XSS

benferreri:
Okay, I escaped the values for discogs as well as all the other links. I think I did that properly, but let me know if not. Also, I realized that the allmusic link was not working at all, so I fixed that. It was a little ugly because the url value does not get UTF-8 encoded (only the parameters do), so I had to use `%22` for `"`. Let me know if that causes unintended complications.

fxthomas:
I'll merge this, thanks a lot for the help!

airsonic-main/src/main/webapp/WEB-INF/jsp/albumMain.jsp
```diff
@@ -245,21 +245,21 @@
...
+            <sub:param name="q" value="\"${fn:escapeXml(model.artist)}\" \"${fn:escapeXml(model.album)}\""/>
         </sub:url>
         <sub:url value="https://en.wikipedia.org/wiki/Special:Search" var="wikipediaUrl" encoding="UTF-8">
-            <sub:param name="search" value="\"${model.album}\""/>
+            <sub:param name="search" value="\"${fn:escapeXml(model.album)}\""/>
             <sub:param name="go" value="Go"/>
         </sub:url>
-        <sub:url value="allmusic.view" var="allmusicUrl">
-            <sub:param name="album" value="${model.album}"/>
+        <sub:url value="https://www.allmusic.com/search/albums/%22${fn:escapeXml(model.artist)}%22+%22${fn:escapeXml(model.album)}%22" var="allmusicUrl">
```
eharris: This probably shouldn't be one big line with character encodings, it should use use a template substitution so it's understandable.

PR #1266, opened 2019-10-02
Fix incorrect absolute paths in css

### LINKED ISSUE #1256: CSS file causes fetch for nonexistent spin.png
## Problem description

I noticed in the web console that browsing pages using the webapp causes a failed fetch for `/script/spin.png` on just about every page load.

`jquery-3.4.0.min.js:2 GET https://someserver.name/script/spin.png 404 (Not Found)`

After further investigation, it looks like this was caused by change 15a1e5e7d63083f7b11a041d9ce877dc3b5a696f where `#` was used on some added lines.  Perhaps that was meant to be a comment, and just used the wrong comment style?

eharris:
Fixes #1256


eharris:
I don't think it is, because everything else works correctly.  And you will note that the other paths in that same file are also relative, not absolute.

muff1nman:
Consistency above all else.

PR #1274, opened 2019-10-06
Ensure commons logging is excluded

muff1nman:
Ensures that we don't have commons-logging anywhere. 


jvoisin:
Why do you want to exclude this?

muff1nman:
All logging should be through slf4j interfaces which routes all logging to the logback implementation. There should only be one logging implementation on the classpath.

PR #1278, opened 2019-10-06
Remove unnecessary exceptions, again

jvoisin:


airsonic-main/src/main/java/org/airsonic/player/service/MediaFileService.java
```diff
@@ -542,13 +541,9 @@ private MediaFile createMediaFile(File file) {
...
-                    try {
-                        File coverArt = findCoverArt(children);
-                        if (coverArt != null) {
-                            mediaFile.setCoverArtPath(coverArt.getPath());
-                        }
-                    } catch (IOException x) {
-                        LOG.error("Failed to find cover art.", x);
+                    File coverArt = findCoverArt(children);
+                    if (coverArt != null) {
+                        mediaFile.setCoverArtPath(coverArt.getPath());
```
muff1nman: This seems more like a logic change that shouldn't be included with the more trivial changes.

jvoisin: Not really: checkstyle complains about it, since the exception can't be raised anymore.

muff1nman: Ah I see what you mean.

PR #1281, opened 2019-10-06
WIP: Add a preference to disable Google Cast

jvoisin:
This commit adds a preference to disable/enable
support for the Google Cast protocol.

It's disabled by default, since I'm not comfortable
with the default airsonic setup doing requests
to Google stuff.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/SettingsService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/SettingsService.java
@@ -123,6 +123,8 @@ public class SettingsService {
     private static final String KEY_RECAPTCHA_SITE_KEY = "ReCaptchaSiteKey";
     private static final String KEY_RECAPTCHA_SECRET_KEY = "ReCaptchaSecretKey";
 
+    private static final String KEY_GOOGLE_CAST_ENABLED = "GoogleCast";
+
     // Database Settings
     private static final String KEY_DATABASE_CONFIG_TYPE = "DatabaseConfigType";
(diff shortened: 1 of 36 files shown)


muff1nman:
Should this not be wrapped up with the other changes in #993 ? Also I don't think having this disabled by default is good for the first run of this. Maybe in a subsequent release this can be changed.

jvoisin:
I think that doing it step by step is a better approach than trying to do everything at once.

airsonic-main/src/main/java/org/airsonic/player/service/SettingsService.java
```diff
@@ -123,6 +123,8 @@
     private static final String KEY_RECAPTCHA_SITE_KEY = "ReCaptchaSiteKey";
     private static final String KEY_RECAPTCHA_SECRET_KEY = "ReCaptchaSecretKey";
 
+    private static final String KEY_GOOGLE_CAST_ENABLED = "GoogleCast";
```
fxthomas: Minor nit: I would rather call it `GoogleCastEnabled` or something to that effect.

airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
```diff
@@ -687,13 +706,13 @@
         currentStreamUrl = song.streamUrl;
         updateCurrentImage();
 
-        // Handle ChromeCast player.
-        if (CastPlayer.castSession) {
+<c:if test="${model.googleCast}">
```
fxthomas: The mixing of JSP and JS conditions is hard to read. How about doing something like setting the Google Cast instance to `null` at initialization if it's not enabled, and checking on that in JS only?

PR #1282, opened 2019-10-06
WIP: Get rid of com.jayway.jsonpath

jvoisin:


muff1nman:
I think this one is dependent on #1247 . Cause if we keep jackson around, then we should be consistent and use jackson across the board.

airsonic-main/src/main/java/org/airsonic/player/service/VersionService.java
```diff
@@ -249,7 +251,13 @@ private void readLatestVersion() throws IOException {
...
         }
 
-        List<String> unsortedTags = JsonPath.read(content, JSON_PATH);
+        final JsonReader jsonreader = Json.createReader(new StringReader(content));
+        JsonArray jsonobject = jsonreader.readArray();
+        List<String> unsortedTags = new ArrayList<>();
+        for (int j = 0; j < jsonobject.size(); j++) {
+            unsortedTags.add(jsonobject.getJsonObject(j).getString("tag_name"));
+        }
+        jsonreader.close()
```
randomnicode: I don't think this is an equivalent implementation of the jsonpath specified in JSON_PATH

JSON_PATH = "$..tag_name" (line 227)

$is document root
*..* is the recursive descent operator
tag_name is the node name

JSON_PATH selects all instances of tag_name in the input JSON document/string no matter where in the hierarchy it is.

Consider this object:
```
{
  tag_name: "bla",
  more: {
    tag_name: 3,
    stuff: {
      additional: {
        tag_name: {
          anobject: true
        }
      }
    }
  }
}
```

JSON_PATH will query and select all 3 tag_name nodes (the string, the number and the object).

Don't think this implementation is equivalent

PR #1286, opened 2019-10-07
Make test results quieter

muff1nman:

### DIFF
--- a/airsonic-main/src/test/java/org/airsonic/player/TestCaseUtils.java
+++ b/airsonic-main/src/test/java/org/airsonic/player/TestCaseUtils.java
@@ -4,6 +4,8 @@ import org.airsonic.player.controller.JAXBWriter;
 import org.airsonic.player.dao.DaoHelper;
 import org.airsonic.player.service.MediaScannerService;
 import org.apache.commons.io.FileUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.context.ApplicationContext;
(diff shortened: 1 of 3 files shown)


airsonic-main/src/test/java/org/airsonic/player/TestCaseUtils.java
```diff
@@ -16,6 +18,8 @@
 
 public class TestCaseUtils {
 
+    private static final Logger logger = LoggerFactory.getLogger(TestCaseUtils.class);
```
eharris: The majority of the rest of the codebase seems to use `LOG` as the logger variable rather than `logger`.  I don't really care which it is, but for the sake of consistency, would it be better to stick with the majority convention?

muff1nman: Probably - `logger` is just a sub conscious habit for me :) Might need to add a checkstyle so we can enforce.

muff1nman: Proposal: Build some custom checkstyles. See https://github.com/airsonic/checkstyle. Only gotcha is that we would probably need to publish this to the maven central.

eharris: I like the checkstyle approach, that is a good way to enforce going forward!

muff1nman:
This was merged in 85f85fd88caac06ffe2628b86c130767eff946ff.

PR #1289, opened 2019-10-07
Replace custom circular buffer with Apache commons(CircularFifoQueue) 

### LINKED ISSUE #1284: Get rid of airsonic's custom circular buffer implementation
Airsonic has its [own implementation]( https://github.com/airsonic/airsonic/blob/master/airsonic-main/src/main/java/org/airsonic/player/util/BoundedList.java ) of a circular buffer.

It would be nice to replace it with [either]( https://stackoverflow.com/questions/5498865/size-limited-queue-that-holds-last-n-elements-in-java ) the one from Apache Commons or Guava.

kessiler:
Closes #1284 
----

Guidelines for Contributing
---------------------------


airsonic-main/src/main/java/org/airsonic/player/domain/TransferStatus.java
```diff
@@ -298,5 +298,9 @@ public SampleHistory(SampleHistory other) {
             super(HISTORY_LENGTH);
             addAll(other);
         }
+
+        public Sample getLast() {
+            return this.get(this.size());
```
kessiler: Had to add this method as the data structure has changed to a queue

airsonic-main/pom.xml
```diff
@@ -164,6 +164,12 @@
             <artifactId>commons-collections</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-collections4</artifactId>
```
jvoisin: I'm not sure that we want to have two versions of the same package. Can't you use `commons-collections` instead?

kessiler: CircularFifoQueue is just available in commons-collections4.
I left both packages due to incompatibility with the CollectionsUtils class.
Do you want to keep a single package? I can do that. I would rather keep collections4 which is more powerful.

jvoisin: Keep the latest commons-collection one if possible then :)

kessiler: @jvoisin should be good to go ;)

PR #1290, opened 2019-10-07
Minor refactor of CsrfSecurityRequestMatcher

jvoisin:
- use plain string comparison instead of regexp
- Simplify data structures
- Simplify boolean condition

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/security/CsrfSecurityRequestMatcher.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/security/CsrfSecurityRequestMatcher.java
@@ -32,8 +32,8 @@ public class CsrfSecurityRequestMatcher implements RequestMatcher {
 
     @Override
     public boolean matches(HttpServletRequest request) {
-        boolean isCsrfNeeded = allowedMethods.contains(request.getMethod()) ||
+        boolean skipCSRF = allowedMethods.contains(request.getMethod()) ||
             whiteListedMatchers.stream().anyMatch(matcher -> matcher.matches(request));
-        return !isCsrfNeeded;
+        return !skipCSRF;
     }
 }
--- a/airsonic-main/src/main/java/org/airsonic/player/security/CsrfSecurityRequestMatcher.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/security/CsrfSecurityRequestMatcher.java
@@ -6,9 +6,8 @@ import org.springframework.stereotype.Component;
 
 import javax.servlet.http.HttpServletRequest;
 
-import java.util.ArrayList;
-import java.util.Collection;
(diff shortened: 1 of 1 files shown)


airsonic-main/src/main/java/org/airsonic/player/security/CsrfSecurityRequestMatcher.java
```diff
@@ -20,24 +19,20 @@
...
     @Override
     public boolean matches(HttpServletRequest request) {
-
-        boolean skipCSRF =
-                allowedMethods.matcher(request.getMethod()).matches() ||
-                whiteListedMatchers.stream().anyMatch(matcher -> matcher.matches(request));
-
-        return !skipCSRF;
+        return !allowedMethods.contains(request.getMethod()) &&
+                whiteListedMatchers.stream().noneMatch(matcher -> matcher.matches(request));
```
muff1nman: Tbh: I liked the previous boolean logic. It was easier to read than the double negative proposed here.

eharris: I agree that the original logic was easier to read, but I also like getting rid of the temporary variable.  Maybe a decent compromise would be to use the original logic and just negate the whole thing on the return?

muff1nman: Why is the temporary variable an issue?

muff1nman: In fact I think the temporary variable is what improves the readability greatly.

eharris: I just prefer to avoid using temp variables where possible.  But you are right, it does help readability, so is probably justified in this case.

airsonic-main/src/main/java/org/airsonic/player/security/CsrfSecurityRequestMatcher.java
```diff
@@ -20,24 +19,21 @@
...
     public boolean matches(HttpServletRequest request) {
-
-        boolean skipCSRF =
-                allowedMethods.matcher(request.getMethod()).matches() ||
-                whiteListedMatchers.stream().anyMatch(matcher -> matcher.matches(request));
-
-        return !skipCSRF;
+        boolean isCsrfNeeded = allowedMethods.contains(request.getMethod()) ||
+            whiteListedMatchers.stream().anyMatch(matcher -> matcher.matches(request));
+        return !isCsrfNeeded;
```
muff1nman: But its not isCsrfNeeded. The method should return true when csrf is needed.

PR #1307, opened 2019-10-11
Replace deprecated Date by Calendar

jvoisin:
Apparently, Date.setYear and Date.getYear are
deprecated in favour of Calendar, since Java6!

Source: https://docs.oracle.com/javase/6/docs/api/java/util/Date.html

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/SubsonicRESTController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/SubsonicRESTController.java
@@ -53,6 +53,7 @@ import org.subsonic.restapi.PodcastStatus;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletRequestWrapper;
 import javax.servlet.http.HttpServletResponse;
+import javax.xml.datatype.XMLGregorianCalendar;
 
 import java.util.*;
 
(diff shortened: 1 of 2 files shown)


airsonic-main/src/main/java/org/airsonic/player/controller/SubsonicRESTController.java
```diff
@@ -164,13 +165,13 @@ public void getLicense(HttpServletRequest request, HttpServletResponse response)
...
-
         license.setEmail("airsonic@github.com");
         license.setValid(true);
-        Date farFuture = new Date();
-        farFuture.setYear(farFuture.getYear() + 100);
-        license.setLicenseExpires(jaxbWriter.convertDate(farFuture));
-        license.setTrialExpires(jaxbWriter.convertDate(farFuture));
+        Calendar calendar = Calendar.getInstance();
+        calendar.set(Calendar.YEAR, calendar.get(Calendar.YEAR) + 1900 + 100);
+        XMLGregorianCalendar georgianCalendar = jaxbWriter.convertCalendar(calendar);
```
eharris: While I think you meant this to be `gregorianCalendar`, I would prefer if you kept the name as something more meaningful, like the original `farFuture`, or maybe `farFutureCalendarDate`.

And I think you could avoid having the `calendar` variable as well as simplifying the date calculation by chaining the method calls:
`XMLGregorianCalendar farFuture = jaxbWriter.convertCalendar(Calendar.getInstance().add(Calendar.YEAR, 100));`

jvoisin: Unfortunately, the `.add` method returns `void` :/

PR #1309, opened 2019-10-11
Use the try-with-resource construct where possible

jvoisin:


airsonic-main/src/main/java/org/airsonic/player/service/VersionService.java
```diff
@@ -193,16 +193,11 @@ private String readLineFromResource(String resourceName) {
         if (in == null) {
```
eharris: Maybe it would be worth trying to move this function and combine it in a more generic form with the very similar `readLines()` in StringUtil.java?

airsonic-main/src/main/java/org/airsonic/player/service/VersionService.java
```diff
@@ -188,23 +185,12 @@ public boolean isNewBetaVersionAvailable() {
      * @param resourceName The resource name.
      * @return The first line of the resource.
      */
-    private String readLineFromResource(String resourceName) {
-        InputStream in = VersionService.class.getResourceAsStream(resourceName);
-        if (in == null) {
+    private String readLineFromResource(@NonNull String resourceName) {
+        try (BufferedReader br = new BufferedReader(new FileReader(resourceName))) {
```
fxthomas: Does that do the same thing as `getResourceAsStream`?

jvoisin: I think it does, yes, but I prefer this construction.

fxthomas: The reason I ask is that it looks like `FileReader` is expecting a path on a real filesystem, but `resourceName` could be a path to a resource inside the .war file.

PR #1310, opened 2019-10-11
Display file id and path in debug scanning logs

fxthomas:
Why have 100s of lines with "Detected unmodified file" with no additional info if we're spammed with debug logs anyway? :smile: 

Plus this could be useful for debugging scanning issues along with the debug logger settings in https://github.com/airsonic/airsonic-docs/pull/77.


fxthomas:
Example with debug loggers on, after running `touch` on one file to update its last-modified date:

image

PR #1321, opened 2019-10-12
Minor optimisations for MediaFileService

jvoisin:
- Factorize a comparison so we don't have to run it twice
- Use modern sorting method, since we don't care about Java7 anymore
- Simplify a condition

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/MediaFileService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/MediaFileService.java
@@ -196,21 +196,19 @@ public class MediaFileService {
         List<MediaFile> result = new ArrayList<MediaFile>();
         for (MediaFile child : mediaFileDao.getChildrenOf(parent.getPath())) {
             child = checkLastModified(child, useFastCache);
-            if (child.isDirectory() && includeDirectories && includeMediaFile(child)) {
-                result.add(child);
-            }
-            if (child.isFile() && includeFiles && includeMediaFile(child)) {
-                result.add(child);
+            if (includeMediaFile(child)) {
+                if (child.isDirectory() && includeDirectories) {
+                    result.add(child);
+                }
+                if (child.isFile() && includeFiles) {
+                    result.add(child);
+                }
             }
         }
 
         if (sort) {
             Comparator<MediaFile> comparator = new MediaFileComparator(settingsService.isSortAlbumsByYear());
-            // Note: Intentionally not using Collections.sort() since it can be problematic on Java 7.
-            // http://www.oracle.com/technetwork/java/javase/compatibility-417013.html#jdk7
-            Set<MediaFile> set = new TreeSet<MediaFile>(comparator);
-            set.addAll(result);
-            result = new ArrayList<MediaFile>(set);
+            result.sort(comparator);
         }
 
         return result;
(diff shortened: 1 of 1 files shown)


airsonic-main/src/main/java/org/airsonic/player/service/MediaFileService.java
```diff
@@ -196,21 +196,19 @@ private MediaFile checkLastModified(MediaFile mediaFile, boolean useFastCache) {
         List<MediaFile> result = new ArrayList<MediaFile>();
         for (MediaFile child : mediaFileDao.getChildrenOf(parent.getPath())) {
             child = checkLastModified(child, useFastCache);
-            if (child.isDirectory() && includeDirectories && includeMediaFile(child)) {
-                result.add(child);
-            }
-            if (child.isFile() && includeFiles && includeMediaFile(child)) {
-                result.add(child);
+            if (includeMediaFile(child)) {
+                if (child.isDirectory() && includeDirectories) {
```
eharris: I think the reason I didn't do this initially was that for efficiency the lighter weight boolean check should be before the much heavier regex and string manipulations, since I think this is more often called with either files or dirs enabled but not both.

airsonic-main/src/main/java/org/airsonic/player/service/MediaFileService.java
```diff
@@ -196,21 +196,19 @@ private MediaFile checkLastModified(MediaFile mediaFile, boolean useFastCache) {
...
         }
 
         if (sort) {
             Comparator<MediaFile> comparator = new MediaFileComparator(settingsService.isSortAlbumsByYear());
-            // Note: Intentionally not using Collections.sort() since it can be problematic on Java 7.
-            // http://www.oracle.com/technetwork/java/javase/compatibility-417013.html#jdk7
-            Set<MediaFile> set = new TreeSet<MediaFile>(comparator);
-            set.addAll(result);
-            result = new ArrayList<MediaFile>(set);
+            result.sort(comparator);
```
eharris: Might still be worth leaving a comment here about Java7 concern even if we no longer support it.

airsonic-main/src/main/java/org/airsonic/player/service/MediaFileService.java
```diff
@@ -415,10 +413,7 @@ public boolean includeMediaFile(MediaFile candidate) {
 
     public boolean includeMediaFile(File candidate) {
         String suffix = FilenameUtils.getExtension(candidate.getName()).toLowerCase();
-        if (!isExcluded(candidate) && (FileUtil.isDirectory(candidate) || isAudioFile(suffix) || isVideoFile(suffix))) {
-            return true;
-        }
-        return false;
+        return !isExcluded(candidate) && (FileUtil.isDirectory(candidate) || isAudioFile(suffix) || isVideoFile(suffix));
```
eharris: Minor nit but might be nice to break this line into two so it's not as wide.

airsonic-main/src/main/java/org/airsonic/player/service/MediaFileService.java
jvoisin: After this commit, there is only one callsite, with files and dirs set to true.

jvoisin: If #1322 is merged, I think that there won't be any performance-critical callsite not using both files and dirs set to true.

airsonic-main/src/main/java/org/airsonic/player/service/MediaFileService.java
jvoisin: This pattern is used in a lot of other places, I don't think it's worth keeping the comment, especially since the public version of Java7 isn't supported [since 2015]( https://en.wikipedia.org/wiki/Java_version_history ).

PR #1322, opened 2019-10-12
WIP: Optimise a bit mediascanner

jvoisin:


airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -237,10 +237,7 @@ private void scanFile(MediaFile file, MusicFolder musicFolder, Date lastScanned,
         indexManager.index(file);
 
         if (file.isDirectory()) {
-            for (MediaFile child : mediaFileService.getChildrenOf(file, true, false, false, false)) {
-                scanFile(child, musicFolder, lastScanned, albumCount, genres, isPodcast);
-            }
-            for (MediaFile child : mediaFileService.getChildrenOf(file, false, true, false, false)) {
+            for (MediaFile child : mediaFileService.getChildrenOf(file, true, true, false, false)) {
```
eharris: Are we sure that this doesn't have side effects?  It seems unlikely that this was done explicitly separately unless there was a reason to.

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -159,64 +159,56 @@ private void doScanLibrary() {
         LOG.info("Starting to scan media library.");
         Date lastScanned = DateUtils.truncate(new Date(), Calendar.SECOND);
         LOG.debug("New last scan date is " + lastScanned);
+        // Maps from artist name to album count.
+        Map<String, Integer> albumCount = new HashMap<String, Integer>();
+        Genres genres = new Genres();
 
-        try {
```
eharris: I don't think this `try` .. `catch` is useless.  It makes sure that the finalization code (`finally` block) gets run even if an exception occurs during the processing.

PR #1323, opened 2019-10-12
Make some fields final when possible

jvoisin:
This was done automatically via IntelliJ IDEA


jvoisin:
It's not really worth it tbh.

PR #1338, opened 2019-10-16
Renamed more link to refresh (what it actually does)

eharris:
Also added `Refresh` SVG icons to match the text instead of reusing a not very appropriate css style.

Before:

Screen Shot 2019-10-16 at 00 11 57

After:

Screen Shot 2019-10-16 at 00 09 34


eharris:
I realized I had a mismatched `<td>` block, so fixed that.

Issue #1179, opened 2019-07-28
Automatic page refresh timer doesn't get reset by activity


eharris:
## Problem description

Performing actions on the currently displayed page in the web UI does not reset the timer for the automatic page refresh.  This is most problematic when using the random feature since the page content may change unexpectedly while you are actively doing things, but may also be problematic in other circumstances (I'm not sure if there are other places where the automatic refresh is used).  The related setting is on the `Personal` settings page and is called `Delay when reloading album list`.

### Steps to reproduce

1. Set the `Delay when reloading album list` value on your `Personal` settings page to some reasonable value for testing (like 60 seconds).
2. Go to the home page (using the `Random` display tab).
3. Every few seconds, select a new album from the list to play, resetting your play queue.
4. Observe that the page will refresh and the content will change when the time set in step 1 expires when counting from the time you originally loaded the home page, rather than counting from the time you last performed a play action.

This occurs using the current master head version.

eharris:
This is fixed/obviated by #1339.

Issue #729, opened 2018-04-14
Disable auto refresh web tab tab or windows isn't focused


jooola:
## Problem description

We should not auto reload the home page when not focused...

eharris:
This is fixed/obviated by #1339.

PR #1339, opened 2019-10-16
Removed the auto-refresh feature

### LINKED ISSUE #1003: Add a simple toggle to disable autorefresh
It would be nice to be able to easily disable autorefresh temporarily.

eharris:
This feature just generated a lot of unnecessary traffic to the server and hurt usability because the page would refresh unexpectedly while a user was actively looking at content.  Fixes #1003, #1179, #729.

Did not remove the column for that setting in the user database (just coded around it), since that would lead to complications if needing to downgrade for some reason.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/dao/UserDao.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/dao/UserDao.java
@@ -224,7 +224,7 @@ public class UserDao extends AbstractDao {
                 settings.getAvatarScheme().name(), settings.getSystemAvatarId(), settings.getChanged(),
                 settings.isShowArtistInfoEnabled(), settings.isAutoHidePlayQueue(),
                 settings.isViewAsList(), settings.getDefaultAlbumList().getId(), settings.isQueueFollowingSongs(),
-                settings.isShowSideBar(), settings.getListReloadDelay(), settings.isKeyboardShortcutsEnabled(),
+                settings.isShowSideBar(), 60 /* Unused listReloadDelay */, settings.isKeyboardShortcutsEnabled(),
                 settings.getPaginationSize());
     }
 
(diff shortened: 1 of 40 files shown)


fxthomas:
Great, that's a good idea! From a UX perspective, can you make the active top link clickable (right now if e.g. "Recently played" is the current page the link is not)?

eharris:
@fxthomas:  I don't really want to change the design of the active top link (your first suggestion) in this PR since that is completely unrelated.

And while I agree that your second suggestion of adding a refresh link in all the pages (like the random page already has) would be a good idea, doing so in a consistent place and fixing the table elements to have appropriate html spacing in all circumstances while also maintaining well formatted blocks in the jsp would require a pretty complete overhaul of the whole `homePager.jsp`, and I'd rather not pollute this PR with that.  I do have a PR ready to go to add that though, and can submit it after this PR is merged since otherwise it would cause merge conflicts.

fxthomas:
Okay, fair enough. Looks good to me then!

airsonic-main/src/main/java/org/airsonic/player/dao/UserDao.java
```diff
@@ -224,7 +224,7 @@ public void updateUserSettings(UserSettings settings) {
                 settings.getAvatarScheme().name(), settings.getSystemAvatarId(), settings.getChanged(),
                 settings.isShowArtistInfoEnabled(), settings.isAutoHidePlayQueue(),
                 settings.isViewAsList(), settings.getDefaultAlbumList().getId(), settings.isQueueFollowingSongs(),
-                settings.isShowSideBar(), settings.getListReloadDelay(), settings.isKeyboardShortcutsEnabled(),
+                settings.isShowSideBar(), 60 /* Unused listReloadDelay */, settings.isKeyboardShortcutsEnabled(),
```
muff1nman: Can we remove this column instead?

eharris: As I noted in the original PR comment, I chose not to remove the column and just code around it so that it would not cause a problem if users downgraded or otherwise switched between versions.  I think it would be appropriate to remove it when we are preparing for the next major release, but for now I think it's safer not to.

eharris:
This has two approving reviews, and I think I've addressed @muff1nman's concern, so I'm inclined to merge this. Anyone else care to do the honors?

airsonic-main/src/main/java/org/airsonic/player/dao/UserDao.java
muff1nman: Can you maybe add a todo for us? Maybe a github issue so we can track it?

PR #1342, opened 2019-10-17
Refactor a bit VersionService.java

jvoisin:
- Tighten the exceptions
- Remove a useless global variable
- Use try-with-resource

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/VersionService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/VersionService.java
@@ -38,6 +38,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.text.DateFormat;
+import java.text.ParseException;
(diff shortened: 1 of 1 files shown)


airsonic-main/src/main/java/org/airsonic/player/service/VersionService.java
```diff
@@ -249,7 +244,7 @@ private void readLatestVersion() throws IOException {
             return;
         }
 
-        List<String> unsortedTags = JsonPath.read(content, JSON_PATH);
+        List<String> unsortedTags = JsonPath.read(content, "$..tag_name");
```
eharris: I think it was clearer with `JSON_PATH`.

jvoisin:
Then there isn't anything to change here :)

eharris:
The "try-with-resource" was still a good change! :-)

jvoisin:
It will eventually show up again, it's ok :)

Issue #1340, opened 2019-10-16
External (MySQL+Postgresql) database MUCH slower than hsql AND airsonic  10.4.1 slower than 10.4.


userjp123:
## Problem description
External (MySQL+Postgresql) database much slower than hsql 
AND 
airsonic  10.4.1 slower than 10.4.

### Steps to reproduce
120000 media files

__airsonic Version 10.4.1__
Scan with:
- HSQL: __2 min__
   2019-10-14 13:11:49.449  INFO --- o.a.p.s.MediaScannerService : Starting to scan media library.
   2019-10-14 13:13:44.995  INFO --- o.a.p.s.MediaScannerService : Scanned media library with 120000 entries.

- MySQL+Postgresql: __90 min__
   2019-10-16 09:01:19.074  INFO --- o.a.p.s.MediaScannerService : Starting to scan media library.
   2019-10-16 10:32:24.525  INFO --- o.a.p.s.MediaScannerService : Scanned media library with 120000 entries.

__airsonic Version 10.4.0__
Scan with:
- MySQL: __40 min__

## System information

 * **Airsonic version**: 10.4.0 + 10.4.1 
 * **Operating system**: Ubuntu Server 16.04 LTS
 * **Java version**: Java 8

fxthomas:
Is this a scan from scratch on a new instance, or is this a rescan over an existing collection?

hamfactorial:
I've had the same experience scanning my library of 12,469 files (mostly FLAC, some mp3s).  The daily scan ran for 96 minutes, according to the Docker logs.

I had the default HSQL database set up, then set up a fresh 10.4.1 Docker instance (via the linuxserver :latest image) to test out MariaDB (via linuxserver:latest image).  Since an in-place database migration isn't supported, I simply re-added my library so this can be considered a fresh installation.

userjp123:
> Is this a scan from scratch on a new instance, or is this a rescan over an existing collection?

There was no significant difference between a scan on a new instance or a rescan.

I am in an evaluation phase and test around, what software I want to use. (I think airsonic is great)
The advantage of an external database is too big, I do not want to go back to HSQL.

Because of this thread I made the tests with PostgreSQL: https://github.com/airsonic/airsonic/issues/860
Result: No difference

I also tested with different jdbc-drivers and some jdbc-url-parameters: No difference.

It's strange:
Because of support of MariaDB in the release-changes, I decided to use airsonic 10.4.1 with an external DB.
[...]

PR #1345, opened 2019-10-18
Automatically upgrade HSQLDB to 2.5.0

fxthomas:
I am piggybacking on @harawata's work in #876 to add an automatic upgrade to HSQLDB 2.5.0, rebased on current master.

I tried to make this as safe and controlled as possible, and I've also been running HSQLDB 2.5.0 in production with the same upgrade method for a few weeks now, without noticing issues.

Thoughts and tests are most welcome!

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/Application.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/Application.java
@@ -14,7 +14,7 @@ import org.springframework.boot.autoconfigure.jmx.JmxAutoConfiguration;
 import org.springframework.boot.autoconfigure.liquibase.LiquibaseAutoConfiguration;
 import org.springframework.boot.autoconfigure.web.servlet.MultipartAutoConfiguration;
 import org.springframework.boot.builder.SpringApplicationBuilder;
-import org.springframework.boot.context.event.ApplicationEnvironmentPreparedEvent;
+import org.springframework.boot.context.event.ApplicationPreparedEvent;
 import org.springframework.boot.web.server.WebServerFactoryCustomizer;
 import org.springframework.boot.web.servlet.FilterRegistrationBean;
 import org.springframework.boot.web.servlet.ServletRegistrationBean;
@@ -167,8 +167,10 @@ public class Application extends SpringBootServletInitializer implements WebServ
 
     private static SpringApplicationBuilder doConfigure(SpringApplicationBuilder application) {
         // Handle HSQLDB database upgrades from 1.8 to 2.x before any beans are started.
-        application.application().addListeners((ApplicationListener<ApplicationEnvironmentPreparedEvent>) event -> {
-            LegacyHsqlUtil.upgradeHsqldbDatabaseSafely();
+        application.application().addListeners((ApplicationListener<ApplicationPreparedEvent>) event -> {
+            if (event.getApplicationContext().getEnvironment().acceptsProfiles("legacy")) {
+                LegacyHsqlUtil.upgradeHsqldbDatabaseSafely();
+            }
         });
 
         // Customize the application or call application.sources(...) to add sources
--- a/airsonic-main/src/main/java/org/airsonic/player/Application.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/Application.java
@@ -1,6 +1,7 @@
 package org.airsonic.player;
 
 import org.airsonic.player.filter.*;
+import org.airsonic.player.util.LegacyHsqlUtil;
 import org.directwebremoting.servlet.DwrServlet;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -13,11 +14,13 @@ import org.springframework.boot.autoconfigure.jmx.JmxAutoConfiguration;
 import org.springframework.boot.autoconfigure.liquibase.LiquibaseAutoConfiguration;
 import org.springframework.boot.autoconfigure.web.servlet.MultipartAutoConfiguration;
 import org.springframework.boot.builder.SpringApplicationBuilder;
+import org.springframework.boot.context.event.ApplicationEnvironmentPreparedEvent;
 import org.springframework.boot.web.server.WebServerFactoryCustomizer;
 import org.springframework.boot.web.servlet.FilterRegistrationBean;
 import org.springframework.boot.web.servlet.ServletRegistrationBean;
 import org.springframework.boot.web.servlet.server.ConfigurableServletWebServerFactory;
 import org.springframework.boot.web.servlet.support.SpringBootServletInitializer;
+import org.springframework.context.ApplicationListener;
 import org.springframework.context.annotation.Bean;
 import org.springframework.util.ReflectionUtils;
 
@@ -163,6 +166,11 @@ public class Application extends SpringBootServletInitializer implements WebServ
     }
 
     private static SpringApplicationBuilder doConfigure(SpringApplicationBuilder application) {
+        // Handle HSQLDB database upgrades from 1.8 to 2.x before any beans are started.
+        application.application().addListeners((ApplicationListener<ApplicationEnvironmentPreparedEvent>) event -> {
+            LegacyHsqlUtil.upgradeHsqldbDatabaseSafely();
+        });
+
         // Customize the application or call application.sources(...) to add sources
         // Since our example is itself a @Configuration class (via @SpringBootApplication)
         // we actually don't need to override this method.
--- a/airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
@@ -39,7 +39,6 @@ public class LegacyHsqlDaoHelper extends GenericDaoHelper {
             LOG.debug("Database shutdown in progress...");
             JdbcTemplate jdbcTemplate = getJdbcTemplate();
             try (Connection conn = DataSourceUtils.getConnection(jdbcTemplate.getDataSource())) {
-                conn.setAutoCommit(true);
                 jdbcTemplate.execute("SHUTDOWN");
             }
             LOG.debug("Database shutdown complete.");
--- a/airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
@@ -31,29 +31,26 @@ public class LegacyHsqlDaoHelper extends GenericDaoHelper {
         LOG.debug("Database checkpoint complete.");
     }
 
-    @PreDestroy
-    public void onDestroy() {
-        Connection conn = null;
+    /**
+     * Shutdown the embedded HSQLDB database. After this has run, the database cannot be accessed again from the same DataSource.
+     */
+    private void shutdownHsqldbDatabase() {
         try {
-            // Properly shutdown the embedded HSQLDB database.
             LOG.debug("Database shutdown in progress...");
             JdbcTemplate jdbcTemplate = getJdbcTemplate();
-            conn = DataSourceUtils.getConnection(jdbcTemplate.getDataSource());
-            conn.setAutoCommit(true);
-            jdbcTemplate.execute("SHUTDOWN");
+            try (Connection conn = DataSourceUtils.getConnection(jdbcTemplate.getDataSource())) {
+                conn.setAutoCommit(true);
+                jdbcTemplate.execute("SHUTDOWN");
+            }
             LOG.debug("Database shutdown complete.");
 
         } catch (SQLException e) {
-            LOG.error("Database shutdown failed: " + e);
-            e.printStackTrace();
-
-        } finally {
-            try {
-                if (conn != null)
-                    conn.close();
-            } catch (Exception ex) {
-                ex.printStackTrace();
-            }
+            LOG.error("Database shutdown failed", e);
         }
+    }
+
+    @PreDestroy
+    public void onDestroy() {
+        shutdownHsqldbDatabase();
     }
 }
--- a/airsonic-main/src/main/java/org/airsonic/player/service/SettingsService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/SettingsService.java
@@ -286,12 +286,24 @@ public class SettingsService {
         return home.contains("libresonic") ? "libresonic" : "airsonic";
     }
 
+    public static String getDefaultJDBCPath() {
+        return getAirsonicHome().getPath() + "/db/" + getFileSystemAppName();
+    }
+
     public static String getDefaultJDBCUrl() {
-        return "jdbc:hsqldb:file:" + getAirsonicHome().getPath() + "/db/" + getFileSystemAppName() + ";sql.enforce_size=false";
+        return "jdbc:hsqldb:file:" + getDefaultJDBCPath() + ";sql.enforce_size=false";
     }
 
     public static int getDefaultUPnPPort() {
         return Optional.ofNullable(System.getProperty(KEY_UPNP_PORT)).map(x -> Integer.parseInt(x)).orElse(DEFAULT_UPNP_PORT);
+    }
+
+    public static String getDefaultJDBCUsername() {
+        return "sa";
+    }
+
+    public static String getDefaultJDBCPassword() {
+        return "";
     }
 
     public static File getLogFile() {
--- a/airsonic-main/src/main/java/org/airsonic/player/service/SettingsService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/SettingsService.java
@@ -287,7 +287,7 @@ public class SettingsService {
     }
 
     public static String getDefaultJDBCUrl() {
-        return "jdbc:hsqldb:file:" + getAirsonicHome().getPath() + "/db/" + getFileSystemAppName();
+        return "jdbc:hsqldb:file:" + getAirsonicHome().getPath() + "/db/" + getFileSystemAppName() + ";sql.enforce_size=false";
     }
 
     public static int getDefaultUPnPPort() {
--- a/airsonic-main/src/main/java/org/airsonic/player/spring/DatabaseConfiguration.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/spring/DatabaseConfiguration.java
@@ -51,8 +51,8 @@ public class DatabaseConfiguration {
         BasicDataSource dataSource = new BasicDataSource();
         dataSource.setDriverClassName("org.hsqldb.jdbcDriver");
         dataSource.setUrl(SettingsService.getDefaultJDBCUrl());
-        dataSource.setUsername("sa");
-        dataSource.setPassword("");
+        dataSource.setUsername(SettingsService.getDefaultJDBCUsername());
+        dataSource.setPassword(SettingsService.getDefaultJDBCPassword());
         return dataSource;
     }
 
--- a/airsonic-main/src/main/java/org/airsonic/player/spring/DatabaseConfiguration.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/spring/DatabaseConfiguration.java
@@ -13,7 +13,6 @@ import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.DependsOn;
 import org.springframework.context.annotation.Profile;
 import org.springframework.jdbc.datasource.DataSourceTransactionManager;
-import org.springframework.jdbc.datasource.DriverManagerDataSource;
 import org.springframework.jdbc.datasource.lookup.JndiDataSourceLookup;
 import org.springframework.transaction.annotation.EnableTransactionManagement;
 
@@ -49,7 +48,7 @@ public class DatabaseConfiguration {
     @Bean
     @Profile("legacy")
     public DataSource legacyDataSource() {
-        DriverManagerDataSource dataSource = new DriverManagerDataSource();
+        BasicDataSource dataSource = new BasicDataSource();
         dataSource.setDriverClassName("org.hsqldb.jdbcDriver");
         dataSource.setUrl(SettingsService.getDefaultJDBCUrl());
         dataSource.setUsername("sa");
--- a/airsonic-main/src/main/java/org/airsonic/player/util/LegacyHsqlUtil.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/util/LegacyHsqlUtil.java
@@ -111,7 +111,7 @@ public class LegacyHsqlUtil {
 
         String timestamp = DateTimeFormatter.ofPattern("yyyyMMddHHmmss").format(LocalDateTime.now());
         Path source = Paths.get(SettingsService.getDefaultJDBCPath()).getParent();
-        Path destination = Paths.get(String.format("%s.backup.%s", SettingsService.getDefaultJDBCPath(), timestamp));
+        Path destination = source.resolveSibling(String.format("%s.backup.%s", source.getFileName().toString(), timestamp));
 
         LOG.debug("Performing HSQLDB database backup...");
         FileUtils.copyDirectory(source.toFile(), destination.toFile());
--- a/airsonic-main/src/main/java/org/airsonic/player/util/LegacyHsqlUtil.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/util/LegacyHsqlUtil.java
@@ -1,0 +1,167 @@
+package org.airsonic.player.util;
+
+import org.airsonic.player.service.SettingsService;
+import org.apache.commons.io.FileUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.sql.*;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.Properties;
+
+public class LegacyHsqlUtil {
+
+    private static final Logger LOG = LoggerFactory.getLogger(LegacyHsqlUtil.class);
+
+    /**
+     * Return the current version of the HSQLDB database, as reported by the database properties file.
+     */
+    public static String getHsqldbDatabaseVersion() {
+        Properties prop = new Properties();
+        File configFile = new File(SettingsService.getDefaultJDBCPath() + ".properties");
+        if (!configFile.exists()) {
+            LOG.debug("HSQLDB database doesn't exist, cannot determine version");
+            return null;
+        }
+        try (InputStream stream = new FileInputStream(configFile)) {
+            prop.load(stream);
+            return prop.getProperty("version");
+        } catch (IOException e) {
+            LOG.error("Failed to determine HSQLDB database version", e);
+            return null;
+        }
+    }
+
+    /**
+     * Create a new connection to the HSQLDB database.
+     */
+    public static Connection getHsqldbDatabaseConnection() throws SQLException {
+        String url = SettingsService.getDefaultJDBCUrl();
+        Properties properties = new Properties();
+        properties.put("user", SettingsService.getDefaultJDBCUsername());
+        properties.put("password", SettingsService.getDefaultJDBCPassword());
+        return DriverManager.getConnection(url, properties);
+    }
+
+    /**
+     * Check if a HSQLDB database upgrade will occur and backups are needed.
+     *
+     * DB   Driver      Likely reason                                Decision
+     * null -           new db or non-legacy                         false
+     * -    null or !2  something went wrong, we better make copies  true
+     * 1.x  2.x         this is the big upgrade                      true
+     * 2.x  2.x         already up to date                           false
+     *
+     * @return true if a database backup/migration should be performed
+     */
+    public static boolean isHsqldbDatabaseUpgradeNeeded() {
+        // Check the current database version
+        String currentVersion = getHsqldbDatabaseVersion();
+        if (currentVersion == null) {
+            LOG.debug("HSQLDB database not found, skipping upgrade checks");
+            return false;
+        }
+
+        // Check the database driver version
+        String driverVersion = null;
+        try {
+            Driver driver = DriverManager.getDriver(SettingsService.getDefaultJDBCUrl());
+            driverVersion = String.format("%d.%d", driver.getMajorVersion(), driver.getMinorVersion());
+            if (driver.getMajorVersion() != 2) {
+                LOG.warn("HSQLDB database driver version {} is untested ; trying to connect anyway, this may upgrade the database from version {}", driverVersion, currentVersion);
+                return true;
+            }
+        } catch (SQLException e) {
+            LOG.warn("HSQLDB database driver version cannot be determined ; trying to connect anyway, this may upgrade the database from version {}", currentVersion, e);
+            return true;
+        }
+
+        // Log what we're about to do and determine if we should perform a controlled upgrade with backups.
+        if (currentVersion.startsWith(driverVersion)) {
+            // If we're already on the same version as the driver, nothing should happen.
+            LOG.debug("HSQLDB database upgrade unneeded, already on version {}", driverVersion);
+            return false;
+        } else if (currentVersion.startsWith("2.")) {
+            // If the database version is 2.x but older than the driver, the upgrade should be relatively painless.
+            LOG.debug("HSQLDB database will be silently upgraded from version {} to {}", currentVersion, driverVersion);
+            return false;
+        } else if ("1.8.0".equals(currentVersion) || "1.8.1".equals(currentVersion)) {
+            // If we're on a 1.8.0 or 1.8.1 database and upgrading to 2.x, we're going to handle this manually and check what we're doing.
+            LOG.info("HSQLDB database upgrade needed, from version {} to {}", currentVersion, driverVersion);
+            return true;
+        } else {
+            // If this happens we're on a completely untested version and we don't know what will happen.
+            LOG.warn("HSQLDB database upgrade needed, from version {} to {}", currentVersion, driverVersion);
+            return true;
+        }
+    }
+
+    /**
+     * Perform a backup of the HSQLDB database, to a timestamped directory.
+     * @return the path to the backup directory
+     */
+    public static Path performHsqldbDatabaseBackup() throws IOException {
+
+        String timestamp = DateTimeFormatter.ofPattern("yyyyMMddHHmmss").format(LocalDateTime.now());
+        Path source = Paths.get(SettingsService.getDefaultJDBCPath()).getParent();
+        Path destination = Paths.get(String.format("%s.backup.%s", SettingsService.getDefaultJDBCPath(), timestamp));
+
+        LOG.debug("Performing HSQLDB database backup...");
+        FileUtils.copyDirectory(source.toFile(), destination.toFile());
+        LOG.info("HSQLDB database backed up to {}", destination.toString());
+
+        return destination;
+    }
+
+    /**
+     * Perform an in-place database upgrade from HSQLDB 1.x to 2.x.
+     */
+    public static void performHsqldbDatabaseUpgrade() throws SQLException {
+
+        LOG.debug("Performing HSQLDB database upgrade...");
+
+        // This will upgrade HSQLDB on the first connection. This does not
+        // use Spring's DataSource, as running SHUTDOWN against it will
+        // prevent further connections to the database.
+        try (Connection conn = getHsqldbDatabaseConnection()) {
+            LOG.debug("Database connection established. Current version is: {}", conn.getMetaData().getDatabaseProductVersion());
+            // On upgrade, the official documentation recommends that we
+            // run 'SHUTDOWN SCRIPT' to compact all the database into a
+            // single SQL file.
+            //
+            // In practice, if we don't do that, we did not observe issues
+            // immediately but after the upgrade.
+            LOG.debug("Shutting down database (SHUTDOWN SCRIPT)...");
+            try (Statement st = conn.createStatement()) {
+                st.execute("SHUTDOWN SCRIPT");
+            }
+        }
+
+        LOG.info("HSQLDB database has been upgraded to version {}", getHsqldbDatabaseVersion());
+    }
+
+    /**
+     * If needed, perform an in-place database upgrade from HSQLDB 1.x to 2.x after having created backups.
+     */
+    public static void upgradeHsqldbDatabaseSafely() {
+        if (LegacyHsqlUtil.isHsqldbDatabaseUpgradeNeeded()) {
+            try {
+                performHsqldbDatabaseBackup();
+            } catch (Exception e) {
+                throw new RuntimeException("Failed to backup HSQLDB database before upgrade", e);
+            }
+            try {
+                performHsqldbDatabaseUpgrade();
+            } catch (Exception e) {
+                throw new RuntimeException("Failed to upgrade HSQLDB database", e);
+            }
+        }
+    }
+}
--- a/airsonic-main/src/main/java/org/airsonic/player/util/LegacyHsqlUtil.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/util/LegacyHsqlUtil.java
@@ -4,11 +4,11 @@ import org.airsonic.player.service.SettingsService;
 import org.apache.commons.io.FileUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.core.io.FileSystemResource;
+import org.springframework.core.io.support.PropertiesLoaderUtils;
 
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.IOException;
-import java.io.InputStream;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.sql.*;
@@ -24,15 +24,14 @@ public class LegacyHsqlUtil {
      * Return the current version of the HSQLDB database, as reported by the database properties file.
      */
     public static String getHsqldbDatabaseVersion() {
-        Properties prop = new Properties();
         File configFile = new File(SettingsService.getDefaultJDBCPath() + ".properties");
         if (!configFile.exists()) {
             LOG.debug("HSQLDB database doesn't exist, cannot determine version");
             return null;
         }
-        try (InputStream stream = new FileInputStream(configFile)) {
-            prop.load(stream);
-            return prop.getProperty("version");
+        try {
+            Properties properties = PropertiesLoaderUtils.loadProperties(new FileSystemResource(configFile));
+            return properties.getProperty("version");
         } catch (IOException e) {
             LOG.error("Failed to determine HSQLDB database version", e);
             return null;
--- a/airsonic-main/src/test/java/org/airsonic/player/api/jukebox/AbstractAirsonicRestApiJukeboxIntTest.java
+++ b/airsonic-main/src/test/java/org/airsonic/player/api/jukebox/AbstractAirsonicRestApiJukeboxIntTest.java
@@ -70,7 +70,6 @@ public abstract class AbstractAirsonicRestApiJukeboxIntTest {
     private static String AIRSONIC_API_VERSION;
 
     private static boolean dataBasePopulated;
-    private static DaoHelper staticDaoHelper;
 
     @Autowired
     protected PlayerService playerService;
@@ -101,13 +100,6 @@ public abstract class AbstractAirsonicRestApiJukeboxIntTest {
         dataBasePopulated = false;
     }
 
-    @AfterClass
-    public static void cleanDataBase() {
-        staticDaoHelper.getJdbcTemplate().execute("DROP SCHEMA PUBLIC CASCADE");
-        staticDaoHelper = null;
-        dataBasePopulated = false;
-    }
-
     /**
      * Populate test datas in the database only once.
      *
@@ -119,7 +111,6 @@ public abstract class AbstractAirsonicRestApiJukeboxIntTest {
      */
     private void populateDatabase() {
         if (!dataBasePopulated) {
-            staticDaoHelper = daoHelper;
 
             assertThat(musicFolderDao.getAllMusicFolders().size()).isEqualTo(1);
             MusicFolderTestData.getTestMusicFolders().forEach(musicFolderDao::createMusicFolder);
@@ -147,6 +138,12 @@ public abstract class AbstractAirsonicRestApiJukeboxIntTest {
         testJukeboxPlayer.getPlayQueue().addFiles(true,
                 mediaFileDao.getSongsForAlbum("_DIR_ Ravel", "Complete Piano Works"));
         assertThat(testJukeboxPlayer.getPlayQueue().size()).isEqualTo(2);
+    }
+
+    @After
+    public void cleanDataBase() {
+        daoHelper.getJdbcTemplate().execute("DROP SCHEMA PUBLIC CASCADE");
+        dataBasePopulated = false;
     }
 
     protected abstract void createTestPlayer();
--- a/airsonic-main/src/test/java/org/airsonic/player/dao/PodcastDaoTestCase.java
+++ b/airsonic-main/src/test/java/org/airsonic/player/dao/PodcastDaoTestCase.java
@@ -134,10 +134,10 @@ public class PodcastDaoTestCase extends DaoTestCaseBean2 {
 
         List<PodcastEpisode> episodes = podcastDao.getEpisodes(channelId);
         assertEquals("Error in getEpisodes().", 4, episodes.size());
-        assertEpisodeEquals(a, episodes.get(0));
-        assertEpisodeEquals(c, episodes.get(1));
-        assertEpisodeEquals(b, episodes.get(2));
-        assertEpisodeEquals(d, episodes.get(3));
+        assertEpisodeEquals(d, episodes.get(0));
+        assertEpisodeEquals(a, episodes.get(1));
+        assertEpisodeEquals(c, episodes.get(2));
+        assertEpisodeEquals(b, episodes.get(3));
     }
 
 
--- a/airsonic-main/src/test/java/org/airsonic/player/dao/UserDaoTestCase.java
+++ b/airsonic-main/src/test/java/org/airsonic/player/dao/UserDaoTestCase.java
@@ -120,7 +120,7 @@ public class UserDaoTestCase extends DaoTestCaseBean2 {
         assertUserEquals(user, newUser);
 
         assertNull("Error in getUserByName().", userDao.getUserByName("sindre2", true));
-        assertNull("Error in getUserByName().", userDao.getUserByName("sindre ", true));
+        assertNotNull("Error in getUserByName().", userDao.getUserByName("sindre ", true));
         assertNull("Error in getUserByName().", userDao.getUserByName("bente", true));
         assertNull("Error in getUserByName().", userDao.getUserByName("", true));
         assertNull("Error in getUserByName().", userDao.getUserByName(null, true));
--- a/airsonic-main/pom.xml
+++ b/airsonic-main/pom.xml
@@ -223,9 +223,9 @@
         </dependency>
 
         <dependency>
-            <groupId>hsqldb</groupId>
+            <groupId>org.hsqldb</groupId>
             <artifactId>hsqldb</artifactId>
-            <version>1.8.0.7</version>
+            <version>2.5.0</version>
             <scope>runtime</scope>
         </dependency>
 
--- a/airsonic-main/src/main/java/org/airsonic/player/dao/PlayQueueDao.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/dao/PlayQueueDao.java
@@ -35,7 +35,7 @@ import java.util.List;
 @Repository
 public class PlayQueueDao extends AbstractDao {
 
-    private static final String INSERT_COLUMNS = "username, current, position_millis, changed, changed_by";
+    private static final String INSERT_COLUMNS = "username, \"CURRENT\", position_millis, changed, changed_by";
     private static final String QUERY_COLUMNS = "id, " + INSERT_COLUMNS;
     private final RowMapper<SavedPlayQueue> rowMapper = new PlayQueueMapper();
 
--- a/airsonic-main/src/test/java/org/airsonic/player/TestCaseUtils.java
+++ b/airsonic-main/src/test/java/org/airsonic/player/TestCaseUtils.java
@@ -75,7 +75,7 @@ public class TestCaseUtils {
         List<String> tableNames = daoHelper.getJdbcTemplate().queryForList("" +
                       "select table_name " +
                       "from information_schema.system_tables " +
-                      "where table_name not like 'SYSTEM%'"
+                      "where table_type <> 'SYSTEM TABLE'"
               , String.class);
 
         return tableNames.stream()


randomnicode:
This PR is mostly in alignment with the test PR I submitted #847 ages ago that upgraded HSQLDB (which incidentally also included thoughts and commits about Lucene upgrades). I've been using a derivative of that PR for about a year now, without issues.

harawata:
Thank you, @fxthomas !

I just tested the upgrade from a fresh 10.4.1 installation and it seems to be working as expected.
Here is the relevant part of the log:

```
2019-10-19 22:31:33.960  INFO --- o.a.p.dao.LegacyHsqlDaoHelper            : HSQLDB database upgrade needed, from version 1.8.0 to 2.5.
2019-10-19 22:31:33.964  INFO --- o.a.p.dao.LegacyHsqlDaoHelper            : Backing up HSQLDB database to /Users/xxx/airsonic2/db/airsonic.backup.20191019223133...
2019-10-19 22:31:33.969  INFO --- o.a.p.dao.LegacyHsqlDaoHelper            : Copying /Users/xxx/airsonic2/db to /Users/xxx/airsonic2/db/airsonic.backup.20191019223133
2019-10-19 22:31:33.970  INFO --- o.a.p.dao.LegacyHsqlDaoHelper            : Copying /Users/xxx/airsonic2/db/airsonic.properties to /Users/xxx/airsonic2/db/airsonic.backup.20191019223133/airsonic.properties
2019-10-19 22:31:33.972  INFO --- o.a.p.dao.LegacyHsqlDaoHelper            : Copying /Users/xxx/airsonic2/db/airsonic.script to /Users/xxx/airsonic2/db/airsonic.backup.20191019223133/airsonic.script
2019-10-19 22:31:33.976  INFO --- o.a.p.dao.LegacyHsqlDaoHelper            : HSQLDB database backup complete.
2019-10-19 22:31:33.976  INFO --- o.a.p.dao.LegacyHsqlDaoHelper            : Performing database upgrade...
2019-10-19 22:31:34.382  INFO --- o.a.p.dao.LegacyHsqlDaoHelper            : Database connection established. Current version is: 2.5.0.
2019-10-19 22:31:34.382  INFO --- o.a.p.dao.LegacyHsqlDaoHelper            : Shutting down database...
2019-10-19 22:31:34.502  INFO --- o.a.p.dao.LegacyHsqlDaoHelper            : Database connection closed. Current version is: 2.5.0.
2019-10-19 22:31:34.502  INFO --- o.a.p.dao.LegacyHsqlDaoHelper            : Database upgrade complete. Current version is: 2.5.0.
2019-10-19 22:31:34.606  INFO --- o.a.p.service.SettingsService            : Java: 11.0.3, OS: Mac OS X
2019-10-19 22:31:34.856  INFO --- org.airsonic.player.Application          : Detected Tomcat web server
2019-10-19 22:31:39.432  INFO --- o.a.p.service.PodcastService             : Automatic Podcast update scheduled to run every 24 hour(s), starting at Sat Oct 19 22:36:39 JST 2019
2019-10-19 22:31:39.536  INFO --- o.a.p.s.search.IndexManager              : Index directory was created (index version 16). 
2019-10-19 22:31:39.537  INFO --- o.a.p.s.MediaScannerService              : Automatic media library scanning scheduled to run every 1 day(s), starting at Sun Oct 20 03:00:00 JST 2019
2019-10-19 22:31:40.490  INFO --- l.executor.jvm.JdbcExecutor              : SELECT COUNT(*) FROM PUBLIC.DATABASECHANGELOGLOCK
2019-10-19 22:31:40.492  INFO --- l.executor.jvm.JdbcExecutor              : SELECT COUNT(*) FROM PUBLIC.DATABASECHANGELOGLOCK
2019-10-19 22:31:40.500  INFO --- l.l.StandardLockService                  : Successfully acquired change log lock
```

Let me know if you need more info.
And feel free to close #876 .

fxthomas:
@harawata Thanks a lot for the testing!
@randomnicode Ah, thanks as well for the information! I'm really glad to know that there are no issues with running HSQLDB 2.x for an extended period of time.

One thing I'm still uneasy about is that the controlled upgrade procedure requires that nothing opens a connection to the data source while we're doing it (otherwise the database would be in fact upgraded, but without the additional protection of doing `SHUTDOWN SCRIPT`, which we cannot do on the main data source).

During my tests this was the case, but with how it is currently setup this is not in fact guaranteed. I was wondering if there was a better way?

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
```diff
@@ -31,29 +51,226 @@ public void checkpoint() {
...
-    @PreDestroy
-    public void onDestroy() {
-        Connection conn = null;
+    /**
+     * Return the current version of the HSQLDB database, as reported by the database properties file.
+     */
+    private String getHsqldbDatabaseVersion() {
+        Properties prop = new Properties();
+        File configFile = new File(settingsService.getDefaultJDBCPath() + ".properties");
+        try {
```
jvoisin: Please use a try-with-resource construction here ♥

fxthomas: Ah, you're entirely right, thanks for noticing!

harawata:
@fxthomas ,

In my understanding, airsonic uses HSQLDB in `file:` mode and there can be no multiple connections to the database.

If we need to care about users who use HSQLDB in server-client mode, that's a different story.
Is this usage even supported?

fxthomas:
That's definitely out of scope. What I was pointing at was more along the lines of : theoretically our own code could start using the `dataSource` before the migration is done in a controlled way... thereby doing it in an entirely uncontrolled way, without backups or checks.

I'm not sure what logic Spring uses to start dependencies, but the chain here looks like this:

```
dataSource -> daoHelper
           -> liquibase
```

If somehow Spring decides to switch the order, then we're in trouble.

harawata:
The order looks solid to me, but it should be reviewed by someone who has deep knowledge about the internals of airsonic and Spring. :)
Apologies for the noise!

tesshucom:
I have been looking forward to this very much.

However, there are currently several PR related to DB.
Not a bad sign. I think all of them are important.
But happy puzzle.

In reality there will be an order in the application of PR, but do you have consensus among the major members of Airsonic?

#1368, #1363 , and this PR.

If the order is simply shown, the gaze of the committers may be concentrated.

tesshucom:
I have been looking forward to this very much.

However, there are currently several PR related to DB.
Not a bad sign. I think all of them are important.
But happy puzzle.

@fxthomas
In reality there will be an order in the application of PR, but do you have consensus or story among the major members of Airsonic?

#1358, #1368, #1363 , and this PR.

If the order is simply shown, the gaze of the committers may be concentrated.

fxthomas:
@tesshucom A nice puzzle indeed, although I don't mind rebasing this -- it is part of development! :smile: 

I initially set the milestone for this to 11.0, so that the next major release could be the "big upgrade", but this was not really the result of a consensus.

Ideally I'd like for this and Spring Boot 2 to be done on the same release. Since these 2 are relatively risky upgrades it would make sense to get feedback on both at the same time, and finalize the release once we have a stable release candidate.

What do you think of this plan?

tesshucom:
Well, I'm not dissatisfied with any plan. &#x1f609;

What we know is that if a scenario is shown to some extent, there will probably be more committers that will preferentially collaborate on high priority topics.
Otherwise, the interest of the third party will fade.

A series of PR is an engineer's concern and can be useful later.
It will also attract engineers who are used to newer libraries, so ivery benefitable.

But #1340 is a problem for some users, and there are people who are in trouble now.
If anything, I feel #1340 is different category.

randomnicode mentions Hikari.
So if thread pooling improvements are part of muff1nman's plan, It is recommended that you resolve and merge early.
At that time, verification of #1340 is necessary.
(# 1340 should have many partners. In various environments.)

Anyway, by closing what is definitely a high priority, the number of puzzle pieces is reduced.

tesshucom:
Well, I'm not dissatisfied with any plan. &#x1f609;

What we know is that if a scenario is shown to some extent, there will probably be more committers that will preferentially collaborate on high priority topics.
Otherwise, the interest of the third party will fade.

A series of PR is an engineer's concern and can be useful later.
It will also attract engineers who are used to newer libraries, so ivery benefitable.

But #1340 is a problem for some users, and there are people who are in trouble now.
If anything, I feel #1340 is different category.

randomnicode mentions Hikari.
So if thread pooling improvements are part of muff1nman's plan, It is recommended that you resolve and merge early.
At that time, verification of #1340 is necessary.
(#1340 should have many partners. In various environments.)

Or assuming there is a problem with the current implementation, #1340 is declared as a block item.
We can't move on until we fix the bugs together.

Anyway, by closing what is definitely a high priority, the number of puzzle pieces is reduced.

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
randomnicode: please use try-with-resources here?

there is a possible minor resource leak here because the inputstream isn't being closed explicitly

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
```diff
@@ -31,29 +51,226 @@ public void checkpoint() {
...
+        }
+    }
+
+    /**
+     * Perform a backup of the HSQLDB database, to a timestamped directory.
+     * @return the path to the backup directory
+     */
+    private Path performHsqldbDatabaseBackup() throws IOException {
+
+        String timestamp = new SimpleDateFormat("yyyyMMddHHmmss").format(new Date());
```
randomnicode: Maybe use the newer java.time classes here?
```
DateTimeFormatter.ofPattern("xyz").format(Instant.now());
```

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
```diff
@@ -31,29 +51,226 @@ public void checkpoint() {
...
+                Path currentSource = sources.get(i);
+                Path currentDestination = destinations.get(i);
+                if (currentDestination.toFile().isDirectory()) {
+                    currentDestination.toFile().mkdirs();
+                } else {
+                    currentDestination.getParent().toFile().mkdirs();
+                    Files.copy(sources.get(i), destinations.get(i));
+                }
+            }
+        }
```
randomnicode: I think you're just trying to copy a folder over here?
We can replace all this by just using `FileUtils.copyDirectory(source,destination)` ?

randomnicode: Additionally, out of curiosity, why not use the BACKUP db command?
```
BACKUP DATABASE TO <file path> SCRIPT
```
if the file path is a folder, it'll checkpoint, compact, name and gzip a file in the folder. The name will have the date, and the name of the db, which sounds like what is happening here?

one concern with the manual file copying is the possibility of a file lock, which might be held by something else, possibly the db itself when the file copying is done?

randomnicode: I guess I spoke out of context, it seems the shutdown command is issued before a copying is done. Regardless, BACKUP allows the db itself to generate the backup.

randomnicode: PS, I have no idea if BACKUP is an option in 1.8. Maybe check first?

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
```diff
@@ -31,29 +51,226 @@ public void checkpoint() {
...
+        LOG.info("Database connection closed. Current version is: {}.", getHsqldbDatabaseVersion());
+    }
+
+    /**
+     * Checks if the database is correctly shut down
+     */
+    private boolean isHsqldbDatabaseCorrectlyShutdown() {
+
+        // The db/airsonic.data file may or may not exist (contains CACHED tables).
+        // The db/airsonic.backup file may or may not exist (not used during normal operation).
```
randomnicode: Is this true after `SHUTDOWN SCRIPT`?
I think only the script and the properties file remains.

In fact, I thought this is how SHUTDOWN SCRIPT is different from SHUTDOWN COMPACT?
The former compresses down to script files, the latter additionally generates the backup and data files?

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
```diff
@@ -31,29 +51,226 @@ public void checkpoint() {
...
+    }
+
+    private void upgradeHsqldbDatabaseSafely() {
+        try {
+            // Check that it is safe to do the upgrade
+            if (!isHsqldbDatabaseCorrectlyShutdown()) {
+                LOG.warn(
+                    "Database has not been shut down properly. " +
+                    "Airsonic should be stopped properly before attempting to upgrade the database.");
+            }
```
randomnicode: hmmm, i don't understand this logic.
- the bean's instantiated
- postconstruct checks if upgrade is needed, then calls this method
- why should the db have been shutdown at this point? didn't the application just start?
- wouldn't this generate a warn for everyone?

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
```diff
@@ -31,29 +51,226 @@ public void checkpoint() {
...
+            // On upgrade, the official documentation recommends that we
+            // run 'SHUTDOWN SCRIPT' to compact all the database into a
+            // single SQL file.
+            //
+            // In practice, if we don't do that, we did not observe issues
+            // immediately but after the upgrade.
+            LOG.info("Shutting down database...");
+            try (Statement st = conn.createStatement()) {
+                st.execute("SHUTDOWN SCRIPT");
+            }
```
randomnicode: shouldn't we just use the datasource jdbctemplate injected into this class?
```
getJdbcTemplate().execute("SHUTDOWN SCRIPT");
```

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
```diff
@@ -31,29 +51,226 @@ public void checkpoint() {
...
+        LOG.info("HSQLDB database backup complete.");
+        return destination;
+    }
+
+    /**
+     * Perform an in-place database upgrade from HSQLDB 1.x to 2.x.
+     */
+    private void performHsqldbDatabaseUpgrade() throws SQLException {
+        // This will upgrade HSQLDB from 1.8 to 2.5 on the first connection.
+        try (Connection conn = DriverManager.getConnection(settingsService.getDefaultJDBCUrl())) {
```
randomnicode: What i'm most curious about is how this is able to establish a connection without a username/password. That jdbc url is protected by a username at least, right?

I would prefer to be safe and replace all this with a `getJdbcTemplate.execute()`

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
randomnicode: I would actually prefer to add another statement immediately afterwards to check if the connection is reestablished after shutdown

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
```diff
@@ -31,29 +51,226 @@ public void checkpoint() {
...
+            //
+            // In practice, if we don't do that, we did not observe issues
+            // immediately but after the upgrade.
+            LOG.info("Shutting down database...");
+            try (Statement st = conn.createStatement()) {
+                st.execute("SHUTDOWN SCRIPT");
+            }
+        }
+
+        LOG.info("Database connection closed. Current version is: {}.", getHsqldbDatabaseVersion());
```
randomnicode: I'm hesitant about the overreliance of this method. It is disk IO bound and slow. Must we log it here when we've logged it a hundred times elsewhere? maybe we could actually use the check above and log that version?

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
```diff
@@ -31,29 +51,226 @@ public void checkpoint() {
...
+        try {
+            // Check that it is safe to do the upgrade
+            if (!isHsqldbDatabaseCorrectlyShutdown()) {
+                LOG.warn(
+                    "Database has not been shut down properly. " +
+                    "Airsonic should be stopped properly before attempting to upgrade the database.");
+            }
+
+            // Perform the upgrade
+            Path backupPath = performHsqldbDatabaseBackup();
```
randomnicode: I think this might fail if the database's lock file is being held

airsonic-main/src/test/java/org/airsonic/player/api/jukebox/AbstractAirsonicRestApiJukeboxIntTest.java
```diff
@@ -144,6 +135,12 @@ public void setup() throws Exception {
         assertThat(testJukeboxPlayer.getPlayQueue().size()).isEqualTo(2);
     }
 
+    @After
+    public void cleanDataBase() {
+        daoHelper.getJdbcTemplate().execute("DROP SCHEMA PUBLIC CASCADE");
+        dataBasePopulated = false;
+    }
```
randomnicode: Any idea why this is part of the PR?

It'll slow down the tests... Was it polluting the db for the next test? Aren't the tests just start and stop jukebox?

airsonic-main/src/test/java/org/airsonic/player/dao/UserDaoTestCase.java
```diff
@@ -120,7 +120,7 @@ public void testGetUserByName() {
         assertUserEquals(user, newUser);
 
         assertNull("Error in getUserByName().", userDao.getUserByName("sindre2", true));
-        assertNull("Error in getUserByName().", userDao.getUserByName("sindre ", true));
+        assertNotNull("Error in getUserByName().", userDao.getUserByName("sindre ", true));
```
randomnicode: vaguely remember fixing this test... I'm not sure why this is behaving differently

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
```diff
@@ -18,6 +35,9 @@
 
     private static final Logger LOG = LoggerFactory.getLogger(LegacyHsqlDaoHelper.class);
 
+    @Autowired
```
randomnicode: This uses field injection.

Maybe consider putting this on the setter so a method injector is used?
Or just remove the getter/setter altogether?

randomnicode:
Sorry for commenting on this pr a bit late. I've commented on the relevant code sections.

Aside from that, here are my general thoughts on the PR:

- I don't think there are any concerns regarding using the datasource while migrations are happening. Aside from liquibase, the AbstractDAO is the parent class for all DAOs, and since AbstractDAO needs the daoHelper injected, no Dao bean may be instantiated while the daoHelper is still going through its instantiation. The only concern here would be some class injecting/autowiring the JDBCTemplate, NamedJdbcTemplate or DataSource directly without injecting DaoHelper, and you can easily check for references to those classes and see where they end up. Liquibase, I can see has already been made dependent on daoHelper.
- A concern I _do_ have is regarding the timing of the backup copy operation. Your threadpool might already have instantiated a connection to the db by the time the daoHelper is going through its postconstruct and be in the process of upgrading it when we start to copy over files. We're probably just getting lucky, I suspect DBCP2's initial pool size is 0, and spring isn't overriding it. Other threadpools might not be so forgiving with defaults, or you might get a user who's overriding the initial pool size in his properties. Additionally, your db is live while you're copying files over. That might be an issue. So, the best place to do this copy, in my opinion, is probably in the Application prior to running the Application. That ensures the DB pool hasn't been instantiated.

- I do think the one particular class (LegacyHsqlHelperDao) is too extrapolated. In my opinion, most of these checks are too drawn out. All we're looking to do is, if there is an upgrade, make a backup copy before it and execute a (shutdown) command after it. Everything else is just logging or fluff. We can't stop the upgrade, we're just looking to save the state before it and consolidate after it:
  - if db is hsqldb and db version mismatch with driver
    - make a backup copy
    - shutdown db

That's about it. All the other checks about presence or absence of files are redundant in my opinion, and it's not like we do anything anyway about them.  The connection pool will establish anyway no matter what we return from those checks. Just get out of the way and let the driver do it's thing (which we do anyway at the end of the day). 

So, in short, I would restructure the PR:
- In Application: Read the properties and the db driver and copy files if necessary, and set an env variable to do a db shutdown later. Fun fact, you'll have to figure out what profile you're running to see what url to use, because it might not be legacy... you might just decide to copy and backup anyway if there's a file at that location, whether its a used db or not...
```
public class Application {
...
private static void doConfigure(...) {
  ...
  if (getDBDriver().equals("HSQLDB") && getDriverVersion() != getDBVersion()) {
    logger.warn("backing up db...");
    try {
       FileUtils.copyDirectory(getDBLocation(), getDBBackupLocation());
    } catch (Exception e) { logger.warn("couldn't backup db", e); }
    System.setProperty("shutdownDbAtStart", "shutdown");
    logger.warn("done with db backup...");
  }
```

- In LegacyHSqlDaoHelper: In postconstruct, just shutdown the db if env variable is set
```
if ( StringUtils.equals(System.getProperty("shutdownDbAtStart"), "shutdown") {
  getJdbcTemplate().execute("SHUTDOWN SCRIPT");
  getJdbcTemplate().query*(get db version);
  //log db version
}
```
That should be all you need.
Hope this helps?

randomnicode:
One more note, regarding landing the PRs:
- This PR is definitely a non-backwards compatible breaking change. It will need a major revision change.
- The rest of the PRs quoted above by @tesshucom are not breaking changes. They're significant infrastructural changes for sure, but none of them are breaking API changes. By semver, they don't need a major version increment. I would probably land them and increment maybe the minor version for enhanced functionality, but definitely not the major version.
- As such, the order matters. I would land the minor semver changes, and then do the switch to the major upgrade. As such, people in the previous release who don't want to upgrade to the next major revision can also enjoy the benefits of an upgraded platform without worrying about backwards-compability issues.

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
tesshucom: If  check for unexpected startups other than embedded mode, not harmful logic.
It may be unnecessary if the case can be avoid by the framework locking mechanism.
The check logic that can be useful against the legacy framework implementation of hsql is implemented...?

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
tesshucom: Airsonic's getter / setter injections are mostly bad habits and can be removed.
On the other hand partially, there are places where getters / setters are used for delayed injection.
(Without using @Lazy)

Because there is no advantage, I agree with the opinion of eliminating all getter / setter injections.

randomnicode: Don't think getter/setters are bad habits, in fact, I think they're actually recommended over field injection (which is considered harmful). Regardless, discussions over code purity aside, we should pick one and either put the Autowired over the setter if we're putting a setter out there, or remove the setter if we're doing field injection.

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
randomnicode: Hmmm, not sure I understand. What other startups other than embedded mode?

While it's not harmful logic, I agree, because it's just issuing warnings, it (a) slows down startup due to file checks, and (b) is an unnecessary check: this condition always evaluates to true. We don't even need to check.

It always is true because, unless there was manual intervention involved, the database isn't shutdown, there is no logic prior to this I see that shuts it down in the code, and there is no reasonable expectation for it to have been shut down automatically. In fact, since the application is starting, the db should be starting up and I'd be worried if it _was_ shutdown.

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
```diff
@@ -31,29 +51,226 @@ public void checkpoint() {
...
+        // The db/airsonic.log file should not exist (contains db logs that are not supported in the upgrade).
+        if (new File(settingsService.getDefaultJDBCPath() + ".log").exists()) {
+            LOG.warn("The db/airsonic.log file exists. The database may not be shut down correctly.");
+            return false;
+        }
+        // The db/airsonic.lck file should not exist (lock file, indicates another running process or unclean shutdown).
+        if (new File(settingsService.getDefaultJDBCPath() + ".lck").exists()) {
+            LOG.warn("The db/airsonic.lck file exists. The database may not be shut down correctly.");
+            return false;
+        }
```
randomnicode: I don't think any of these checks are needed. All they seem to be doing is logging about the state of the db, which I don't see how it is useful to us, or even a concern, considering there isn't anything we're doing about it. It's just slowing up startup time for the sake of a few warns.

Consider this:
 - DB is corrupt. we issue a warn, and then what? we upgrade the db
 - DB wasn't shutdown. we issue a warn and then what? we upgrade the db
 - DB is fine. we don't issue anything, and then what? we upgrade the db

Everything leads to the same outcome. I would consider them useful if they were altering the decision to upgrade (to follow the decision tree), but they're not. 

I think DB monitoring is a valuable tool, but a different issue. It's probably good to know your DB is corrupt or wasn't shutdown properly, but that's an issue to be taken care of regardless of whether an upgrade is happening (using monitoring tools).

Why does the user need to know the status of his db at the time of the upgrade and not otherwise?
If it was corrupt, and we weren't upgrading, would we have issued a warn? (we don't)

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
tesshucom: The reason why the Spring team deprecates field injection is because it recommends constructor injection.
(setter / setter allows the most ambiguous description for circular references.)
I think it's okay to turn off either one here.

randomnicode: Hmm, actually, Spring promotes both constructor and setter injection based on mandatory or optional nature of dependencies respectively. It's just field injection it doesn't recommend (though it's not deprecated)

Field injections are reflection-based and have issues with immutable fields along with some more semantic problems.

In any case, it seems we're in agreement here, to remove one of the two.

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
fxthomas: Good catch, thanks!

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
fxthomas: Thanks for the tip, I'll change that!

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
fxthomas: The big issue about using `BACKUP` at this point is that, when you're connected to the db, it's already too late, the database has been upgraded. And I looked everywhere, but I haven't found a way to make HSQLDB do upgrades other than actually connecting to the database...

fxthomas: On the other hand I never knew about `FileUtils.copyDirectory` (why don't these SO answers mention that instead of tens of lines of code if it exists?). Thanks for mentioning!

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
fxthomas: After `SHUTDOWN SCRIPT` only script and properties remain, you're right. But I'm also calling this before the upgrade to check the assumptions I've made on the state of the db, so it needs to handle un-upgraded databases too.

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
fxthomas: The condition doesn't always evaluates to true -- in fact in 99% cases at this point in the startup sequence it should be false. This bean should start before any other database operation has been run, so at this point the db isn't started yet and should be in shut down state.

The scenario I want to check is that for some reason the previous Airsonic version (before the upgrade) is not shut down properly (either power loss, timeout that kills the database, etc.). This leaves a log file behind, which is not supported at all by the upgrade process (this means that such users _will_ lose data).

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
fxthomas: We cannot actually use the `DataSource` that Airsonic is going to use after the upgrade, because it will be closed. I did not find any way to restore the connection after a `SHUTDOWN`, do you have an idea?

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
fxthomas: In this case the username/password are embedded in the JDBC URL.

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
fxthomas: You mean `getHsqldbDatabaseVersion`? It's only once, so even if it is disk IO bound it's not going to impact a lot of things. And I would rather log too much during the upgrade than not enough.

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
fxthomas: You're right. A big assumption of `performHsqldbDatabaseBackup` is that you are not running it in parallel with any other connection.

airsonic-main/src/test/java/org/airsonic/player/api/jukebox/AbstractAirsonicRestApiJukeboxIntTest.java
fxthomas: I am not really satisfied with this, but this was the only way I could find to actually run these tests without running into issues of the db being polluted. I did not have too much time to investigate, but do you see better ways?

airsonic-main/src/test/java/org/airsonic/player/dao/UserDaoTestCase.java
fxthomas: See ae4d49f59a0b7f10d11648699391888993ad8e37.

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
fxthomas: Which one do you guys think is better? I actually don't see a lot of issues with using field injection like this in Airsonic, it usually works well.

(Also, we should probably do this in another PR if we want to change the standard everywhere in Airsonic.)

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
fxthomas: The presence of a `.log` file at least should be checked and logged, because it will lose data during the upgrade process. I'll look at reducing these checks.

fxthomas:
So, aside from PR comments:

> A concern I do have is regarding the timing of the backup copy operation.

Me too.

> So, the best place to do this copy, in my opinion, is probably in the Application prior to running the Application.

That may actually be a good idea. I'm going to restructure the PR in order to do this.

> In my opinion, most of these checks are too drawn out.

I don't trust the HSQLDB upgrade process, so I'd rather check my assumptions at each step of the process. The checks are not there for the sole purpose of checking, they're here to warn users that something may have been wrong, instead of finding it out much later after they removed the backup.

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
randomnicode: Spring would recommend setter/constructor injection
I personally think it's better to keep it consistent, and change it with everything else if we choose to, so field injection.

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
randomnicode: I don't see it? Where? 
`settingsService.getDefaultJDBCUrl()` points to https://github.com/airsonic/airsonic/pull/1345/files#diff-0291e6bf8d7243eb6dd454bd00f440dfR295
I don't see a user/password in that url

fxthomas: Eh, you're actually right, I haven't revisited this since I wrote the code and I spoke too fast.

Not sure how it works, but the username/password used by Airsonic is probably the default in HSQLDB if you don't specify anything? In this case I agree, it makes sense to specify it explicitly. I'll write something to that effect.

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
randomnicode: Hmmm, first, I'm not convinced it'll lose data if there is a log file. What makes you say that? Compaction isn't a requirement for an upgrade, is it? Is there some documentation that says you'll lose data? I mostly upgrade without compaction, and I don't think I've ever lost data due to an upgrade. It's usually _helpful_ and therefore _recommended_ to have a compact database when upgrading, but why is there fear of data loss?

Second, this method is called twice (before and after the upgrade). Each time you're not affecting the course of the decision-making.

Okay, let's say there is a log file, and let's say there is data loss involved if there is one. So what? What are we going to do by checking for it?
It's not like we're going to be able to stop the upgrade from happening.
Either way this upgrade is happening _no matter what_, right?

So what does checking for a db log file accomplish? What is the purpose? Just to log it to a logger?
Logging does what? Are the users going to be able to recover the data from the log file that they wouldn't otherwise if we didn't have that check or didn't log it? That's certainly not true.

This seems to be a case of logging creep. I really don't think this entire method is needed. Log files or no log files, however or whatever state the db is in, it's getting upgraded, there is simply no choice. Checking for presence or absence of files, certainly in my opinion, doesn't accomplish anything. If the driver doesn't find certain files, it'll throw an exception itself. Us logging about it beforehand and then feeding it to the driver anyway just adds another redundant check.

fxthomas: The whole point of this, in my opinion, was to help figuring out what happened during a borked upgrade. It's mostly for forensics, and will not help at all in recovering the data. I'll remove them then.

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
randomnicode: I don't think that's true at all. There are three issues here:

- > at this point the db isn't started yet and should be in shut down state.

You're thinking that the db is already shutdown when you start an application. While that may be true of recent versions (which I suspect you're using for testing), _all_ versions of subsonic/libresonic/airsonic prior to recent times _didn't_ shutdown the DB at termination, to my recall (termination of threads to the DB doesn't shut it down unless you specify in the connection parameters). Every person jumping from prior to whenever the shutdown hook was introduced (which was just this year/recently?) will have this evaluation true. And so will everyone who migrates from subsonic to airsonic. And so will every person who just shuts down their JVM without a proper SIGTERM signal (which is most cloud deployments, anyone who issues a `docker kill` to stop their container etc). So, by definition, you're only counting an extremely narrow subset of people who shutdown their embedded hsql dbs properly last time (that is, only people who have the most latest version of airsonic, not using docker kills), and excluding most of them who have no reasonable expectation of a shutdown db. And even for that narrow segment:

- > This bean should start before any other database operation has been run 

You can't make that assertion. There are no guarantees when the connection pool instantiates a connection to the database. That's not under user control. While you may control user-defined DB accesses, connections pools might eagerly establish connections just to check if they're alive etc _prior_ to a user command execution. You're getting lucky that your connection pool isn't establishing a connection to the DBs prior to this check, and should not rely on that behavior.

Both the issues above highlight that minus a very specific subset of installs and ideal situations based on luck or undocumented pool behaviors, the DB won't be shutdown and shouldn't be expected to. And even then, it doesn't matter because:

- > The scenario I want to check is that for some reason the previous Airsonic version (before the upgrade) is not shut down properly (either power loss, timeout that kills the database, etc.). This leaves a log file behind, which is not supported at all by the upgrade process (this means that such users will lose data).

Your check here won't stop that upgrade, I'm not sure what you mean by "not supported". Arguably _nothing_ you do here will stop an upgrade (unless you throw exceptions preventing the application from starting up).

Let's say this condition evaluates to false, and we skip this conditional. Your DB is still getting upgraded at next SQL instruction.

We're just trying to make a backup. You're not actually doing the upgrade, that's done by the driver automatically, there's no safe or unsafe. That's happening no matter what.

fxthomas: > There are no guarantees when the connection pool instantiates a connection to the database.

That's, I think, exactly my point when I said, earlier, "one thing I'm still uneasy about is that the controlled upgrade procedure requires that nothing opens a connection to the data source while we're doing it".

> So the only thing we can do is:
>    Make a backup prior to first connection (do checking of db and driver version match to see if its needed)
>    Connect (for upgrade)
>    Compact db

I'll update the PR and do that.

fxthomas:
Also: once this is released, how about recommending that people upgrade to the latest Airsonic version in the 10.x series and start it at least once before upgrading to this?

airsonic-main/src/test/java/org/airsonic/player/api/jukebox/AbstractAirsonicRestApiJukeboxIntTest.java
randomnicode: Not sure why your db is getting polluted between the tests, but I don't see anything in the tests that is writing to the db. They all seem to be read operations. They just start and stop jukeboxes for two different jukeboxes. I don't think this should be part of the PR.

airsonic-main/src/test/java/org/airsonic/player/dao/UserDaoTestCase.java
randomnicode: Yep, and https://github.com/airsonic/airsonic/pull/847/files#diff-979b441261c64e3f4523f0e4fe90711dR118

airsonic-main/src/test/java/org/airsonic/player/api/jukebox/AbstractAirsonicRestApiJukeboxIntTest.java
fxthomas: There is more info in the original PR in https://github.com/airsonic/airsonic/pull/876. I'll probably try to get it working at some point, but this'll take more digging.

airsonic-main/src/test/java/org/airsonic/player/dao/PodcastDaoTestCase.java
```diff
@@ -134,10 +134,10 @@ public void testGetEpisodes() {
...
         List<PodcastEpisode> episodes = podcastDao.getEpisodes(channelId);
         assertEquals("Error in getEpisodes().", 4, episodes.size());
-        assertEpisodeEquals(a, episodes.get(0));
-        assertEpisodeEquals(c, episodes.get(1));
-        assertEpisodeEquals(b, episodes.get(2));
-        assertEpisodeEquals(d, episodes.get(3));
+        assertEpisodeEquals(d, episodes.get(0));
+        assertEpisodeEquals(a, episodes.get(1));
+        assertEpisodeEquals(c, episodes.get(2));
+        assertEpisodeEquals(b, episodes.get(3));
```
randomnicode: Hmmm, is this because null values are being treated higher order than populated now in the new db?

This will actually be a nuisance for people with podcasts, where all the null values will crowd the top instead of getting the newly imported ones first. People will never see the fresh podcasts!

I suggest we add `sql.nulls_first=false` to the jdbc connection String to retain the old behavior

so your connection string will be something like: `"return jdbc:hsqldb:file:" + getAirsonicHome().getPath() + "/db/" + getFileSystemAppName() + ";sql.enforce_size=false;sql.nulls_first=false";`

Documented here: http://hsqldb.org/doc/2.0/guide/dbproperties-chapt.html

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
randomnicode: Hmmm, then how do we expect the application to start up after the shutdown? Do people have to start the application twice?
One idea is to shutdown the connection pool and then restart it...

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
randomnicode: just use the JdbcTemplate, it has the user/pass from the source inside it. No need to custom write anything else

`getJdbcTemplate().execute("SHUTDOWN SCRIPT");'

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
```diff
@@ -31,29 +47,162 @@ public void checkpoint() {
...
-            // Properly shutdown the embedded HSQLDB database.
             LOG.debug("Database shutdown in progress...");
             JdbcTemplate jdbcTemplate = getJdbcTemplate();
-            conn = DataSourceUtils.getConnection(jdbcTemplate.getDataSource());
-            conn.setAutoCommit(true);
-            jdbcTemplate.execute("SHUTDOWN");
+            try (Connection conn = DataSourceUtils.getConnection(jdbcTemplate.getDataSource())) {
+                conn.setAutoCommit(true);
+                jdbcTemplate.execute("SHUTDOWN");
+            }
```
randomnicode: Not necessary to set autocommit. This isn't in a transaction, and moreover is an atomic command.
```
getJdbcTemplate().execute("SHUTDOWN");
```

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
```diff
@@ -31,29 +47,162 @@ public void checkpoint() {
...
+    /**
+     * Return the paths to all existing database files.
+     */
+    private String[] getHsqldbDatabaseFiles() {
+        String[] extensions = { ".properties", ".script", ".log", ".data", ".backup", ".lck" };
+        return Arrays.stream(extensions)
+                .map(extension -> settingsService.getDefaultJDBCPath() + extension)
+                .filter(path -> new File(path).exists())
+                .toArray(size -> new String[size]);
+    }
```
randomnicode: not needed?

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
```diff
@@ -31,29 +47,162 @@ public void checkpoint() {
...
+        }
+
+        // Check the database driver version
+        String driverVersion = null;
+        try {
+            Driver driver = DriverManager.getDriver(settingsService.getDefaultJDBCUrl());
+            driverVersion = String.format("%d.%d", driver.getMajorVersion(), driver.getMinorVersion());
+            if (driver.getMajorVersion() != 2) {
+                LOG.warn("HSQLDB database driver version {} is untested. Trying to connect anyway, this may upgrade the database from version {}.", driverVersion, currentVersion);
+                return false;
```
randomnicode: I think we should set this true?
This setting determines whether a backup gets made.
So, in my opinion, any scenario where we're unsure of what's about to happen, we should make a backup, right?

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
```diff
@@ -31,29 +47,162 @@ public void checkpoint() {
...
+        } else if (currentVersion.startsWith("2.")) {
+            // If the database version is 2.x but older than the driver, the upgrade should be relatively painless.
+            LOG.info("HSQLDB database will be silently upgraded from version {} to {}.", currentVersion, driverVersion);
+            return false;
+        } else {
+            // If this happens we're on a completely untested version and we don't know what will happen.
+            LOG.warn("HSQLDB database upgrade from version {} to version {} is untested. Trying to connect anyway, this may upgrade the database.", currentVersion, driverVersion);
+            return false;
+        }
+    }
```
randomnicode: Here's the decision matrix I think we should do, to be checked in order (`-` means whatever):

DB | Driver | likely reason | decision |
---|---|---|---|
null | - | new db | false |
\- | null or !2 | something went wrong, we better make copies | true |
1.x | 2.x | yeah, this is the upgrade | true |
2.x | 2.x | already up to date | false |

all else true

airsonic-main/src/test/java/org/airsonic/player/api/jukebox/AbstractAirsonicRestApiJukeboxIntTest.java
randomnicode: hmm, that doesn't look like a test pollution. That looks like just the db shutdown for some reason. Don't know between which tests (in the same suite or the different suite)

Try removing line 41 so it stops trying to dirty the context cache (so it doesn't shut down) and see if that works?

airsonic-main/src/main/resources/applicationContext-db-legacy.xml
```diff
@@ -5,7 +5,7 @@
        profile="legacy">
 
     <bean id="dataSource"
-          class="org.springframework.jdbc.datasource.DriverManagerDataSource">
+          class="org.apache.commons.dbcp2.BasicDataSource">
         <property name="driverClassName" value="org.hsqldb.jdbcDriver" />
         <property name="url"
                   value="#{T(org.airsonic.player.service.SettingsService).defaultJDBCUrl}" />
```
randomnicode: 👍  We're getting closer to removing this separate LegacyDaoHelper class :)

Once this is in a connection pool, then there's no point to keep a separate class

airsonic-main/src/main/java/org/airsonic/player/service/SettingsService.java
```diff
@@ -287,8 +287,12 @@ private static String getFileSystemAppName() {
         return home.contains("libresonic") ? "libresonic" : "airsonic";
     }
 
+    public static String getDefaultJDBCPath() {
+        return getAirsonicHome().getPath() + "/db/" + getFileSystemAppName();
+    }
+
     public static String getDefaultJDBCUrl() {
-        return "jdbc:hsqldb:file:" + getAirsonicHome().getPath() + "/db/" + getFileSystemAppName();
+        return "jdbc:hsqldb:file:" + getDefaultJDBCPath() + ";sql.enforce_size=false";
```
randomnicode: add: `;sql.nulls_first` (see discussion below)

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
```diff
@@ -31,29 +47,162 @@ public void checkpoint() {
...
+                st.execute("SHUTDOWN SCRIPT");
+            }
+        }
+
+        LOG.info("Finished HSQLDB database upgrade. Current version is: {}.", getHsqldbDatabaseVersion());
+    }
+
+    private void upgradeHsqldbDatabaseSafely() {
+        try {
+            performHsqldbDatabaseBackup();
```
randomnicode: Let's do this backup in Application

randomnicode:
> Also: once this is released, how about recommending that people upgrade to the latest Airsonic version in the 10.x series and start it at least once before upgrading to this?

I think that's a good idea, but I'm not sure anyone will heed it (speaking from personal experience). Most people won't even read it. They'll go new major version? Yay, lets update. And even those who read it, probably won't be bothered to first look for and download the latest 10.x release, then run it, then shut it down, then look for and download the latest 11.x release, then run it. I think the inertia and laziness factor coupled with too many steps will be too high.

Might still be worth just putting it in though. Just don't think we should rely on it.

PS We should start merging the big PRs on master soon so this can be ready to go without delay

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
tesshucom: It may be better to use a setter here (often used now).

If the standard is changed in the future, the setter can be stopped and rewritten as a constructor.
In Airsonic, most of the injection fields can be encapsulated, with some exceptions.
(Except when adopting special design patterns, there is no necessity to publish as public. )

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
fxthomas: You're probably right re:autocommit. I'll remove that.

On the other hand, running `SHUTDOWN` with the JDBCTemplate _permanently_ closes the connection, and all connections after that will fail with `SQLException` (I tried...). I don't know how to entirely reload the pool, do you have an idea?

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
fxthomas: Yup, removing that.

airsonic-main/src/main/resources/applicationContext-db-legacy.xml
fxthomas: ...as soon as we figure out a way to actually run `SHUTDOWN` when Airsonic closes.

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
fxthomas: Sounds logical, let's do that.

airsonic-main/src/test/java/org/airsonic/player/dao/PodcastDaoTestCase.java
fxthomas: I don't mind doing that, but will other databases (PostgreSQL/MariaDB) do the same as HSQL? Shouldn't this be handled in another layer in this case?

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
fxthomas: Done.

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
fxthomas: Done.

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
fxthomas: Done.

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
fxthomas: Removed.

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
fxthomas: Fixed by moving this to `Application`.

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
fxthomas: Done.

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
fxthomas: Fixed by explicitly setting username/password.

fxthomas:
I moved the backup/upgrade logic to Application. You're right, it's much better there than where it was previously, and now I am sure that no bean will try to use the database while we're backing it up ;)

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
fxthomas: I actually removed `settingsService` entirely, it's not needed anymore since the logic has been moved to `Application` startup.

fxthomas:
(Rebased on `master`)

airsonic-main/src/test/java/org/airsonic/player/dao/PodcastDaoTestCase.java
randomnicode: You're not using Postgres for the default db. This string is exclusively for our internal db and is thus specific to hsql db. It's in the string itself `return "jdbc:hsqldb:file:"`. I think we should keep it consistent with what the original hsql db was doing.

People with Postgres can provide their own connection parameters for when they connect to Postgres, that's not under our control.

As far as the general behavior of nulls, yes we should be using NULLS LAST in the order by clauses for our queries. But that's probably a separate PR. For now, I think we should keep our embedded DB consistent with what it was doing before, and keep behavior in line with what we expect to be the best (nulls last)

randomnicode:
Lots of rebases around here :)
You're lucky no one stacked a PR on top of this one. There'd be a ton of merge conflicts.

Will review this in a bit.

airsonic-main/src/main/java/org/airsonic/player/dao/LegacyHsqlDaoHelper.java
```diff
@@ -31,29 +31,25 @@ public void checkpoint() {
...
-    public void onDestroy() {
-        Connection conn = null;
+    /**
+     * Shutdown the embedded HSQLDB database. After this has run, the database cannot be accessed again from the same DataSource.
+     */
+    private void shutdownHsqldbDatabase() {
         try {
-            // Properly shutdown the embedded HSQLDB database.
             LOG.debug("Database shutdown in progress...");
             JdbcTemplate jdbcTemplate = getJdbcTemplate();
```
harawata: Note that this conflicts with #1367 .
Recent version of Spring Boot runs Liquibase migration when injecting JdbcTemplate.
https://github.com/spring-projects/spring-boot/commit/48819253eb5171b93ede2b630fdb8e64ca845e62

harawata: Sorry, this is a shutdown, so it should be fine.

fxthomas:
Rebased on master.

muff1nman:
@fxthomas is it time to remove the WIP: flag?

fxthomas:
Yup, good idea!

airsonic-main/src/main/java/org/airsonic/player/spring/DatabaseConfiguration.java
```diff
@@ -85,6 +86,7 @@ public String userTableQuote(@Value("${DatabaseUsertableQuote:}") String value)
     }
 
     @Bean
+    @DependsOn("legacyDaoHelper")
```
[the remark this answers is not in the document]
fxthomas: Good catch, I think it's unneeded (and an artifact from my earlier explorations of this).

airsonic-main/src/main/java/org/airsonic/player/util/LegacyHsqlUtil.java
```diff
@@ -0,0 +1,167 @@
...
+
+    /**
+     * Perform a backup of the HSQLDB database, to a timestamped directory.
+     * @return the path to the backup directory
+     */
+    public static Path performHsqldbDatabaseBackup() throws IOException {
+
+        String timestamp = DateTimeFormatter.ofPattern("yyyyMMddHHmmss").format(LocalDateTime.now());
+        Path source = Paths.get(SettingsService.getDefaultJDBCPath()).getParent();
+        Path destination = Paths.get(String.format("%s.backup.%s", SettingsService.getDefaultJDBCPath(), timestamp));
```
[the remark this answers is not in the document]
fxthomas: Good catch again, thank you!

airsonic-main/src/main/java/org/airsonic/player/Application.java
```diff
@@ -162,6 +165,11 @@ public Filter noCacheFilter() {
     }
 
     private static SpringApplicationBuilder doConfigure(SpringApplicationBuilder application) {
+        // Handle HSQLDB database upgrades from 1.8 to 2.x before any beans are started.
+        application.application().addListeners((ApplicationListener<ApplicationEnvironmentPreparedEvent>) event -> {
+            LegacyHsqlUtil.upgradeHsqldbDatabaseSafely();
+        });
```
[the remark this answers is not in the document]
fxthomas: Wouldn't users on a servlet container (Tomcat) miss this if this is not put inside a listener (such as in the main `main` method)?

(Also, this is minor but I like my logs to honor settings and be consistent, and putting this in the `main` method prevented logs from appearing as configured by Spring.)

Good idea about using the profile though, I'll add that.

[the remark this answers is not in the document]
fxthomas: ...shoot, it's not available in Spring Boot 1.5. Well, let's add a todo.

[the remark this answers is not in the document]
randomnicode: normally yes, but the method isn't just called from `main()`

`doConfigure()` is also called here: https://github.com/airsonic/airsonic/pull/1345/files#diff-f4ea9b91519ec33fc48ca468f4d855fdR180-R182 which i think will ensure that the servlet container runs it. Best to try compiling and running it on a container, and see if stuff backs up

randomnicode:
Additionally:
Now that you've come so far, there is very little stopping you from actually removing the legacy profile altogether, making the "legacy" and "embed" (a bit of a misnomer) actually the same thing.
In short, the logic for the entire app becomes:
If a jdbc url is provided, use that, otherwise use the default built-in url.

All other distinctions have fallen away, and there's just a whole lot of redundant code duplicating the same stuff.
- There is no distinction in db connection pools (both types use it)
- you don't need to do custom checkpoints for legacy (handled by the db itself, or encode in the default url)
- you dont need to do a custom SHUTDOWN for the legacy (you can encode it in the default jdbc url itself via `shutdown=true`)

Both embed and legacy now mean the same thing.

I also think we should rename those parameters.

airsonic-main/src/main/java/org/airsonic/player/Application.java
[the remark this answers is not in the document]
fxthomas: Hum, I think I will just leave it as is. Putting it on application initialization seems like a natural place to me, with the added benefit that all the low-level Spring machinery is already initialized (so, less chance of being in a weird state).

airsonic-main/src/main/java/org/airsonic/player/util/LegacyHsqlUtil.java
```diff
@@ -0,0 +1,167 @@
...
+            LOG.debug("HSQLDB database doesn't exist, cannot determine version");
+            return null;
+        }
+        try (InputStream stream = new FileInputStream(configFile)) {
+            prop.load(stream);
+            return prop.getProperty("version");
+        } catch (IOException e) {
+            LOG.error("Failed to determine HSQLDB database version", e);
+            return null;
+        }
```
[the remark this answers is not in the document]
fxthomas: Thanks, didn't know that!

fxthomas:
Good ideas, but we need to discuss this in more detail and run some tests. I created a spin-off issue with your comment so that we don't forget about it once the migration is done.

airsonic-main/src/main/java/org/airsonic/player/Application.java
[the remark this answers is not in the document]
randomnicode: Okay, just note here then that at this point, for this event, the application has started but none of the initializers have run (including and especially CustomPropertySourceConfigurer), so no properties from airsonic or profiles can be read.

Also, without any checks, this means that you will touch and back up a db at the default location even if it's not being used. Personally, I would expect a program not to touch a location if I'm specifying a completely different location. For instance, this scenario will occur:

- Person1 specifies a url to connect at /var/myactualdb ("embed" i guess)
- Airsonic starts and updates /var/db instead and makes backup even though as a user I shouldn't really be expecting Airsonic to be touching that location
- (additionally it also connects and updates /var/myactualdb, but makes no backups of it)

randomnicode:
One more point in the context of this PR is regarding this migration: https://github.com/fxthomas/airsonic/blob/upgrade-hsqldb-to-2.x/airsonic-main/src/main/resources/liquibase/10.2/setup-hsqldb-checkpoint-defrag.xml

For people upgrading from an existing system, I might be wrong, but this migration seems effectively useless. It wouldn't have applied to HSQLDB 2 when it was run (marking it RAN), it won't run again once they've migrated (it's marked), and now we're no longer on HSQLDB 1 where it did run, so it does not apply again.

Moreover, this migration is particularly dangerous because it runs (kinda) on all DBMSes, regardless of their type, even though it seems specific to HSQLDB. It doesn't seem to be checking that it should be run only on HSQLDB (and the built-in one specifically at that). The only thing saving this migration from causing widespread issues on other external systems is the fact that the version numbers are sufficiently different between _most_ different DBMSes. (MySQL is like 5 or 8, Postgres is usually at 9 or 10 etc.). This distinction disappears if we apply it to a DBMS that has similar versions (like H2).

Furthermore, imagine someone specifying a custom HSQLDB DB url with their own finely-tuned settings. We've effectively changed their settings now.

I would propose that if possible we actually _remove_ this migration.

Set database logsizes and such via the url of the built-in DB. This (a) keeps it specific to the default HSQLDB instance, which we know is exclusively for airsonic, and (b) doesn't make it run for anything else.

airsonic-main/src/main/java/org/airsonic/player/Application.java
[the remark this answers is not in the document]
fxthomas: Ah, this scenario is a good point. I took some additional time to research and found a working solution that should work on both Spring 1.5 and 2 so that we only do the backups for the `legacy` profile. Let me know what you think!

airsonic-main/src/main/java/org/airsonic/player/Application.java
```diff
@@ -162,6 +165,13 @@ public Filter noCacheFilter() {
     }
 
     private static SpringApplicationBuilder doConfigure(SpringApplicationBuilder application) {
+        // Handle HSQLDB database upgrades from 1.8 to 2.x before any beans are started.
+        application.application().addListeners((ApplicationListener<ApplicationPreparedEvent>) event -> {
+            if (event.getApplicationContext().getEnvironment().acceptsProfiles("legacy")) {
+                LegacyHsqlUtil.upgradeHsqldbDatabaseSafely();
+            }
```
randomnicode: I _think_ this works, though you might want to check that this event listener isn't called everytime there is a context refresh. The bean definitions shouldn't matter so long as the context isn't refreshed.

randomnicode:
Actually, one more thing (and it may be minor) is that I think tests won't pick up your application listener registered in this way.

You can choose to register it in spring.factories if you want tests to pick it up

fxthomas:
@airsonic/maintainers will merge that this week-end!

PR #1285, opened 2019-10-07
Translated xml configuration to java configuration

muff1nman:
Signed-off-by: Andrew DeMaria <lostonamountain@gmail.com>

PR #1346, opened 2019-10-19
Update Spring Boot to 2.1.9

randomnicode:
This revives an old branch (the test/comment PR for which went stale). 

No new functionality/APIs introduced/deprecated. Should be a drop in replacement. The significant commits are usually commented in detail in case there are questions.
- Most dependencies, with the notable exception of HSQLDB have been upgraded
  - Spring Boot is the primary target and that brings in a host of upgrades across the board.
- Reliance on xmls has been reduced.
- Tests have been modified to rely on the new structure and config.
  - We may opt to eventually restructure the contexts in Application to make tests faster, including slicing them. For now, they run fine. These optimizations and restructurings are probably left for another PR.

Notes:
- 3 profiles, dependent on the type of DB defined:
  - *legacy* (default): Uses a persistent HSQLDB by default, though a url may be provided to connect to a different location. Does NOT use Connection Pools
  - *embed*: Connects to a DB at the provided url. Uses a Connection Pools (Hikari, by default. Removed Apache's DBCP2. Hikari is significantly better)
  - *jndi*: Searches for a jndi config
- Difference between *legacy* and *embed* is solely connection pooling, otherwise just easier to combine the two. The ancient HSQLDB (1.8) driver we use doesn't support all the contemporary methods that Connection Pools require (such as a Connection.isValid()) etc, thus necessitates keeping two separate profiles for now.
- Migrations are auto done by liquibase now.
- Had to fix some minor bugs along the way. Notably some bothersome liquibase migrations that don't generate the correct SQL.


muff1nman:
This should be broken down into smaller chunks. Yes there are aspects that will need to be together in the same PR but only if absolutely necessary. First we should try to do everything that we can before the actual jump (i.e https://github.com/airsonic/airsonic/pull/1285)  See the existing project https://github.com/airsonic/airsonic/projects/5 for tracking related work.

randomnicode:
While I like small, focused, targeted PRs as much as anyone, I respectfully disagree here. Minor upgrades, aside, I _could not_ make it any more atomic. I even mentioned it in the PR description that no functionality or features were added or deleted. I kept everything feature related for other future PRs.

Could you be more specific here about which portions need to be split up instead of just blanketly closing the PR without (a) a review, or (b) a discussion with the community regarding the merits of this PR?

I did review the open tasks in https://github.com/airsonic/airsonic/projects/5 before submitting this PR and some of them are decidedly not independent and definitely not granular enough, and will eventually come down to a PR like this. About the only separate task I might consider independent enough is localizing Config changes to a separate file, but even that is a change of 1 file at most. Otherwise, consider the first TODO task:
- Upgrade POM to Spring 2.1.x
That fails to point out that the change will bring in a slew of massive changes and updates, which will either cause compilation failures or broken tests, all of which will need to be resolved in the same PR (unless you want to leave master in a broken state).

Other tasks mentioned therein such as removing dependencies or cleaning up POMs etc I would consider are all orthogonal to the actual issue of upgrading Spring Boot and are not actually related except perhaps aesthetically.

This PR is doing most of the work of Spring Boot upgrade, in place. I'd prefer people actually review it instead of being closed without an actual review.

muff1nman:
You have 62 changed files. Looking at the first one its a change to the .gitignore. If this was the only example of something that should be split I would maybe brush it off but its not. Please take a look at #1285 for an example of a PR that is more reviewable and something that we can do to split this work.

airsonic-main/src/test/java/org/airsonic/player/service/MediaScannerServiceTestCase.java
```diff
@@ -161,12 +148,11 @@ public void testScanLibrary() {
 
     @Test
     public void testSpecialCharactersInFilename() throws Exception {
-        Resource resource = resourceLoader.getResource("MEDIAS/piano.mp3");
         String directoryName = "Muff1nman\u2019s \uFF0FMusic";
         String fileName = "Muff1nman\u2019s\uFF0FPiano.mp3";
         File artistDir = temporaryFolder.newFolder(directoryName);
         File musicFile = artistDir.toPath().resolve(fileName).toFile();
-        IOUtils.copy(resource.getInputStream(), new FileOutputStream(musicFile));
+        Resources.copy(Resources.getResource("MEDIAS/piano.mp3"), new FileOutputStream(musicFile));
```
jvoisin: This is irrelevant to the Spring bump, and also unstable.

pom.xml
```diff
@@ -16,7 +16,7 @@
     <properties>
```
jvoisin: A lot of the version bumps here are irrelevant to the Spring version bump.

integration-test/pom.xml
```diff
@@ -89,7 +89,12 @@
         <dependency>
```
jvoisin: I don't think that updating the docker-thingy is relevant to the Spring bump :/

.gitignore
```diff
@@ -110,6 +110,12 @@ local.properties
 # Eclipse Core
 .project
 
+# m2e-code-quality Eclipse IDE plugin temporary configuration files for Eclipse CS Checkstyle / PMD / SpotBugs Plug-Ins
```
jvoisin: This is irrelevant to the Spring version bump

airsonic-main/pom.xml
```diff
@@ -144,21 +153,6 @@
             <scope>runtime</scope>
         </dependency>
 
-        <!-- connection pool manager, fixes #860 (external database performance) -->
```
jvoisin: Why did you remove those?

airsonic-main/pom.xml
```diff
@@ -220,9 +214,9 @@
         </dependency>
 
         <dependency>
-            <groupId>hsqldb</groupId>
+            <groupId>org.hsqldb</groupId>
             <artifactId>hsqldb</artifactId>
-            <version>1.8.0.7</version>
+            <version>1.8.0.10</version>
```
jvoisin: Why are you updating hsqldb? Is this required to upgrade Spring?

airsonic-main/pom.xml
```diff
@@ -551,7 +551,6 @@
         <dependency>
             <groupId>javax.annotation</groupId>
             <artifactId>javax.annotation-api</artifactId>
-            <scope>runtime</scope>
```
jvoisin: Why did you remove this?

airsonic-main/pom.xml
```diff
@@ -571,19 +570,16 @@
         <dependency>
             <groupId>org.postgresql</groupId>
             <artifactId>postgresql</artifactId>
-            <version>42.2.5</version>
```
jvoisin: Why are you removing this, and how is this relevant to the Spring update?

airsonic-main/src/main/java/org/airsonic/player/Application.java
```diff
@@ -1,185 +1,79 @@
...
-        FilterRegistrationBean registration = new FilterRegistrationBean();
-        registration.setFilter(restFilter());
-        registration.addUrlPatterns("/rest/*");
-        registration.setName("RESTFilter");
-        registration.setOrder(3);
-        return registration;
-    }
+    private static SpringApplicationBuilder doConfigure(SpringApplicationBuilder application) {
+        // Customize the application or call application.sources(...) to add sources
+        // Since our example is itself a @Configuration class (via @SpringBootApplication)
```
jvoisin: Our example?

airsonic-main/src/main/java/org/airsonic/player/Application.java
```diff
@@ -224,7 +120,7 @@ public void customize(ConfigurableEmbeddedServletContainer container) {
         }
     }
 
-    public static void main(String[] args) {
+    public static void main(String[] args) throws IOException {
```
jvoisin: Why is this needed?

airsonic-main/src/main/java/org/airsonic/player/service/sonos/SonosHelper.java
```diff
@@ -116,7 +116,7 @@
```
jvoisin: There are only irrelevant stylistic changes here, please put those into an other pull-request.

airsonic-main/src/main/java/org/airsonic/player/spring/HsqlDatabase.java
```diff
@@ -16,4 +16,10 @@ public boolean supportsSchemas() {
             return false;
         }
     }
+
+    @Override
+    public int getPriority() {
+        // to ensure this instance gets selected instead of the parent
```
jvoisin: Can you please elaborate on why this is needed?

jvoisin:
This pull-request is changing a lot of code:
- it updates Spring Boot
- it does some stylistic changes in some files
- it bumps some irrelevant dependencies
- it refactors some tests, which isn't needed for the upgrade

I stopped reviewing at some point, because currently, this isn't really in a merge-able state.
It would be great if you could split this pull-request into smaller, per-topic pull-requests. It would allow:

- people with different area of interest to review the things that they're knowledgeable about. For example, I know one thing or two about tests in Java, but not much about Spring, so I would be able to quickly review the changes about the former and get them merged
- if something breaks, we can pinpoint it down to a single small commit, and only reverse this one, instead of reversing a massive change
- get it merged in an incremental fashion, instead of having to wait for someone to spend a lot of time reviewing and understanding what is going on with your changes. It would also save your from rebasing your PR and resolving conflict every time a smaller one gets in.
- it will make the commit history easier to review and understand: I don't want to have a massive changes stating "Updating Spring Boot" to show up in my blame layer when I'm trying to understand why a particular constant was changed.

airsonic-main/src/main/java/org/airsonic/player/Application.java
randomnicode: Yes, "our example"

This comment is not from me. It is part of the existing class. If you search for the doConfigure method in the existing class, you'll see that comment.

integration-test/pom.xml
randomnicode: You can rest assured that it very much is.

https://github.com/spotify/docker-client/issues/1030

airsonic-main/src/test/java/org/airsonic/player/service/MediaScannerServiceTestCase.java
randomnicode: This is a Guava @Beta annotated class, not "unstable". Please read the javadocs for the annotation for when and when not to use this class.

airsonic-main/src/main/java/org/airsonic/player/spring/HsqlDatabase.java
randomnicode: Sure!

Liquibase has a default HsqlDatabase class that it uses internally (which is the parent class of this class)

When we register this class with Liquibase, we need _this_ class to perform migrations for HSQLDB and not _that_ class. Thus we set the priority for this class to be higher (parent + 1) so it gets selected before that class

airsonic-main/src/main/java/org/airsonic/player/service/sonos/SonosHelper.java
randomnicode: That would be my IDE auto-boyscouting.

PR #1350, opened 2019-10-20
Add some initial PMD checks

eharris:
This touches pretty much all the jsp files, but they were in dire need of cleanup.
I intend to add more PMD rulesets to the checks soon.

https://maven.apache.org/plugins/maven-pmd-plugin/index.html

### DIFF
--- a/airsonic-main/src/main/webapp/WEB-INF/jsp/podcast.jsp
+++ b/airsonic-main/src/main/webapp/WEB-INF/jsp/podcast.jsp
@@ -1,5 +1,6 @@
-<%@ include file="include.jsp" %>
 <%@ page language="java" contentType="text/xml; charset=utf-8" pageEncoding="iso-8859-1"%>
+<%@ taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core" %>
+<%@ taglib prefix="fn" uri="http://java.sun.com/jsp/jstl/functions" %>
 
 <rss version="2.0">
     <channel>
--- a/airsonic-main/pom.xml
+++ b/airsonic-main/pom.xml
@@ -596,6 +596,24 @@
                 <artifactId>maven-checkstyle-plugin</artifactId>
             </plugin>
             <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-pmd-plugin</artifactId>
+                <configuration>
+                    <language>jsp</language>
+                    <rulesets>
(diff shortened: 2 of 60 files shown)


eharris:
@fxthomas @muff1nman @jvoisin please be aware this took a lot of effort, so I would appreciate if you held off any merges that touch any jsp files until after this as merged, as otherwise it will create merge conflicts.

eharris:
Pushed to fix the merge conflict created last night.

fxthomas:
Sure, we can do that!

I don't really know PMD, so my current review consists in "nothing looks fishy" :smile:

eharris:
The JSP files were pretty much a mess.  The vast majority of time spent on this PR (and it was a lot) was just getting the quoting issues in all the JSP's un-ambiguous enough that the PMD parser wouldn't choke and refuse to process the files AND in such a way that Jasper and the other parts of the code would still work properly.

I haven't yet enabled more of PMD's checks because there are so many violations that it will take quite a while to resolve them all.  This PR just gets it to the point that the checks *could* be enabled without causing compilation failures.

airsonic-main/src/main/webapp/WEB-INF/jsp/podcast.jsp
```diff
@@ -1,5 +1,6 @@
-<%@ include file="include.jsp" %>
 <%@ page language="java" contentType="text/xml; charset=utf-8" pageEncoding="iso-8859-1"%>
+<%@ taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core" %>
+<%@ taglib prefix="fn" uri="http://java.sun.com/jsp/jstl/functions" %>
```
eharris: This change was necessary because this one jsp has a different encoding specified, and the include uses UTF8 because everything else does too.  PMD (rightly) complained that two different encodings were not allowed, so I modified this to include the needed taglib directives directly.

eharris:
@muff1nman could I get your feedback and review on this PR please?  I'd like to get this merged before other jsp changes cause merge conflicts.

airsonic-main/src/main/webapp/WEB-INF/jsp/podcast.jsp
jooola: What about only using utf-8 ?

eharris: I don't know why this one had specified a different encoding, but because it was specified (rather than left unset) I have to assume there must have been a reason, so I preserved it.

Issue #1348, opened 2019-10-19
PR for Spring Boot 2.1.9 upgrade closed


randomnicode:
I'd like to reopen this PR: https://github.com/airsonic/airsonic/pull/1346 and actually have it reviewed by folks before it is either:
(a) closed for an actual reason other than it's too long, or
(b) given suggestions which bits to be incorporated (somehow) piecemeal into other PRs.

Here is the apparent task governing it:
https://github.com/airsonic/airsonic/projects/5
Aside from the orthogonal and superficial work presented there, such as cleaning up poms and removing optional runtimes or dependencies (none of which has to do with Spring Boot upgrade itself), it accounts for the bulk of work of the actual upgrade.

muff1nman:
> I'd like to reopen this PR: #1346 and actually have it reviewed by folks before it is either:
> (a) closed for an actual reason other than it's too long, or

Which is not a dismissable reason. If there is not a very very good reason for a large PR it will most likely not be reviewed, and/or stall. 

> (b) given suggestions which bits to be incorporated (somehow) piecemeal into other PRs.

In my comment on #1346 I gave you an example PR that is more composable and allows us to upgrade in a much more controlled fashion. Copying that comment here:

> You have 62 changed files. Looking at the first one its a change to the .gitignore. If this was the only example of something that should be split I would maybe brush it off but its not. Please take a look at #1285 for an example of a PR that is more reviewable and something that we can do to split this work.

randomnicode:
#1285 itself has about 30 files of changes, no?

Regardless, a lot of it seems to be similar in config consolidation to Java as #1346 and #1349 

Let me review that PR and provide comments. Maybe we can rebase #1349 on top of it.

Do we have a due for when it is going to land?

muff1nman:
> #1285 itself has about 30 files of changes, no?

Yes but its also half the size in changes, it should be harmless and a good chunk of that are some new tests. The idea with that PR is that can be done soon without having to wait for a good time to upgrade to spring 2x.

We have a lot of other things that need to be hammered out before we can push more large changes (i.e. indexing changes, mariadb issues, hsql upgrade, all the other stuff that is already in for 10.5). Right now it does not make sense to put a spring upgrade into 10.5 due to how much we have in flight. However small prep PR's can maybe go in provided they are small in scope. 

> Do we have a due for when it is going to land?
No due dates around here :) I'll try to get it updated with master so it can have a chance at getting merged.

randomnicode:
[...]

PR #1353, opened 2019-10-20
Document a bit our review process

jvoisin:
@randomnicode asked some good questions in #1348
about out governance/contribution/review/… processes,
so here is a commit to document them a bit.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,11 +17,77 @@ want a `.war` as quick as possible, you can use this instead:
 $ mvn -Dmaven.test.skip=true clean package 
 ```
 
+
 # Suggesting modifications
 
 Airsonic's source code is hosted on [github](https://github.com/airsonic/airsonic/),
 who provides a [lot of documentation](https://help.github.com/en) on how
 to contribute to projects hosted there.
+
+Keep in mind that this is a non-funded community-driven project maintained by
+a relatively small group of contributors who have many other responsibilities
+and demands on their time. Development, maintenance, and administration of the
+project is done on a best-effort basis, as time and other constraints permit.
+
+
+# Getting your pull-requests reviewed and merged
+
+Once you have submitted a pull-request, a number of factors may determine how
+much attention it receives, how quickly it may be reviewed, and whether or not
+it eventually gets accepted and merged. Here are a few guidelines that can help
+speed the process and increase the chances of a pull request being accepted and
+merged in a timely fashion:
+
+- Limit the scope to the minimum changes necessary to accomplish a discrete and
+	well defined task.
+- If your changes could be broken down into smaller units, then they are much
+	less likely to be accepted.
+- Try not to address unrelated issues in the same set of changes, unless
+	failing to do so would cause other problems (like merge conflicts).
+- Do your best to maintain prior functionality while eliminating (or at least
+	minimizing) side effects.
+- Changes that affect backward compatibility or make it more difficult to
+	upgrade or downgrade versions of libraries or installations will be the most
+	heavily scrutinized and take the longest.
+- Maintain the style and coding standards represented by the codebase.
+- Consistent, simple, and easy-to-understand changes are usually preferred.
+- Do not mix functional changes with code cleanups or style changes.
+- Make it as easy as possible for others to review your changes. Rebasing your
+	PR to address issues is strongly preferred over adding additional commits
+	that make changes to (or undo parts of) prior commits.
+- In general, the more commits in a PR, the harder it is to review and the
+	longer it will take to be reviewed. But do not sacrifice good change
+	isolation by combining commits unnecessarily.
+- If your PR needs more than 2 or 3 commits, then you *probably* need to reduce
+	the scope of your PR. If a single commit touches more than a few files or
+	more than 30-50 lines, then the scope of the commit *probably* needs to be
+	reduced.
+- Keep in mind that we strive to balance stability with new features while best
+	utilizing everybody's limited available free time. As such, a pull request may
+	be rejected if it doesn't strike that balance.
+
+And finally:
+
+- Actively maintain your PR. If any concerns are raised, or tests fail, or some
+	other change occurs in the codebase that affects your PR (like a merge
+	conflict) before your PR is accepted and merged, you are expected to address
+	and resolve those issues or the PR may be rejected.
+
+Once all concerns have been addressed, having a change accepted usually
+requires two (or more, depending on complexity and impact) [core
+contributors](https://github.com/airsonic/airsonic/graphs/contributors): one to
+explicitly approve the pull-request, and another to perform the actual merge.
+
+If you keep sending great code, you might be invited to become a *core
+contributor*.
+
+"Normal releases do not happen on any fixed schedule. They happen when the
+maintainers collectively decide that enough changes and testing have taken
+place that a new release is warranted. Bugfix releases (when a problem has been
+discovered that is likely to impact users significantly), may happen more
+quickly if needed. Even after acceptance, the inclusion of larger changes may
+be delayed until a major version release in order to ensure that the impact to
+users is minimized and that stability is maintained.
 
 # Getting help
 


CONTRIBUTING.md
```diff
@@ -17,12 +17,36 @@ want a `.war` as quick as possible, you can use this instead:
...
 who provides a [lot of documentation](https://help.github.com/en) on how
 to contribute to projects hosted there.
 
+Keep in mind that we are a small project, and that we value stability over new
+features.
+
+
+# Getting your pull-requests reviewed an merged
+
+Once you issued a pull-request, it will be reviewed by other contributors, and
```
muff1nman: This might be a promise that we can't keep. You touch on it below a bit, but if a PR's quality (size, tests, code quality, etc) is below an arbitrary line, it might not be reviewed by anyone..

CONTRIBUTING.md
```diff
@@ -17,12 +17,36 @@ want a `.war` as quick as possible, you can use this instead:
...
+modifications/enhancement could be suggested. Once all the concerns are
+addressed, it might get merged, which usually requires two [core
+contributors](https://github.com/airsonic/airsonic/graphs/contributors): one to
+explicitly approve the pull-request, and a second to do the actual merge.
+Note that incremental/small changes are likely going to get reviewed, accepted
+and merged quicker than huge ones.
```
muff1nman: Do we need a note in here like: "Keep in mind that we strive to balance stability with new features while best utilizing our limited time. As such, your Pull Request might be rejected when it doesn't strike that balance."

randomnicode: Or we could just have better build practices?

Release stability should really not be mixed with branch stability or usability.

If we have that much of a concern about keeping master stable, then we should provide additional branches for people to merge commits to, and merge in those branches into master after releases.

Release cycles should be separated and less tightly coupled to branch commits.

CONTRIBUTING.md
```diff
@@ -17,12 +17,36 @@ want a `.war` as quick as possible, you can use this instead:
...
+
+If you keep sending great code, you might be invited to become a *core
+contributor*.
+
+Releases aren't happening at fixed dates: they happen when enough bugs have
+been fixed, or new features are stable enough. We might delay the inclusion of 
+big additions to the code base, to ensure the stability of airsonic before
+doing a new release.
```
randomnicode: Maybe add something about the nightlies here if people want to be on the latest and greatest _once they become available_

CONTRIBUTING.md
```diff
@@ -17,12 +17,36 @@ want a `.war` as quick as possible, you can use this instead:
...
 
+
 # Suggesting modifications
 
 Airsonic's source code is hosted on [github](https://github.com/airsonic/airsonic/),
 who provides a [lot of documentation](https://help.github.com/en) on how
 to contribute to projects hosted there.
 
+Keep in mind that we are a small project, and that we value stability over new
+features.
```
randomnicode: I really don't understand this sentence?
Build stability? Your travis should take care of that?
What kind of stability are we talking about with this sentence?
Releases cycles should be an orthogonal concern from commits

CONTRIBUTING.md
eharris: Additional text suggested:

Once you have submitted a pull-request, a number of factors may determine how much attention it receives, how quickly it may be reviewed, and whether or not it eventually gets accepted and merged.  Here are a few guidelines that can help speed the process and increase the chances of a pull request being accepted and merged in a timely fashion:

* Limit the scope to the minimum changes necessary to accomplish a discrete and well defined task.
* If your changes could be broken down into smaller units, then they are much less likely to be accepted.
* Try not to address unrelated issues in the same set of changes, unless failing to do so would cause other problems (like merge conflicts).
* Do your best to maintain prior functionality while eliminating (or at least minimizing) side effects.
* Maintain the style and coding standards represented by the codebase.
* Consistent, simple, and easy-to-understand changes are usually preferred.
* Do not mix functional changes with code cleanups or style changes.
* Make it as easy as possible on the maintainers to review your changes.  Rebasing your PR to address issues is preferred over adding additional commits that make changes to (or undo parts of) prior commits.
* In general, the more commits in a PR, the harder it is to review and the longer it will take to be reviewed.  But do not sacrifice good change isolation by combining commits unnecessarily.
* If your PR needs more than 2 or 3 commits, then you probably need to reduce the scope of your PR.  If a single commit touches more than a few files or more than 30-50 lines, then the scope of the commit probably needs to be reduced.

And finally:

* Actively maintain your PR.  If any concerns are raised, or tests fail, or some other change occurs in the codebase that affects your PR (like a merge conflict) before your PR is accepted and merged, you are expected to address and resolve those issues or the PR may be rejected.

CONTRIBUTING.md
```diff
@@ -17,12 +17,36 @@ want a `.war` as quick as possible, you can use this instead:
...
+contributors](https://github.com/airsonic/airsonic/graphs/contributors): one to
+explicitly approve the pull-request, and a second to do the actual merge.
+Note that incremental/small changes are likely going to get reviewed, accepted
+and merged quicker than huge ones.
+
+If you keep sending great code, you might be invited to become a *core
+contributor*.
+
+Releases aren't happening at fixed dates: they happen when enough bugs have
+been fixed, or new features are stable enough. We might delay the inclusion of
```
eharris: "or when changes and new features have had enough testing to be considered stable."

CONTRIBUTING.md
```diff
@@ -17,12 +17,36 @@ want a `.war` as quick as possible, you can use this instead:
...
+explicitly approve the pull-request, and a second to do the actual merge.
+Note that incremental/small changes are likely going to get reviewed, accepted
+and merged quicker than huge ones.
+
+If you keep sending great code, you might be invited to become a *core
+contributor*.
+
+Releases aren't happening at fixed dates: they happen when enough bugs have
+been fixed, or new features are stable enough. We might delay the inclusion of 
+big additions to the code base, to ensure the stability of airsonic before
```
eharris: Suggest using "large" instead of "big", and "creating" rather than "doing"

CONTRIBUTING.md
jvoisin: I don't get why you prefer to have pull-requests in branches, instead of having, … well, pull-requests. What would the advantages be?

CONTRIBUTING.md
```diff
@@ -24,27 +24,67 @@ Airsonic's source code is hosted on [github](https://github.com/airsonic/airsoni
...
+
+And finally:
+
+- Actively maintain your PR. If any concerns are raised, or tests fail, or some
+	other change occurs in the codebase that affects your PR (like a merge
+	conflict) before your PR is accepted and merged, you are expected to address
+	and resolve those issues or the PR may be rejected.
+
+Once all the concerns are addressed, it might get merged, which usually
+requires two [core
```
eharris: Suggested change: "Once all concerns have been addressed, having a change accepted usually
requires two (or more, depending on complexity and impact) [core"

CONTRIBUTING.md
```diff
@@ -24,27 +24,67 @@ Airsonic's source code is hosted on [github](https://github.com/airsonic/airsoni
...
 contributors](https://github.com/airsonic/airsonic/graphs/contributors): one to
 explicitly approve the pull-request, and a second to do the actual merge.
```
eharris: Suggest changing "a second to do" to "and another to perform".

CONTRIBUTING.md
```diff
@@ -24,27 +24,67 @@ Airsonic's source code is hosted on [github](https://github.com/airsonic/airsoni
...
-Note that incremental/small changes are likely going to get reviewed, accepted
-and merged quicker than huge ones.
 
 If you keep sending great code, you might be invited to become a *core
 contributor*.
 
 Releases aren't happening at fixed dates: they happen when enough bugs have
```
eharris: Suggest changing to:

"Normal releases do not happen on any fixed schedule.  They happen when the maintainers collectively decide that enough changes and testing have taken place that a new release is warranted.  Bugfix releases (when a problem has been discovered that is likely to impact users significantly), may happen more quickly if needed.  Even after acceptance, the inclusion of larger changes may be delayed until a major version release in order to ensure that the impact to users is minimized and that stability is maintained.

CONTRIBUTING.md
jvoisin: I'll add this once we have nightlies :)

CONTRIBUTING.md
tesshucom: I think the commit granularity can be set flexibly.
This level of expression may be fine :

```If the PR is known to be very large, the committer can check with the maintainer in advance about the degree of division of the PR.```

I think the search update was this way (I think..).
The maintainer can takes the lead in sizing, gives advice, and reduce friction.

Large and small can become subjective, so it may be desirable to resolve them before accepting them.

CONTRIBUTING.md
jvoisin: I rephrased this a bit :)

Issue #1325, opened 2019-10-13
Do we want to use Scrutinizer?


jvoisin:
Yet an other static analyser running for free in the cloud™, to help us catch bugs.
It's currently running on my fork, results can be seen [here]( https://scrutinizer-ci.com/g/jvoisin/airsonic/issues/master ): it found some interesting bugs.

eharris:
Looks interesting.  I'm a fan of more automatic checks to help us find bugs or improve code.  Looks like it found a lot too, would be nice to have them cleaned up as far as I can see.

eharris:
I actually just submitted a PR (#1355) to fix one of the issues identified by Scrutinizer (duplicated code).

jooola:
I'm all for it. Having the more data possible the health of the code base it a real plus.

The only drawback I see by using all those tools is that we become dependent on Github & Cie.
Even if we know we won't move away from github anytime soon.

jvoisin:
Those fancy tools can usually be used from gitlab/gitea/gogs/… as well :)

muff1nman:
Hence why I like trying as best we can to pull these into the maven lifecycle. Having them in travis/etc should be last resort.

PR #1354, opened 2019-10-21
Changed final constant to add static

eharris:


eharris:
This change was identified and prompted by Scrutinizer (#1325).

PR #1352, opened 2019-10-20
Allow previous Liquibase migration checksums for MariaDB

fxthomas:
If someone was using MariaDB before #1188, they were probably doing so by using the MySQL JDBC driver to pass the migration (which had the proper workarounds in place for some column types).

This commit assumes that the migrations done that way were valid and produced the same result as running the MariaDB migrations after #1188.

This should fix #1333.

PR #1358, opened 2019-10-21
Fix erroneous liquibase migrations

randomnicode:
addPrimaryKey with composite keys is for some reason generating SQL (on HSQL) that borrows the name of the fk constraint of one of the component columns (chosen arbitrarily). That results in an error on subsequent startups because two constraints with the same name conflict.

I'm surprised this bug hasn't been caught before and my only suspicion is (a) this is just locally for me on an upgraded liquibase, or (b) that most folks are probably migrating from older dbs which did not have this migration run via liquibase.

To fix, just annotate the columns directly with the primaryKey tag. Liquibase will automatically generate correctly named composite primary keys.

Also removed the nullable annotation. Primary key annotation makes nullable redundant (pks cannot be nullable, by definition)..

These were the only instances of this pattern found. At best these changes are synonymous with the existing migration.

To reproduce:
- Start airsonic with no existing db (to run migrations via liquibase)
- Either examine the SQL produced in the airsonic.script db file, or try restarting airsonic pointing at the same db and watch it fail gloriously
  - A symptom should exist on line 7 (user_role table)

https://github.com/liquibase/liquibase/pull/167 is a registered bug that says something similar was fixed, but I still see issues as of liquibase 3.8.0 for HSQLDB.


fxthomas:
Definitely in favor of changing to a real primary key, this is just good design.

On the other hand I cannot reproduce the bug on a fresh compiled build (on 666bc66e just before your commit):

```
fx@xxxx migrationsbugfix [666bc66e] $ java -Dairsonic.home=.run -Dserver.port=4040 -jar airsonic.war                                                                            
...much log, migrations, blah blah...
Started Application in ... seconds (JVM running for ...)
...Ctrl-C...

fx@xxxx migrationsbugfix [666bc66e] $ java -Dairsonic.home=.run -Dserver.port=4040 -jar airsonic.war                                                                            
...much log, no crash...
Started Application in ... seconds (JVM running for ...)
```

randomnicode:
> Also, in general make sure you include the validation checksums in the commit, I get this when upgrading from a previous version:

Yep, good catch. Let me add those checksums. Depending on what gets merged first, there'll be a conflict with https://github.com/airsonic/airsonic/pull/1362

Re: reproduction, hmmm, I can't reproduce it with Spring Boot 1.5 either (liquibase version does not seem to matter). I suspect it's some interaction with Spring Boot 2. Regardless, probably best to declare pks on the columns themselves instead of a separate constraint? Either way, this should be a harmless merge. At worst, a no-op.

Famous last words.

fxthomas:
#1362 is for creating a new release (since it impacts a lot of MariaDB users), but it has already been merged in master as #1352. So feel free to use it directly ;)

airsonic-main/src/main/resources/liquibase/legacy/schema25.xml
```diff
@@ -92,21 +92,20 @@
...
+        <validCheckSum>8:73d2b294465ee42488a8f0238cf686cc</validCheckSum>
         <preConditions onFail="MARK_RAN">
             <not>
                 <tableExists tableName="user_role" />
             </not>
         </preConditions>
         <createTable tableName="user_role">
             <column name="username" type="${varchar_type}">
-                <constraints nullable="false" foreignKeyName="ur_u_fk" referencedTableName="user" referencedColumnNames="username"/>
+                <constraints primaryKey="true" foreignKeyName="ur_u_fk" referencedTableName="user" referencedColumnNames="username"/>
```
muff1nman: These should all be done in new migrations. Editing existing ones is a last resort.

randomnicode: Spring Boot 2 is producing an incorrect SQL for these migrations. These migrations have to be changed. Adding another migration later won't change the incorrect SQL generated here that is persisted in the HSQLDB script file.

Please see the output of the failed tests for https://github.com/airsonic/airsonic/pull/1372 which shows the incorrect SQL this migration is generating. That PR won't pass until this is merged into _master_.
```
Error creating bean with name 'liquibase' defined in class path resource [org/airsonic/player/spring/DatabaseConfiguration.class]: Invocation of init method failed; nested exception is liquibase.exception.DatabaseException: java.sql.SQLException: error in script file line: 7 Constraint already exists: UR_U_FK in statement [CREATE MEMORY TABLE USER_ROLE(USERNAME VARCHAR NOT NULL,ROLE_ID INTEGER NOT NULL,CONSTRAINT UR_U_FK PRIMARY KEY(USERNAME,ROLE_ID),CONSTRAINT UR_U_FK FOREIGN KEY(USERNAME) REFERENCES "USER"(USERNAME),CONSTRAINT UR_R_FK FOREIGN KEY(ROLE_ID) REFERENCES "ROLE"(ID))]
```

Notice the name of the primary key is the same as the name of a foreign key (UR_U_FK), thus the duplicate constraint error

randomnicode:
>#1362 is for creating a new release (since it impacts a lot of MariaDB users), but it has already been merged in master as #1352. So feel free to use it directly ;)

@fxthomas I've edited/added the checksums using master

airsonic-main/src/main/resources/liquibase/legacy/schema25.xml
fxthomas: Any idea things are going to be different if a column is defined as a primary key in some user databases and not others? I'd rather guarantee that everyone has the exact same db schema, even if we have to somehow handle this manually in the migrations.

randomnicode: Sorry, not sure I understand the question? Mind clarifying?

This migration retains the same primary keys, so nothing should change. It's just "written" differently in liquibase parlance.

Previously: You add columns and then impose a primary key constraint on those columns in the next line (addPrimaryKey tag)
Now: You add columns and specify they're part of a primary key as part of the column declaration itself (and remove the unnecessary addPrimaryKey tag)

Either way, the same columns end up being primary keys in SQL.

From a migration perspective:
- People with new DBs: won't know the difference. The new migrations get applied and the same columns (as the older migrations) end up being primary keys.
- People with old DBs: no-ops. The table already exists with the same primary keys, the migrations don't even need to run. Only thing that changes are the validation checksums stored by liquibase for the migrations, which we've specified.

harawata:
If I understand the 'bug' correctly, it seems to be better to specify `constraintName` explicitly in each `addPrimaryKey`.
https://www.liquibase.org/documentation/changes/add_primary_key.html

airsonic-main/src/main/resources/liquibase/legacy/schema25.xml
fxthomas: Thanks for the detailed explanation, that makes it clearer! I think there aren't any issues with the PR then.

randomnicode:
@harawata that's the first thing I believe I tried when I started looking at the issue to no avail. It kept taking one of the fk names and ignoring its own.

harawata:
Thanks for the reply, @randomnicode !
It got me curious and, after some digging, I found that it was actually HSQLDB 1.8.0.10 that generated the problematic constraint name.
Here are the definitions of `USER_ROLE` table generated by different versions of HSQLDB (I inserted line breaks).

```
-- HSQLDB 1.8.0.7 (master a9a7b0823091a9d20b9715479d700622b20f9b6c)
CREATE MEMORY TABLE USER_ROLE(
  USERNAME VARCHAR NOT NULL,
  ROLE_ID INTEGER NOT NULL,
  PRIMARY KEY(USERNAME,ROLE_ID),
  CONSTRAINT UR_R_FK FOREIGN KEY(ROLE_ID) REFERENCES "ROLE"(ID),
  CONSTRAINT UR_U_FK FOREIGN KEY(USERNAME) REFERENCES "USER"(USERNAME)
)
-- HSQLDB 1.8.0.10 (randomnicode:migrationsbugfix d38241beb46b3bab3798ce87bb35c4a05e89c3a1)
CREATE MEMORY TABLE USER_ROLE(
  USERNAME VARCHAR NOT NULL,
  ROLE_ID INTEGER NOT NULL,
  CONSTRAINT UR_U_FK PRIMARY KEY(USERNAME,ROLE_ID),
  CONSTRAINT UR_U_FK FOREIGN KEY(USERNAME) REFERENCES "USER"(USERNAME),
  CONSTRAINT UR_R_FK FOREIGN KEY(ROLE_ID) REFERENCES "ROLE"(ID)
)
-- HSQLDB 2.5.0 (PR #1345)
CREATE MEMORY TABLE PUBLIC.USER_ROLE(
  USERNAME VARCHAR(32768) NOT NULL,
  ROLE_ID INTEGER NOT NULL,
  PRIMARY KEY(USERNAME,ROLE_ID),
  CONSTRAINT UR_U_FK FOREIGN KEY(USERNAME) REFERENCES PUBLIC."USER"(USERNAME),
  CONSTRAINT UR_R_FK FOREIGN KEY(ROLE_ID) REFERENCES PUBLIC."ROLE"(ID)
)
```

I'd say just stay away from 1.8.0.10. ;D

randomnicode:
Hi @harawata , thanks for digging in. That's an interesting observation, and I think right on the money. I totally forgot the minor revision change in hsqldb I left in there with all the refactors.

Though I don't think we can compare the tip of master to upgradespringboot2 (because of an additional change in Spring Boot), you were on to something with the hsqldb change.

1.8.0.7 works fine on Spring Boot 1.5
2.5.0 works fine on Spring Boot 1.5
1.8.0.10 does not work with Spring Boot 2

So, I checked upgrading 1.8.0.10 on master to test:
1.8.0.10 on Spring Boot 1.5

Indeed, it's an issue with 1.8.0.10:
```
CREATE MEMORY TABLE USER_ROLE(
  USERNAME VARCHAR NOT NULL,
  ROLE_ID INTEGER NOT NULL,
  CONSTRAINT UR_R_FK PRIMARY KEY(USERNAME,ROLE_ID),
  CONSTRAINT UR_R_FK FOREIGN KEY(ROLE_ID) REFERENCES "ROLE"(ID),
  CONSTRAINT UR_U_FK FOREIGN KEY(USERNAME) REFERENCES "USER"(USERNAME)
)
```

As such, there is less of a need to get this PR in. Still harmless to get it in though...

randomnicode:
What's the hold up on this PR? Can we merge this? It has been up for more than a week and has been approved.

fxthomas:
@muff1nman you okay with merging this?

airsonic-main/src/main/resources/liquibase/legacy/schema25.xml
```diff
@@ -92,21 +92,20 @@
         </rollback>
     </changeSet>
     <changeSet id="schema25_005" author="muff1nman">
-        <validCheckSum>8:4d92e7c97abc6f197f1236b0353ec2f7</validCheckSum>
+        <validCheckSum>8:73d2b294465ee42488a8f0238cf686cc</validCheckSum>
```
muff1nman: Please add - don't remove, existing validCheckSum entries.

fxthomas:
Migration to 2.5 will happen soon anyway, so I'll close it as not needed anymore. Let me know if you want to reopen it.

PR #1359, opened 2019-10-21
Minor fix to static time unit call

randomnicode:
Looks like the keyboard got eager.


fxthomas:
Let's merge this, looks good to me!

PR #1361, opened 2019-10-22
Removed unused variables, imports, empty statements

eharris:
Also some unnecessary whitespace.

Unused vars were found using Scrutinizer (#1325)


randomnicode:
Can we merge this? It was approved.

These kinds of boyscouting PRs don't need a long turnaround time. 11 days, and it's still sitting out here after approval taking up real estate.

It looks good to me with almost no impact. Let's merge this.

PR #1349, opened 2019-10-19
(Reopen) Update Spring Boot to 2.1.9

randomnicode:
Reopening https://github.com/airsonic/airsonic/pull/1346

Please don't close prior to an actual review or discussion regarding it's breakdown into targeted PRs.

This PR targets that actual version upgrade and the consolidation of the config into Java in this checklist:
https://github.com/airsonic/airsonic/projects/5

The rest of the checklist is mostly orthogonal and irrelevant to the actual Spring Boot version upgrade, in my opinion (and should be moved into another generic "upgrade" task).

PR #1364, opened 2019-10-22
Spring Boot 2 Upgrade: Part 1 - Remove xml

randomnicode:
This incorporates and supercedes https://github.com/airsonic/airsonic/pull/1285 and is an incomplete breakdown of  https://github.com/airsonic/airsonic/pull/1349

- Mostly the same as https://github.com/airsonic/airsonic/pull/1285 except rids of some potential bugs
- Merged current master into that commit (to remove conflicts etc)
- Componentized some beans that don't need to be literally translated

This is Step 1 of a series of PRs each of which will build on each other
 - Not sure how it is now, but back in the day, GitHub would show the entire list of changes unless the previous PRs in the series were merged

Next up: Consolidate DB Config


airsonic-main/src/test/java/org/airsonic/player/service/metadata/MetaDataFactoryTestCase.java
```diff
@@ -0,0 +1,86 @@
...
+        HomeRule airsonicRule = new HomeRule() {
+            @Override
+            protected void before() throws Throwable {
+                super.before();
+                String homeParent = TestCaseUtils.airsonicHomePathForTest();
+                System.setProperty("airsonic.home", TestCaseUtils.airsonicHomePathForTest());
+                TestCaseUtils.cleanAirsonicHomeForTest();
+                File dbDirectory = new File(homeParent, "/db");
+                FileUtils.forceMkdir(dbDirectory);
+                org.airsonic.player.util.FileUtils.copyResourcesRecursively(getClass().getResource("/db/pre-liquibase/db"), new File(homeParent));
```
randomnicode: I think this is a potential bug for now.
It also exists in LegacyDatabaseStartupTestCase.

pre-liquibase files are libresonic. The default jdbc url will point to airsonic (because of the presence of string 'airsonic' name in the temporary test folder).

airsonic-main/src/test/java/org/airsonic/player/api/jukebox/AbstractAirsonicRestApiJukeboxIntTest.java
```diff
@@ -176,10 +177,7 @@ private ResultMatcher playListItem1isCorrect() {
             jsonPath("$.subsonic-response.jukeboxPlaylist.entry[0].suffix").value(mediaFile.getFormat()).match(result);
             jsonPath("$.subsonic-response.jukeboxPlaylist.entry[0].duration").value(mediaFile.getDurationSeconds()).match(result);
             jsonPath("$.subsonic-response.jukeboxPlaylist.entry[0].bitRate").value(mediaFile.getBitRate()).match(result);
-
-            // The path is absolute, we remove the folder source with replace the folder with nothing
-            // and the "/" they stay with substring(1), must be a method in mediaFile for method to obtain the relative path is same every where.
-            jsonPath("$.subsonic-response.jukeboxPlaylist.entry[0].path").value(mediaFile.getPath().replace(mediaFile.getFolder(), "").substring(1)).match(result);
+            jsonPath("$.subsonic-response.jukeboxPlaylist.entry[0].path").value(SubsonicRESTController.getRelativePath(mediaFile, settingsService)).match(result);
```
randomnicode: This test is used to be not cross-platform as the deleted comment mentions.

This change should make it cross platform.

airsonic-main/src/test/java/org/airsonic/player/service/MediaScannerServiceTestCase.java
```diff
@@ -161,12 +152,11 @@ public void testScanLibrary() {
 
     @Test
     public void testSpecialCharactersInFilename() throws Exception {
-        Resource resource = resourceLoader.getResource("MEDIAS/piano.mp3");
         String directoryName = "Muff1nman\u2019s \uFF0FMusic";
         String fileName = "Muff1nman\u2019s\uFF0FPiano.mp3";
         File artistDir = temporaryFolder.newFolder(directoryName);
         File musicFile = artistDir.toPath().resolve(fileName).toFile();
-        IOUtils.copy(resource.getInputStream(), new FileOutputStream(musicFile));
+        Resources.copy(Resources.getResource("MEDIAS/piano.mp3"), new FileOutputStream(musicFile));
```
jvoisin: This is marked as unstable, I'm not sure we should use it :/

randomnicode: This is a guava class, so we should be fine. I think it's just annotated as @Beta, which just means it's not feature-frozen at the class-level.

The Beta javadoc for Guava notes should tell you the meaning of the annotation.

airsonic-main/src/main/java/org/airsonic/player/spring/SonosConfiguration.java
```diff
@@ -0,0 +1,26 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.ImportResource;
+
+import javax.xml.ws.Endpoint;
+
+import java.util.Collections;
+
+@Configuration
+@ImportResource({"classpath:META-INF/cxf/cxf.xml", "classpath:META-INF/cxf/cxf-servlet.xml"})
```
fxthomas: Since we're removing XML, shouldn't the Sonos configuration also be part of it?

fxthomas:
Eh, well, crap.

Screenshot from 2019-11-02 19-36-25

fxthomas:
Ah, right, you have to add `index/` at the end of the URL, or it won't work.

Maybe this is something we can correct?

muff1nman:
I'd prefer if we merge  #1285 first as its a smaller set of changes. Regardless let's do all this spring stuff after 10.5 goes out.

airsonic-main/src/main/java/org/airsonic/player/spring/SonosConfiguration.java
randomnicode: Haven't dwelved too much into this, but this config is not coming from airsonic-main. It's probably from another jar or autogenerated. So it isn't part of this PR.

randomnicode:
- It's 35 vs 30 files. I think the size is similar and comparable.
- This PR incorporates the original commit from #1285 (see first commit of this PR).
- In the absence of dev builds, it really doesn't matter whether #1285 gets merged first, or that the "scope" of the changes is smaller in that PR. End users will get all the commits together at release (which isn't going to happen with just #1285 merged), and developers can just checkout to an older commit to test a build. So I don't know why an intermediate PR that was incorporated in another PR should be merged at the risk of causing merge conflicts in 5 additional PRs, particularly since the commit history was rewritten in the intermediate PR.

I think if #1285 had resolved conflicts by merging master instead of a rebase, then this would've been a relatively painless merge point and #1285 could've been merged first. But as it stands now, it'll cause an additional amount of work for everyone, particularly me, where I'll have to merge in conflicts on at least 6 PRs (basically everything in this series).

Additionally there are still issues with #1285 (there seems to be one with one too, but it seems closer at least).
I think we should close #1285 and merge this one, which supercedes it. If people want to see the "smaller" scope of changes, they can just checkout the older commit and build that.

fxthomas:
After testing #1285 I agree with @randomnicode : no need to duplicate effort.

IMO this one can be merged once the last issue in https://github.com/airsonic/airsonic/pull/1364#issuecomment-549069931 is resolved.

tesshucom:
The current master has problem with Some SpringBeans double execution.
I suspected reloading the web, but multiple executions were confirmed by logs from API access, so double execution due to configuration errors often seen in Spring is expected.
I tried to go back in the history of the past, but gave up because it was difficult to compile.

In any case, to solve this problem, you will do exactly the same approach that randomnicode takes.
I want to verify with this PR.
If the original is very clean, I think you should be cautious about the contamination.
But this PR is a good medicine.
(It is a pity that there are currently unrelated Metrix compilation errors &#x1f612;)

tesshucom:
The current master is suspected of dual execution in some SpringBeans.
I have just found it and are currently investigating it whether it is a bug.
Initially I suspected a web reload, but the logs from API access confirmed multiple executions.
For this reason, we expected double execution due to misconfiguration often seen in Spring.
I tried to go back to past history.
However, because it was difficult to compile, I gaved up.

In any case, current config is certainly terrible, so we take the exact same approach that randomnicode takes.
I am checking the operation of this PR and #1285.

If the original is very clean, I think you should be cautious about the contamination.
But this PR and #1285 is a good medicine.
(It is a pity that there are currently unrelated Metrix compilation errors &#x1f612;)

tesshucom:
The current master is suspected of dual execution in some SpringBeans.
I have just found it and are currently investigating it whether it is a bug.
Initially I suspected a web reload, but multiple log putputs from API access confirmed.
(Double execution due to misconfiguration often seen in Spring..?)
I tried to go back to past history.
However, because it was difficult to compile, I gaved up.

In any case, current config is certainly terrible, so we'll take the exact same approach that randomnicode takes.
I am checking the operation of this PR and #1285.

If the original is very clean, I think you should be cautious about the contamination.
But both this PR and #1285 is a good medicine.
(It is a pity that there are currently unrelated Metrix compilation errors &#x1f612;)

randomnicode:
@fxthomas yeah, good eye, turns out we had disabled autoconfiguration by explicitly annotating EnableWebMvc

Pushed out a fix. The WelcomeRegistrationHandler should be back and autoconfigured in.
The error handler should be fixed too.

host:port -> works fine
host:port/ -> works fine
host:port/index -> works fine
host:port/index/ -> shows 404 exceptions on all panels (this is the current behavior also). does not show the '/error is not registered' exception

This should be okay to merge now.

airsonic-main/src/test/java/org/airsonic/player/api/jukebox/AbstractAirsonicRestApiJukeboxIntTest.java
muff1nman: This should be in a separate PR.

airsonic-main/src/test/java/org/airsonic/player/service/metadata/MetaDataFactoryTestCase.java
muff1nman: This isn't correct. See #1285

airsonic-main/src/main/java/org/airsonic/player/spring/LoggingExceptionResolver.java
```diff
@@ -4,12 +4,16 @@
...
+import org.springframework.core.annotation.Order;
+import org.springframework.stereotype.Component;
 import org.springframework.web.servlet.HandlerExceptionResolver;
 import org.springframework.web.servlet.ModelAndView;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+@Component
+@Order(Ordered.HIGHEST_PRECEDENCE)
```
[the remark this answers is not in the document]
randomnicode: I'm okay with this being here. The annotation is actually the first thing in the class and makes it clearer what it's priority order is.

airsonic-main/src/main/java/org/airsonic/player/dao/UserDao.java
```diff
@@ -71,7 +72,7 @@
     private final String userTableQuote;
 
     @Autowired
-    public UserDao(String userTableQuote) {
+    public UserDao(@Value("${DatabaseUsertableQuote:}") String userTableQuote) {
```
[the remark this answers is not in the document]
randomnicode: Hmmm, I disagree, what property is being duplicated again?

The notion of creating a "bean" for a String is just weird. Which then ends up being supplied to the actual constructors that need it again. So instead of 2 places, now that String is in 3 places,.

Let's say I create a "bean", this constructor still needs something for that bean to be injected here, right? So you still end up putting an annotation here anyway and in the other constructor. And then on top of that you now have a managed bean, that is nothing but a String from the environment. What's being duplicated more again?

airsonic-main/src/test/java/org/airsonic/player/service/metadata/MetaDataFactoryTestCase.java
randomnicode: What isn't correct? This test class is from your commit.

airsonic-main/src/test/java/org/airsonic/player/api/jukebox/AbstractAirsonicRestApiJukeboxIntTest.java
randomnicode: On one note, yes, perhaps, but on another, my tests on multiple systems are failing due to this and I can't get this PR to pass locally on a few systems without this change. I don't want to remove a change that is necessary for the PR to pass. So I'm going to keep it in this PR.

PR #1365, opened 2019-10-23
Removed duplicated code and other cleanups

eharris:
Removed duplicated code, changed to use short cut logic, tightened scope of a variable, and changed an exception name to ignored.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/domain/PlayQueue.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/domain/PlayQueue.java
@@ -311,17 +311,7 @@ public class PlayQueue {
      * @param index The playlist index.
      */
     public synchronized void moveUp(int index) {
-        makeBackup();
-        if (index <= 0 || index >= size()) {
-            return;
-        }
-        Collections.swap(files, index, index - 1);
-
-        if (this.index == index) {
-            this.index--;
-        } else if (this.index == index - 1) {
-            this.index++;
-        }
+        moveDown(index - 1);
     }
 
     /**
--- a/airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
@@ -219,42 +219,27 @@ public class JaudiotaggerParser extends MetaDataParser {
      * track numbers on the form "4/12".
      */
     private Integer parseTrackNumber(String trackNumber) {
-        if (trackNumber == null) {
-            return null;
-        }
-
-        Integer result = null;
-
-        try {
-            result = Integer.valueOf(trackNumber);
-        } catch (NumberFormatException x) {
-            Matcher matcher = TRACK_NUMBER_PATTERN.matcher(trackNumber);
-            if (matcher.matches()) {
-                try {
-                    result = Integer.valueOf(matcher.group(1));
-                } catch (NumberFormatException e) {
-                    return null;
-                }
-            }
-        }
-
-        if (Integer.valueOf(0).equals(result)) {
-            return null;
-        }
-        return result;
+        return parseIntegerPattern(trackNumber, TRACK_NUMBER_PATTERN);
     }
 
     private Integer parseYear(String year) {
-        if (year == null) {
+        return parseIntegerPattern(year, YEAR_NUMBER_PATTERN);
+    }
+
+    private Integer parseIntegerPattern(String str, Pattern pattern) {
+        if (str == null) {
             return null;
         }
 
         Integer result = null;
 
         try {
-            result = Integer.valueOf(year);
+            result = Integer.valueOf(str);
         } catch (NumberFormatException x) {
-            Matcher matcher = YEAR_NUMBER_PATTERN.matcher(year);
+            if (pattern == null) {
+                return null;
+            }
+            Matcher matcher = pattern.matcher(str);
             if (matcher.matches()) {
                 try {
                     result = Integer.valueOf(matcher.group(1));
@@ -272,18 +257,7 @@ public class JaudiotaggerParser extends MetaDataParser {
 
     private Integer parseInteger(String s) {
         s = StringUtils.trimToNull(s);
-        if (s == null) {
-            return null;
-        }
-        try {
-            Integer result = Integer.valueOf(s);
-            if (Integer.valueOf(0).equals(result)) {
-                return null;
-            }
-            return result;
-        } catch (NumberFormatException x) {
-            return null;
-        }
+        return parseIntegerPattern(s, null);
     }
 
     /**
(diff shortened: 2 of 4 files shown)


airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
```diff
@@ -219,42 +219,27 @@ private String mapGenre(String genre) {
...
-        return result;
+        return parseIntegerPattern(trackNumber, TRACK_NUMBER_PATTERN);
     }
 
     private Integer parseYear(String year) {
-        if (year == null) {
+        return parseIntegerPattern(year, YEAR_NUMBER_PATTERN);
+    }
+
+    private Integer parseIntegerPattern(String str, Pattern pattern) {
```
jvoisin: Why isn't zero an acceptable return value?

airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
```diff
@@ -219,42 +219,27 @@ private String mapGenre(String genre) {
...
-            result = Integer.valueOf(year);
+            result = Integer.valueOf(str);
         } catch (NumberFormatException x) {
-            Matcher matcher = YEAR_NUMBER_PATTERN.matcher(year);
+            if (pattern == null) {
+                return null;
+            }
+            Matcher matcher = pattern.matcher(str);
             if (matcher.matches()) {
                 try {
```
jvoisin: Please refactor this a bit to avoid nested exceptions ♥

airsonic-main/src/main/java/org/airsonic/player/spring/RegisterPrecompiledJSPInitializer.java
```diff
@@ -59,14 +59,13 @@ private static WebApp parseXmlFragment() {
...
-        JAXBContext jaxbContext;
         try {
-            jaxbContext = new JAXBDataBinding(WebApp.class).getContext();
+            JAXBContext jaxbContext = new JAXBDataBinding(WebApp.class).getContext();
             Unmarshaller unmarshaller = jaxbContext.createUnmarshaller();
             WebApp webapp = (WebApp) unmarshaller.unmarshal(webXmlIS);
             try {
                 webXmlIS.close();
-            } catch (java.io.IOException e) {}
+            } catch (java.io.IOException ignored) {}
```
randomnicode: very very minor: maybe could use a try with resources in the outer try? then you won't need this awkward inner try, and it'll autoclose the stream (though I don't know why the precompiledJspWebXml isn't closed here)
```
try( InputStream precompiledJspWebXml = ....; InputStream webXmlIs = ...) {
   ...
   return (WebApp) unmarshaller.unmarshal(webXmlIs); 
} catch (JAXBException e) {
    ...
} catch (IOException ignored) {}
```

airsonic-main/src/main/java/org/airsonic/player/domain/PlayQueue.java
```diff
@@ -311,17 +311,7 @@ public synchronized void rearrange(int[] indexes) {
      * @param index The playlist index.
      */
     public synchronized void moveUp(int index) {
-        makeBackup();
-        if (index <= 0 || index >= size()) {
-            return;
-        }
```
randomnicode: Just out of curiosity, any reason why the backup isn't made _after_ the noop condition?

airsonic-main/src/main/java/org/airsonic/player/domain/PlayQueue.java
```diff
@@ -311,17 +311,7 @@ public synchronized void rearrange(int[] indexes) {
...
-        Collections.swap(files, index, index - 1);
-
-        if (this.index == index) {
-            this.index--;
-        } else if (this.index == index - 1) {
-            this.index++;
-        }
+        moveDown(index - 1);
```
randomnicode: Reentrant locks in Java, so this should be okay without any deadlock possibility.

logic looks okay to me

airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
eharris: This is used for year, tracknumber and disknumber, none of which really make sense to allow zero for.  But mostly I just left it with the same logic and limitations it originally had, which was that they didn't allow zero for those fields.

airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
eharris: For this pass I was only removing duplicated code.  Refactoring can be done in another PR if desired.

airsonic-main/src/main/java/org/airsonic/player/spring/RegisterPrecompiledJSPInitializer.java
eharris: I agree, try-with-resources would be better, but my objective this pass was only to remove duplicated code.  Refactoring can be done in another PR.

airsonic-main/src/main/java/org/airsonic/player/domain/PlayQueue.java
eharris: You're probably right, that would be better.  I just removed the duplicated code and preserved the original code path and am leaving refactoring/improvement to another PR.

airsonic-main/src/main/java/org/airsonic/player/spring/RegisterPrecompiledJSPInitializer.java
randomnicode: 👍

airsonic-main/src/main/java/org/airsonic/player/domain/PlayQueue.java
randomnicode: 👍

airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
tesshucom: It is better to keep the original.

Airsonic has several null avoidances that are not done in Subsonic.
(Width-Height when Media is VIDEO, PlayCount of Arrist, etc. Unfair null avoidance.)
The part that returns null in the original data specification is almost always an intentional specification.

airsonic-main/src/main/java/org/airsonic/player/service/search/SearchServiceUtilities.java
```diff
@@ -163,7 +163,7 @@ public final boolean addIgnoreNull(Collection<?> collection, IndexType indexType
     public final void addIfAnyMatch(SearchResult dist, IndexType subjectIndexType,
             Document subject) {
         int documentId = getId.apply(subject);
-        if (subjectIndexType == IndexType.ARTIST | subjectIndexType == IndexType.ALBUM
+        if (subjectIndexType == IndexType.ARTIST || subjectIndexType == IndexType.ALBUM
                 | subjectIndexType == IndexType.SONG) {
```
muff1nman: This or should be fixed too, right?

eharris: Good catch, missed that one.  Done.

Issue #1379, opened 2019-10-29
DLNA does not work with docker


Gt1ult:
Dear All,

I am using docker. Airsonic is installed using the networking in bridge mode port 4040. DLNA is not working. In the log the message "Enabling UPnP/DLNA media server" is reported. I looks like DLNA cannot be finalized. I would like to share my music to my yamaha media player.
I tried to use the networking in host as well, but nothing is changed.
Do you have any idea? Thank you in advance.

randomnicode:
Actually, just quickly glancing at it, I think the issue is that we're creating an ephemeral port for the DLNA service at startup that isn't exposed by the container (which seems to just expose 4040).

https://github.com/airsonic/airsonic/blob/master/airsonic-main/src/main/java/org/airsonic/player/service/UPnPService.java#L122

This is probably a bug. I wonder if anyone else has noticed this running inside docker? Are there any successful uses of DLNA from airsonic inside docker that might prove me wrong?

We should likely (a) fix a port for the DLNA so it's not ephemeral, and (b) expose it in the container

Gt1ult:
Hi Randomnicode, thank you so much for your help. I figured out how to do it. I am still learning and I am not so good.
So, I am using OpenMediaVault ,I set the networking to Hostand I typed the address http://192.168.2.100:4040/ in LAN Server address. It is working fine! Airsonic it is very cool. 
Now I have the problem to stream the music to my yamaha network player. I am trying to use jukebox but I am loosing :(

fxthomas:
Hey all, I'm doing a cleanup of the remaining issues: I believe this has been fixed in `master` and can be closed?

tesshucom:
It can be said ... that automatic allocation is no longer performed.
It will be silent feature change unless documented.
Will the description be added to the web page?

fxthomas:
Closing now, feel free to open a new issue if this still happens.

PR #1380, opened 2019-10-31
Fix DLNA/UPNP port, and expose them in Docker

randomnicode:
In response to and possible fix for https://github.com/airsonic/airsonic/issues/1379

- Fixes the listening port to 4041 for UPNP/DLNA (otherwise an ephemeral port is auto-assigned)
  - Can be customized by a JVM variable `UPNP_PORT`
    - E.g. `java -DUPNP_PORT=4042 -jar airsonic.war`
- Exposes tcp port 4041 in Dockerfile
- Exposes udp port 1900 in Dockerfile (needed for SSDP in UPnP)
- Removes an extraneous class

Note that if the default port is modified using `UPNP_PORT`, then user needs to expose and bind the custom port manually if running a container.


airsonic-main/src/main/java/org/airsonic/player/service/UPnPService.java
```diff
@@ -61,10 +62,17 @@
...
+    
     private UpnpService upnpService;
+    
     @Autowired
     @Qualifier("dispatchingContentDirectory")
     private CustomContentDirectory dispatchingContentDirectory;
+    
+    @Autowired
+    @Value("#{ systemProperties['UPNP_PORT'] ?: T(org.airsonic.player.service.SettingsService).DEFAULT_UPNP_PORT }")
+    private int port;
```
muff1nman: This doesn't match the existing methodology for managing settings. Should instead use a getter through settingservice.

randomnicode: `T(org.airsonic.player.service.SettingsService).DEFAULT_UPNP_PORT`

is the public variable in the SettingsService. do you want it to be a getter? There's no data encapsulation happening

muff1nman: Yes we should be consistent with all settings. In addition, we should be fetching from the settings service for the value, not from the environment.

randomnicode:
@muff1nman moved port to SettingsService

airsonic-main/src/main/java/org/airsonic/player/service/SettingsService.java
```diff
@@ -290,6 +292,10 @@ private static String getFileSystemAppName() {
     public static String getDefaultJDBCUrl() {
         return "jdbc:hsqldb:file:" + getAirsonicHome().getPath() + "/db/" + getFileSystemAppName();
     }
+    
+    public static int getDefaultUPnpPort() {
+        return Optional.ofNullable(System.getProperty("UPNP_PORT")).map(x -> Integer.parseInt(x)).orElse(DEFAULT_UPNP_PORT);
```
fxthomas: For consistency with all other settings, could you use the same kind of casing (e.g. `UpnpPort`) as the other keys and store it at the top of the file in a constant?

Issue #975, opened 2019-04-04
DNLA can not be used


tesshucom:
## Problem description

DNLA can be started without error.
However, the client can not search.

### Steps to reproduce

Check "Enable DLNA server" in DLNA / UPnP
and specify and save a valid "LAN Server address".
(Normal DLNA / UPnP startup operation)

## System information

Checked with common android apps.
xperia video & tomcat standalone at Windows10

0d78c30 last safety point
143a220 Many library updates at the same time. DWR Failure point.
8a1f36c DWR fix point
f04ec61 current

Problems occur in all revisions after 143a220.
DNLA starts without error but can not search from the client.

## Additional notes

cling-core has been updated in 143a220.

cling-core 2.0.1 -> 2.1.2
cling-support 2.0.1 -> 2.1.2
seamless-util 1.1.0 -> 1.1.2

Downgrading these did not improve the situation.
___

In addition to the above, the situation improved when the class dependency was corrected.
> org.airsonic.player.service.upnp.ApacheUpnpServiceConfiguration
> (Revert to the previous state)

I'm sorry.
It seems best to do better with Configuration, but I'm not familiar with DLNA / UPnP.

muff1nman:
Thanks for the PR - should be fixed now.

PR #1381, opened 2019-10-31
Upgrade cling UPnP/DLNA libs

randomnicode:
Upgrade cling libraries


tesshucom:
#975
There was a degrate like this before.
Can this series of PR(Changing the port etc ...) be considered an update that resolves these issues?


If so, I am very happy:blush:

randomnicode:
Hi @tesshucom thanks for pointing that out.
These PR series fix _a_ problem, though I don't know if it's _the_ problem.

I have no idea if that's the root cause because I don't know https://github.com/airsonic/airsonic/issues/975 was docker-related (which the port thing fixes)

Additionally https://github.com/airsonic/airsonic/issues/975 seems to indicate downgrading didn't help. Maybe @muff1nman can shed some light about what was happening?

muff1nman:
Yep already tried updating these dependencies once. I don't remember exactly what broke before and I'm not sure that it was port related.

Whatever it is, this will need some serious testing before this goes in.

tesshucom:
There are 3 classes of StreamClientConfigurationImpl implementation.
org.fourthline.cling.transport.impl, impl.apache, impl.jetty

Probably, the previous failure trigger was to replace to the apache configuration implementation.
apache was more strict with UDA, and certainly looked better if it could be replaced.
However, It returned to the original state without investigating because it was necessary to recover immediately.
Not only the settings, but also the library version was restored because we selected the safest option.

At that time, lib related to communication was updated at the same time with some trouble , so I felt that simultaneous verification was difficult.
This is the main reason for reverting.

Since randomnicode divides PR, I think that individual verification is possible.

@randomnicode
It may have been a side street about stream config, but the topic was that there was such a problem before.

fxthomas:
Well, this will probably need a rebase. I personally can't test this (I don't use DLNA/UPnP) so I'll let you all handle this :wink:

randomnicode:
Not maintining branches on this fork anymore so I'm just going to close this. Someone else can merge master onto these commits (or cherry-pick) and post a new PR if they wish

Issue #1373, opened 2019-10-24
Establish Release Channels


randomnicode:
I think we should establish Release Channels to help segregate development from releases.

This issue is intended to create a discussion around it.

My proposal is simple and very similar to what's already here:
- 3 channels: stable, beta, dev/master/edge
- Publish builds at every merge to master, or at least nightly. That's the edge channel.
- All dev happens on master. As such master is fluid and forever moving forward.
   - If people have issues with the name, then call it dev or something. 'master' is just a branch name, there's nothing special about it.
- From time to time, once builds are deemed stable enough, or we want to do a release, they are promoted to beta
   - Promotion could mean a few different things, and we can discuss what it should mean to us. I suspect for such a simple project, just tag a particular dev build and do a release (do we do releases manually here?).
- Once all issues with the beta have been resolved (with more dev on master and more promotions to beta), do a stable release. Tag it and release it.
- Out of band hotfix releases don't really mean very much for such a simple project, but say there is an issue with a previous API version (say 10.4) and master has moved on (say 11.4), create a branch for it off the previous stable release's (10.4) tag and do dev there. Do release from the new branch (assess if backport into master is necessary). The next hotfix to 10.4 stays on the 10.4 branch (to incorporate the previous hotfix). In short, hotfixes cause branches for stable releases.

Thoughts?

eharris:
I would be very happy if there were automated nightly development builds available for users to be able to more easily assist in testing the latest changes without having to have a build environment (and build know-how) themselves.

As far as adding a "beta" channel, if someone wants to take on managing that, I think it'd probably be ok.  I'm just not sure that we have the resources to afford splitting off and maintaining another branch on an ongoing basis without diverting needed resources from the main (master) branch.

randomnicode:
I guess my rationale about a beta channel are:
- It's not that much overhead. A beta channel is just a pre-release release.
- It provides a semi-state between the extremely-fast release channel and the extremely-slow release channel. Someone may wish to not be on the nightly build which might potentially have a bug or so (despite best efforts), but they also don't want to be on a release that was put out half a year ago. This allows that intermediate behavior. A somewhat stable build that we think is ready for a release, but we're "beta-testing it". If no problems surface, we do a release.
- This isn't a big project, but it allows for the possibility of "feature-freezing" a particular release, if we ever need it.

Regardless, we don't have to. We can always set it up later too.

randomnicode:
What is the actual release process currently?

[...]

PR #1383, opened 2019-10-31
Travis changes to build dev builds

randomnicode:
This PR automates dev builds to be pushed out to GitHub Releases and Docker Hub under the Edge Release Channel per discussion in https://github.com/airsonic/airsonic/issues/1373

I also think we should automate the Stable Releases, but that's a later priority. I'm eager to get the dev builds going so the massive backlog of PRs is reduced.

This PR:
- Divides builds into 2 stages: `test` and `deploy_edge`
- `deploy_edge` only runs
  - on `master`
  - for merges/pushes (i.e. no PRs)
  - If all jobs in the `test` stage pass
- Edge deploys
  - Build airsonic war
  - Build the docker container
  - Tag the commit it's building on (necessary for GitHub Release)
  - push to GitHub Releases with a Preleaase Build
  - push to Docker Hub
  - Tags are of the type `MAVEN_BUILD+DATETIMEOFBUILD+COMMITHASH` in keeping with Semantic Versioning
    - E.g. `10.5.0-SNAPSHOT+20191029183312-a9a7b082`
  - Deploys to both Docker Hub and GitHub Releases are done on the Edge Release Channel, and doesn't affect the Stable Release

This PR is INCOMPLETE and needs the following from the maintainer(s):
  - Generation of a GITHUB_OAUTH_TOKEN with appropriate privileges from the GitHub settings page of the repo. This is for Travis to use.
  - Encryption of the following strings with the travis public key for the repo, and insertion into appropriate place in the PR
    - `<GITHUB_OAUTH_TOKEN>` generated above
    - `DOCKER_USERNAME=<dockerusername>`
    - `DOCKER_PASSWORD=<dockerpassword>` 
  - For encryption, you may use the `travis` gem or do it manually with the travis public key. More reading here: https://docs.travis-ci.com/user/environment-variables/#defining-encrypted-variables-in-travisyml

Notes:
  - I considered adding automatic Changelog building commits so manual changelog updates aren't required, but we can add that later.
  - Stable releases are still manual for now. We need to see if this build works appropriately.
  - Can we please expedite this PR so we can actually start merging things?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,26 @@ language: java
 cache:
   directories:
     - $HOME/.m2
+
+services:
+  - docker
+
+env:
+  global:
+    - secure: "encrypted DOCKER_USERNAME=user"
+    - secure: "encrypted DOCKER_PASSWORD=pass"
+      
+stages:
+  - name: deploy_edge
+    # deploy only on commits to master branch
+    if: branch = master && type = push
+
 script:
   - mvn verify -B -P integration-test
 
 matrix:
   include:
+  # tests
   - name: coverity
     script:
       - echo "coverity scan script"
@@ -30,3 +45,32 @@ matrix:
   - jdk: openjdk11
   - jdk: openjdk12
   - jdk: openjdk13
+  
+  # deploys
+  - stage: deploy_edge
+    name: "Deploy Edge Release"
+    jdk: openjdk8
+  before_script:
+    - export PROJECT_MAVEN_VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
+    - export TRAVIS_TAG=${TRAVIS_TAG:-$PROJECT_MAVEN_VERSION+$(date +'%Y%m%d%H%M%S')-$TRAVIS_COMMIT}
+    script:
+      - mvn package -DskipTests -Dcheckstyle.skip=true
+      - docker build -f install/docker/Dockerfile -t airsonic/airsonic:latest -t airsonic/airsonic:edge/$TRAVIS_TAG .
+    before_deploy:
+      # Set up git user name and tag this commit
+      - git config --local user.name "Airsonic Travis"
+      - git config --local user.email "airsonic@tutanota.com"
+      - git tag $TRAVIS_TAG -m "Edge Release $TRAVIS_TAG"
+    deploy:
+    - provider: releases
+      api_key:
+        secure: "encrypted GITHUB_OAUTH_TOKEN"
+      file: airsonic-main/target/airsonic.war
+      skip_cleanup: true
+      options:
+        - prerelease: true
+        - name: "Edge Release $TRAVIS_TAG"
+      on:
+        tags: true
+    - provider: script
+      script: bash install/docker/dockerhub_push.sh
--- a/install/docker/dockerhub_push.sh
+++ b/install/docker/dockerhub_push.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
+docker push airsonic/airsonic:latest
+docker push airsonic/airsonic:edge/$TRAVIS_TAG


jvoisin:
You want to automatically do a pre-release for each commit on the master branch, or am I misunderstanding something?

randomnicode:
No, we do a release for each _push_ to master, not each _commit_.

This is exactly the developmental build history we need:. It'll allow us to pinpoint which regressions were introduced by which PR and which features were introduced by which PR.

That's literally what dev builds are. We're releasing them on the edge channel, and not the stable channel.

People can test drive a given build and test out their features.

muff1nman:
I've discussed this with @jvoisin in some of the travis related PRs, and will say it again. I'm not a fan of integrating our release process this heavily into travis.

Furthermore, I don't see a good idea to be pushing edge builds to docker hub. That seems like it would land us into trouble - users potentially mixing them up with release ones and also greatly inflate our docker repo storage impact.

randomnicode:
@muff1nman @jvoisin Mind putting links for these discussions here so I can apprise myself of these objections?

Without knowing any of these concerns, I respectfully disagree.

Stable builds are released once in a while, so I think I can still _kind of_ understand adhering to manual releases on them, though I'd just be confused about why you _wouldn't_ want to automate them, and save everyone including yourself the hassle. Regardless, this PR doesn't address them anyway. They're still manual. Do your Stable releases by hand, whenever you want to.

But Edge releases cannot be manual. That's just a stupendous amount of work for everyone for no good reason when an automated tool exists for exactly this purpose and you're using it for testing somehow, but refuse to use it for deployments. Any kind of manual work inherently leads to slowdown of the process, and will make people reticent to do a release. There is literally not one thing a person would be doing here that the build system isn't already doing and can't do with less mistakes once set up.

Regarding your second point, I don't particularly care nor use the docker for airsonic, so I can remove it, but arbitrarily picking it for release on one platform (GitHub Releases) and not another (Docker Hub) for a particular release channel just sounds inconsistent to me. Plenty of people would like to be able to use the edge development on docker, we're effectively forcing them to build their own containers, which isn't hard to do, but not sure why they should need to when it's literally just a few lines of code investment for us.

Additionally, I'm not buying the argument users would get mixed up, given _everything is tagged_. Have you ever been gone downloading Postgres 10 image and become confused and downloaded Postgres 8 instead? That cannot happen because you're going to use tags to distinguish between the releases and the channels.

And finally, what storage impact are we talking about? Docker Hub's? Docker Hub has no limits for image storage. Indeed, I would know. You can store "infinite number" of images. It's 2019, disk space is incredibly cheap, and you don't really need to worry about it. It's just the same as worrying about the number of VMs we're using each time anyone does a PR.

Let's do a simple back of the envelope calculation here for an estimation of our data footprint: how many images we'd have at most by now if we had had this from the very start? We're currently on PR 1393. The first one was in 2013. Assuming _everything_ was merged, in _six years_, you'd have built a backlog of _less than_ 1500 builds. That's practically nothing if we're going to talk about the number of bits here. People's Netflix footprint for a single day is probably more.

And let's say we try it and it gets unwieldy? We can literally delete them from the Hub and "reclaim our space"

Summarily:
- We have to automate Edge Releases. There's no good reason to stick to an archaic philosophy of doing manual deploys, not when releases are happening frequently, which they should. We're just increasing everyone's workload unnecessarily by not automating it. Everyone volunteers, no one has free time to be doing or be expected to do something manually that we can safely automate. That's just not respecting people's time.
- We should deploy them to Docker as well as GitHub Releases. There is no issue of disk space, nor is there an issue of people getting confused. I've never heard that argument. That's like saying people will get confused and download 10.4.0-RELEASE instead of 10.5.0-RELEASE because both exist in the docker repo. If you want to rename the tags to a different naming scheme, I can understand, but otherwise I have no idea where this concern originates from.

For reference, here's a sample Docker Hub repo for:
openjdk: https://hub.docker.com/_/openjdk?tab=tags
node: https://hub.docker.com/_/node?tab=tags&page=93

Want to count how many images they have? Every image for every build, for every platform

muff1nman:
Point taken on docker storage.

I agree a nightly/snapshot release process would be great to have - I never said it was a bad idea, but I'm very distrustful of travis and am reluctant to rely on it more.

Let me think on this a bit more.

randomnicode:
Hmmm, I'm unfamiliar with our history with travis. Why are we so distrustful of them? And if we are, why haven't we switched?

If this is about being hacked and leaked, then, first, I think that could happen to any site we use. Having said that, we need to look at the mitigation and we need the blast radius.

I think we need to give travis two pieces of sensitive data to make autodeployments work:
- A GitHub OAuth Token
- Docker Hub's User/Pass

The former isn't a big issue in my mind. We create a new one just for travis, restrict it to specific permissions on the repo that we need for deployment, and even if it's compromised, we can just revoke it.

The latter is a bigger problem since a compromised password is an issue. Fortunately, Docker seems to have just released capability for personal access tokens that we can use to authenticate instead of using our password: https://www.docker.com/blog/docker-hub-new-personal-access-tokens/

With the advent of that, we can just create and give travis a personal access token. Again, if it's compromised, not a big deal, we can just revoke it.

I think that covers our bases. All Travis has is access tokens from either site which can be just revoked and a new one introduced if compromised or using whatever key rotation strategy we want to employ.

randomnicode:
May we move forward with this one? How long is a PR supposed to sit before it ages like cheese?

I feel like this is the primary reason everyone hesitates merging on master. Moving forward on this will make folks less reticent to merge on master.

.travis.yml
```diff
@@ -30,3 +45,32 @@ matrix:
...
+  # deploys
+  - stage: deploy_edge
+    name: "Deploy Edge Release"
+    jdk: openjdk8
+  before_script:
+    - export PROJECT_MAVEN_VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
+    - export TRAVIS_TAG=${TRAVIS_TAG:-$PROJECT_MAVEN_VERSION+$(date +'%Y%m%d%H%M%S')-$TRAVIS_COMMIT}
+    script:
+      - mvn package -DskipTests -Dcheckstyle.skip=true
+      - docker build -f install/docker/Dockerfile -t airsonic/airsonic:latest -t airsonic/airsonic:edge/$TRAVIS_TAG .
```
muff1nman: Why not use the `docker` maven profile instead?

install/docker/dockerhub_push.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
+docker push airsonic/airsonic:latest
```
muff1nman: > users potentially mixing them up with release ones

I do not want to be pushing a latest tag.

.travis.yml
```diff
@@ -30,3 +45,32 @@ matrix:
...
+    before_deploy:
+      # Set up git user name and tag this commit
+      - git config --local user.name "Airsonic Travis"
+      - git config --local user.email "airsonic@tutanota.com"
+      - git tag $TRAVIS_TAG -m "Edge Release $TRAVIS_TAG"
+    deploy:
```
muff1nman: Would this not pollute our releases page? I don't like the idea of users having to sift through hundreds of "nightly" builds to find actual releases.

PR #1385, opened 2019-11-01
Massively Parallelize MediaScanner and thread safety

randomnicode:
The effective diff for this PR is: https://github.com/airsonic/airsonic/compare/04ae407ca3b8e09fffd02b7dc0d108c95462c2b9...b739108 (the last commit)
It is built on top of JSR310.

MediaScanner is currently a single-threaded low-priority operation full of object level pessimistic synchronization locking. This PR attempts to massively parallelize it and additionally alleviate some of the object locks to significantly boost throughput.

- Add thread safety for touching objects because now they are subject to massive number of concurrent threads
- Uses a custom forkjoin pool to fork tasks. This keeps the commonPool free and allows work-stealing to distribute load across all cores
  - Parallelism can be specified by environment variable `MEDIASCANNER_PARALLELISM`
  - Default parallelism is `number of core + 1`
- All FJP threads retain their low priority status, so it should have minimal impact on the system

I'll annotate the PR to help follow through. Feel free to ask questions or point out if you see an issue.

This PR should make a considerable bit of difference for slow media scans. My systems see a fairly large boost. Most systems are nowadays multi-core. We can likely optimize even more once Spring 2 and Hikari are merged because I suspect now the bottleneck will be the DB threadpool and we can tweak to increase that threadpool size.

The only caveat here is the DB connection. If it's single-threaded, then you're just effectively making it a single-threaded operation by bottlenecking it at the db connection level. You need a connection pool to fully see the benefits.

**Benchmarks:**
For a miniscule amount of data: (about 3295 entries that I could conjure up for a test bench) on a new internal db for both test running in serial on the same 6 core, 12 SMT (which would make parallelism 13 for the latter test by default), I see about an 800% performance boost, without any optimizations.
- Old single-threaded scan: ~50s
```
2019-11-06 17:39:35.440  INFO --- o.a.p.s.MediaScannerService              : Starting to scan media library.
...
2019-11-06 17:40:24.378  INFO --- o.a.p.s.MediaScannerService              : Completed media library scan.
2019-11-06 17:40:24.909  INFO --- o.a.p.service.PlaylistService            : Starting playlist import.
2019-11-06 17:40:24.909  INFO --- o.a.p.service.PlaylistService            : Completed playlist import.
```

- Newer multithreaded parallel scan: ~6s
```
2019-11-06 17:43:13.753  INFO --- o.a.p.s.MediaScannerService              : Starting to scan media library.
...
2019-11-06 17:43:18.065  INFO --- o.a.p.s.MediaScannerService              : Completed media library scan.
2019-11-06 17:43:20.393  INFO --- o.a.p.s.MediaScannerService              : Media library scan took 6s
2019-11-06 17:43:20.399  INFO --- o.a.p.service.PlaylistService            : Starting playlist import.
2019-11-06 17:43:20.399  INFO --- o.a.p.service.PlaylistService            : Completed playlist import.
```


airsonic-main/src/main/java/org/airsonic/player/domain/Album.java
```diff
@@ -31,12 +32,12 @@
...
     private String artist;
-    private int songCount;
-    private int durationSeconds;
+    private final AtomicInteger songCount = new AtomicInteger(0);
+    private final AtomicInteger durationSeconds = new AtomicInteger(0);
     private String coverArtPath;
     private Integer year;
     private String genre;
-    private int playCount;
+    private final AtomicInteger playCount = new AtomicInteger(0);
```
randomnicode: Switched to AtomicInteger for threadsafe atomic updates

airsonic-main/src/main/java/org/airsonic/player/domain/Album.java
```diff
@@ -143,11 +152,15 @@ public void setGenre(String genre) {
...
+        return playCount.get();
     }
 
     public void setPlayCount(int playCount) {
-        this.playCount = playCount;
+        this.playCount.set(playCount);
+    }
+    
+    public void incrementPlayCount() {
+        this.playCount.incrementAndGet();
```
randomnicode: Added for primary use case of the previously used setter.
For instance, `setPlayCount(getPlayCount() + 1)` is not atomic. Two threads may access getPlayCount() at the same time and increment by 1 each.

This allows us to update atomically without read-eval-write cycle which might not work with concurrent threads.

airsonic-main/src/main/java/org/airsonic/player/domain/Artist.java
```diff
@@ -73,11 +74,11 @@ public void setCoverArtPath(String coverArtPath) {
     }
 
     public int getAlbumCount() {
-        return albumCount;
+        return albumCount.get();
     }
 
     public void setAlbumCount(int albumCount) {
-        this.albumCount = albumCount;
+        this.albumCount.set(albumCount);
```
randomnicode: thread safety again

airsonic-main/src/main/java/org/airsonic/player/domain/Genre.java
```diff
@@ -28,36 +30,36 @@
 public class Genre {
 
     private final String name;
-    private int songCount;
-    private int albumCount;
+    private final AtomicInteger songCount = new AtomicInteger(0);
+    private final AtomicInteger albumCount = new AtomicInteger(0);
```
randomnicode: thread safety

airsonic-main/src/main/java/org/airsonic/player/domain/Genres.java
```diff
@@ -32,28 +32,17 @@
  */
 public class Genres {
 
-    private final Map<String, Genre> genres = new HashMap<String, Genre>();
```
randomnicode: not thread safe

airsonic-main/src/main/java/org/airsonic/player/domain/Genres.java
```diff
@@ -32,28 +32,17 @@
...
+    private final Map<String, Genre> genres = new ConcurrentHashMap<>();
```
randomnicode: Allows concurrent access without too much synchronization

airsonic-main/src/main/java/org/airsonic/player/domain/Genres.java
```diff
@@ -32,28 +32,17 @@
...
 
     public void incrementAlbumCount(String genreName) {
-        Genre genre = getOrCreateGenre(genreName);
-        genre.incrementAlbumCount();
```
randomnicode: not atomic

airsonic-main/src/main/java/org/airsonic/player/domain/Genres.java
```diff
@@ -32,28 +32,17 @@
...
+        genres.computeIfAbsent(genreName, k -> new Genre(genreName)).incrementAlbumCount();
```
randomnicode: Atomic and threadsafe for concurrent hashmap

airsonic-main/src/main/java/org/airsonic/player/domain/MediaLibraryStatistics.java
```diff
@@ -33,73 +36,73 @@
...
-    private int artistCount;
-    private int albumCount;
-    private int songCount;
-    private long totalLengthInBytes;
-    private long totalDurationInSeconds;
+    private AtomicInteger artistCount = new AtomicInteger(0);
+    private AtomicInteger albumCount = new AtomicInteger(0);
+    private AtomicInteger songCount = new AtomicInteger(0);
+    private AtomicLong totalLengthInBytes = new AtomicLong(0);
+    private AtomicLong totalDurationInSeconds = new AtomicLong(0);
```
randomnicode: thread safety

airsonic-main/src/main/java/org/airsonic/player/service/MediaFileService.java
```diff
@@ -703,7 +698,7 @@ public void incrementPlayCount(MediaFile file) {
         Album album = albumDao.getAlbum(file.getAlbumArtist(), file.getAlbumName());
         if (album != null) {
             album.setLastPlayed(now);
-            album.setPlayCount(album.getPlayCount() + 1);
+            album.incrementPlayCount();
```
randomnicode: switch to atomic operation

airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
```diff
@@ -110,9 +111,9 @@
     @Autowired
     private AlbumDao albumDao;
 
-    private EnumMap<IndexType, SearcherManager> searchers = new EnumMap<>(IndexType.class);
+    private Map<IndexType, SearcherManager> searchers = new ConcurrentHashMap<>();
```
randomnicode: Two options here,
- either synchronize the EnumMap, 
  - `Collections.synchronizedMap(new EnumMap<>(...));`
- or switch to ConcurrentHashMap
  - I chose this, because I think it will be more performant. We'll look the efficiency of the EnumMap, but will allow concurrent access without needing for too much locking

airsonic-main/src/test/java/org/airsonic/player/service/MediaScannerServiceTestCase.java
```diff
@@ -105,6 +106,15 @@ public Statement apply(Statement base, Description description) {
...
 
 
+    @Before
+    public void setup() {
+        while (mediaScannerService.isScanning()) {
+            try {
+                Thread.sleep(1000);
+            } catch (InterruptedException e) {}
+        }
+    }
```
randomnicode: This is needed because when Airsonic comes up, the MediaScannerService runs (because it's a "new" test installation). If a test runs a media scan of its own, it is just skipped depending on whether the previous media scan was finished. That's a jiggler waiting to happen

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -53,7 +66,7 @@
 
     private MediaLibraryStatistics statistics;
 
-    private boolean scanning;
+    private volatile boolean scanning;
```
randomnicode: set to volatile so all threads can see it from main memory (can also make it atomic, but isn't necessary)

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -71,7 +84,12 @@
     private ArtistDao artistDao;
     @Autowired
     private AlbumDao albumDao;
-    private int scanCount;
+    
+    @Autowired
+    @Value("#{ systemProperties['MEDIASCANNER_PARALLELISM'] ?: T(java.lang.Runtime).getRuntime().availableProcessors() + 1 }")
+    private int scannerParallelism;
```
randomnicode: Parallelization config option
- Default: Processor cores + 1
- Can customize with the environment variable `MEDIASCANNER_PARALLELISM`

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -71,7 +84,12 @@
...
+    
+    private AtomicInteger scanCount = new AtomicInteger(0);
```
randomnicode: This needs atomic updates

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -124,16 +142,26 @@ public synchronized void schedule() {
     /**
      * Returns whether the media library is currently being scanned.
      */
-    public synchronized boolean isScanning() {
```
randomnicode: no need for synchronized here, unnecessarily locks the entire object

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -124,16 +142,26 @@ public synchronized void schedule() {
...
+    
+    private ForkJoinWorkerThreadFactory mediaScannerThreadFactory = new ForkJoinWorkerThreadFactory() {
+        @Override           
+        public ForkJoinWorkerThread newThread(ForkJoinPool pool) {
+            final ForkJoinWorkerThread worker = ForkJoinPool.defaultForkJoinWorkerThreadFactory.newThread(pool);
+            worker.setName("MediaLibraryScanner-" + worker.getPoolIndex());
+            worker.setPriority(Thread.MIN_PRIORITY);
+            return worker;
+        }
+    };
```
randomnicode: The fork join thread factory for our pool. Just sets the name (so it's identifiable via debugging) and priority (we can consider making this configurable).

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -145,68 +173,76 @@ public synchronized void scanLibrary() {
...
-
-        thread.setPriority(Thread.MIN_PRIORITY);
-        thread.start();
+        ForkJoinPool pool = new ForkJoinPool(scannerParallelism, mediaScannerThreadFactory, null, true);
+        
+        CompletableFuture.runAsync(() -> doScanLibrary(), pool)
+                .thenRunAsync(() -> playlistService.importPlaylists(), pool)
+                .thenRunAsync(() -> mediaFileDao.checkpoint(), pool)
+                .thenRun(() -> pool.shutdown())
+                .thenRun(() -> scanning = false);
```
randomnicode: Do in order:
- initialize the pool
- scan library (do task within the pool)
  - this will spawn more tasks within the pool that other threads will start stealing and working on
  - this is where parallelization becomes most helpful
- import playlists (do task within the pool)
- consolidate db (do task within the pool)
- shutdown the pool (outside the pool, using the common pool)
- set the scanning variable (using common pool)

This is a synchronized call, which means only 1 thread can access it at a time. thus update to scanning variable here is guaranteed to be threadsafe operation

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -145,68 +173,76 @@ public synchronized void scanLibrary() {
...
         LOG.info("Starting to scan media library.");
-        Instant lastScanned = Instant.now().truncatedTo(ChronoUnit.SECONDS);
+        Instant lastScanned = Instant.now();
         LOG.debug("New last scan date is {}", lastScanned);
 
         try {
-
             // Maps from artist name to album count.
-            Map<String, Integer> albumCount = new HashMap<String, Integer>();
+            Map<String, AtomicInteger> albumCount = new ConcurrentHashMap<>();
```
randomnicode: Switch to concurrent access map for thread safety

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -145,68 +173,76 @@ public synchronized void scanLibrary() {
...
+            Map<String, Artist> artists = new ConcurrentHashMap<>();
+            Map<MediaFile, Album> albums = new ConcurrentHashMap<>();
```
randomnicode: Two additions so we don't keep going back and forth to the db, and possibly wreaking havoc on the threads.
Build it in memory first, then commit to db

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -145,68 +173,76 @@ public synchronized void scanLibrary() {
...
             Genres genres = new Genres();
 
-            scanCount = 0;
+            scanCount.set(0);
             statistics.reset();
 
             mediaFileService.setMemoryCacheEnabled(false);
             indexManager.startIndexing();
 
-            mediaFileService.clearMemoryCache();
```
randomnicode: Unnecessary call
`mediaFileService.setMemoryCacheEnabled(false);` already does it for us

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -145,68 +173,76 @@ public synchronized void scanLibrary() {
...
-
             // Recurse through all files on disk.
-            for (MusicFolder musicFolder : settingsService.getAllMusicFolders()) {
-                MediaFile root = mediaFileService.getMediaFile(musicFolder.getPath(), false);
-                scanFile(root, musicFolder, lastScanned, albumCount, genres, false);
-            }
```
randomnicode: This is in serial order
- Get the first MusicFolder
- scan it
- Get the next MusicFolder
- scan it
... rinse, repeat

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -145,68 +173,76 @@ public synchronized void scanLibrary() {
         }
         scanning = true;
 
-        Thread thread = new Thread("MediaLibraryScanner") {
-            @Override
-            public void run() {
-                doScanLibrary();
-                playlistService.importPlaylists();
-                mediaFileDao.checkpoint();
-            }
```
randomnicode: Replace a single thread with a fork join pool

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -145,68 +173,76 @@ public synchronized void scanLibrary() {
...
             // Recurse through all files on disk.
-            for (MusicFolder musicFolder : settingsService.getAllMusicFolders()) {
-                MediaFile root = mediaFileService.getMediaFile(musicFolder.getPath(), false);
-                scanFile(root, musicFolder, lastScanned, albumCount, genres, false);
-            }
-
+            ForkJoinTask.invokeAll(
+                    settingsService.getAllMusicFolders()
+                        .parallelStream()
+                        .map(musicFolder -> ForkJoinTask.adapt(() -> scanFile(mediaFileService.getMediaFile(musicFolder.getPath(), false), musicFolder, lastScanned, albumCount, artists, albums, genres, false)))
```
randomnicode: This is parallel.
Queues all the MusicFolders together and submits them to the pool
Thereafter this thread joins the pool workers itself until the work is finished and then returns.

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -145,68 +173,76 @@ public synchronized void scanLibrary() {
...
+                        .collect(Collectors.toList()));
+            
             // Scan podcast folder.
             File podcastFolder = new File(settingsService.getPodcastFolder());
             if (podcastFolder.exists()) {
-                scanFile(mediaFileService.getMediaFile(podcastFolder), new MusicFolder(podcastFolder, null, true, null),
-                         lastScanned, albumCount, genres, true);
+                ForkJoinTask.invokeAll(ForkJoinTask.adapt(() -> scanFile(mediaFileService.getMediaFile(podcastFolder), new MusicFolder(podcastFolder, null, true, null),
+                         lastScanned, albumCount, artists, albums, genres, true)));
```
randomnicode: This is a single task, and we _could_ just run it here by itself, this thread itself will work on
But just submit to the pool

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -145,68 +173,76 @@ public synchronized void scanLibrary() {
...
             }
 
             LOG.info("Scanned media library with " + scanCount + " entries.");
 
-            LOG.info("Marking non-present files.");
-            mediaFileDao.markNonPresent(lastScanned);
-            LOG.info("Marking non-present artists.");
-            artistDao.markNonPresent(lastScanned);
-            LOG.info("Marking non-present albums.");
-            albumDao.markNonPresent(lastScanned);
```
randomnicode: goes to the db serially for each of the tasks

we can just open 3 parallel connections to the db and concurrently execute them (they're independent)

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -145,68 +173,76 @@ public synchronized void scanLibrary() {
...
-
             // Update statistics
             statistics.incrementArtists(albumCount.size());
-            for (Integer albums : albumCount.values()) {
-                statistics.incrementAlbums(albums);
-            }
-
-            // Update genres
-            mediaFileDao.updateGenres(genres.getGenres());
+            statistics.incrementAlbums(albumCount.values().parallelStream().mapToInt(x -> x.get()).sum());
```
randomnicode: sum in parallel and increment in one go

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -145,68 +173,76 @@ public synchronized void scanLibrary() {
...
+            
+            ForkJoinTask.invokeAll(
+                    Streams.concat(
+                            albums.values().parallelStream().map(a -> ForkJoinTask.adapt(() -> albumDao.createOrUpdateAlbum(a))),
+                            artists.values().parallelStream().map(a -> ForkJoinTask.adapt(() -> artistDao.createOrUpdateArtist(a))),
```
randomnicode: Persist artists and albums first (we built an in-memory model and need to commit at the end)

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -145,68 +173,76 @@ public synchronized void scanLibrary() {
...
+                                artistDao.markNonPresent(lastScanned);
+                            })),
+                            Stream.of(ForkJoinTask.adapt(() -> {
+                                LOG.info("Marking non-present albums.");
+                                albumDao.markNonPresent(lastScanned);
+                            })),
+                            Stream.of(ForkJoinTask.adapt(() -> {
+                                LOG.info("Updating genres");
+                                mediaFileDao.updateGenres(genres.getGenres());
+                            }))
```
randomnicode: Update the rest of the stuff in parallel

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -218,14 +254,12 @@ private void doScanLibrary() {
...
-            scanning = false;
         }
     }
 
     private void scanFile(MediaFile file, MusicFolder musicFolder, Instant lastScanned,
-                          Map<String, Integer> albumCount, Genres genres, boolean isPodcast) {
-        scanCount++;
-        if (scanCount % 250 == 0) {
+                          Map<String, AtomicInteger> albumCount, Map<String, Artist> artists, Map<MediaFile, Album> albums, Genres genres, boolean isPodcast) {
+        if (scanCount.incrementAndGet() % 250 == 0) {
```
randomnicode: atomic update

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -240,23 +274,21 @@ private void scanFile(MediaFile file, MusicFolder musicFolder, Instant lastScann
...
-                scanFile(child, musicFolder, lastScanned, albumCount, genres, isPodcast);
-            }
-            for (MediaFile child : mediaFileService.getChildrenOf(file, false, true, false, false)) {
-                scanFile(child, musicFolder, lastScanned, albumCount, genres, isPodcast);
-            }
+            ForkJoinTask.invokeAll(
+                    mediaFileService.getChildrenOf(file, true, true, false, false)
+                        .parallelStream()
+                        .map(child -> ForkJoinTask.adapt(() -> scanFile(child, musicFolder, lastScanned, albumCount, artists, albums, genres, isPodcast)))
+                        .collect(Collectors.toList()));
```
randomnicode: fork into more tasks and submit more tasks to the pool (each child becomes another task)

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -240,23 +274,21 @@ private void scanFile(MediaFile file, MusicFolder musicFolder, Instant lastScann
...
         } else {
             if (!isPodcast) {
-                updateAlbum(file, musicFolder, lastScanned, albumCount);
-                updateArtist(file, musicFolder, lastScanned, albumCount);
+                updateAlbum(file, musicFolder, lastScanned, albumCount, albums);
+                updateArtist(file, musicFolder, lastScanned, albumCount, artists);
```
randomnicode: extra variables needed to build in-memory (committed at the end)

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -240,23 +274,21 @@ private void scanFile(MediaFile file, MusicFolder musicFolder, Instant lastScann
...
             }
             statistics.incrementSongs(1);
         }
 
         updateGenres(file, genres);
         mediaFileDao.markPresent(file.getPath(), lastScanned);
-        artistDao.markPresent(file.getAlbumArtist(), lastScanned);
```
randomnicode: in-memory model already holds the current status, no need to update db (only at the end)

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -279,20 +311,43 @@ else if (file.isAudio()) {
...
+            }
+            if (file.isAudio()) {
+                a.incrementSongCount();
+            }
+            
+            a.setLastScanned(lastScanned);
+            a.setPresent(true);
+            
+            return a;
+        });
```
randomnicode: This is guaranteed to be synchronized for the same object on the ConcurrentHashMap

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -279,20 +311,43 @@ else if (file.isAudio()) {
...
             return;
         }
+        
+        final AtomicBoolean firstEncounter = new AtomicBoolean(false);
+        Album album = albums.compute(file, (k,v) -> {
+            Album a = v;
+            
+            if (a == null) {
+                a = albumDao.getAlbumForFile(k);
+            }
```
randomnicode: only fetch once for the first thread to get here.
thereafter all threads will be able to get the in-memory model

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -307,63 +362,58 @@ private void updateAlbum(MediaFile file, MusicFolder musicFolder, Instant lastSc
...
-        }
-        if (file.getDurationSeconds() != null) {
-            album.setDurationSeconds(album.getDurationSeconds() + file.getDurationSeconds());
-        }
-        if (file.isAudio()) {
-            album.setSongCount(album.getSongCount() + 1);
-        }
-        album.setLastScanned(lastScanned);
-        album.setPresent(true);
-        albumDao.createOrUpdateAlbum(album);
```
randomnicode: this is done at the end now

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -307,63 +362,58 @@ private void updateAlbum(MediaFile file, MusicFolder musicFolder, Instant lastSc
...
-        if (firstEncounter) {
+            albumCount.computeIfAbsent(artist, k -> new AtomicInteger(0)).incrementAndGet();
```
randomnicode: atomic update

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -307,63 +362,58 @@ private void updateAlbum(MediaFile file, MusicFolder musicFolder, Instant lastSc
...
+            if (a == null) {
+                a = artistDao.getArtist(k);
+            }
+            
+            if (a == null) {
+                a = new Artist();
+                a.setName(k);
+            }
+            
+            Integer n = Math.max(Optional.ofNullable(albumCount.get(a.getName())).map(x -> x.get()).orElse(0), Optional.ofNullable(a.getAlbumCount()).orElse(0));
```
randomnicode: this is slightly thread-unsafe, but the important thing to note here is that albumCount should be monotonically increasing.

thus if another thread updates it to a higher number than what this thread is going to update it to (maybe this thread was preempted), we need to retain the higher number

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
```diff
@@ -307,63 +362,58 @@ private void updateAlbum(MediaFile file, MusicFolder musicFolder, Instant lastSc
...
+        
+        if (firstEncounter.get()) {
             artist.setFolderId(musicFolder.getId());
-        }
-        Integer n = albumCount.get(artist.getName());
-        artist.setAlbumCount(n == null ? 0 : n);
-
-        artist.setLastScanned(lastScanned);
-        artist.setPresent(true);
-        artistDao.createOrUpdateArtist(artist);
```
randomnicode: done at the end now

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
jvoisin: Can't this be `AtomicInteger(1);` instead of `AtomicInteger(0)).incrementAndGet();`?

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
randomnicode: yep

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
randomnicode: No. You're misplacing the parentheses on the `computeIfAbsent()` which give a hint about the effective purpose.
- `.computeIfAbsent(k -> new Value())`
  - if value is present for key `k`, returns it
  - if value isn't present, uses the function to create a new value

So `AtomicInteger(1)` will always set the albumCount to be 1. We don't want to set it to 1. We want to increment the _existing_ count by 1. If it already exists in the map, we retrieve it and increment by 1. If it doesn't exist, then _we create it_, and increment by 1.

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
fxthomas: Since we already have most of our settings configured via the `.properties` file, shouldn't we also use that here for consistency?

randomnicode:
I've noticed a few bugs/deficiencies in here and will correct them.
Most overtly:
- Albums in-memory map needs to have a different key. Otherwise we're not going to merge files in the same album together leading to uniqueness failures in the db when we try to write the same album for two different files to the db (twice).
- The order of operations needs to be a bit different. We can't parallelize Artists being written to the db while marking non-existing artists.
- Need to change hashcodes/equals for the MediaLibraryStatistics class due to Atomic* classes
- I think we can do better with marking media files presence.

Finally, we should switch to nio instead of keeping the old io based file structures. That'll be a separate PR for the full scope. I'll partially implement in the new commits.

randomnicode:
@jvoisin
Benchmark notes: (see description)

We're effectively hamstrung by absence of a connection pool to the DB until https://github.com/airsonic/airsonic/pull/1345 is merged. With that being said:

For a miniscule amount of data: (about 3295 entries that I could conjure up for a test bench) on a new db for both
- Old single-threaded scan: ~50s
```
2019-11-06 17:39:35.440  INFO --- o.a.p.s.MediaScannerService              : Starting to scan media library.
...
2019-11-06 17:40:24.378  INFO --- o.a.p.s.MediaScannerService              : Completed media library scan.
2019-11-06 17:40:24.909  INFO --- o.a.p.service.PlaylistService            : Starting playlist import.
2019-11-06 17:40:24.909  INFO --- o.a.p.service.PlaylistService            : Completed playlist import.
```

- Newer multithreaded parallel scan: ~6s
```
2019-11-06 17:43:13.753  INFO --- o.a.p.s.MediaScannerService              : Starting to scan media library.
...
2019-11-06 17:43:18.065  INFO --- o.a.p.s.MediaScannerService              : Completed media library scan.
2019-11-06 17:43:20.393  INFO --- o.a.p.s.MediaScannerService              : Media library scan took 6s
2019-11-06 17:43:20.399  INFO --- o.a.p.service.PlaylistService            : Starting playlist import.
2019-11-06 17:43:20.399  INFO --- o.a.p.service.PlaylistService            : Completed playlist import.
```

That's about a 10x performance boost while being hamstrung by a db connection pool.

airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
randomnicode: Thought `systemProperties` took `.properties` files into account, but maybe not. I'll move this to SettingsService just to be safe as you suggested

randomnicode:
Going to close this PR and switch to another repo

Tak-MK:
I've tried this merge and airsonic now scan absurdely faster than before, I hope this get merged soon!

randomnicode:
Not going to maintain this PR anymore (been about a month since it was opened, and it seems to have conflicts now).

I've forked this repo pending discussion in #1285, and this PR (and its additions for other benefits) have been merged here: https://github.com/airsonic-advanced/airsonic-advanced (the successor to this particular PR is: https://github.com/airsonic-advanced/airsonic-advanced/pull/20).

The GitHub release for the succesor to this PR's merge is here: 
- https://github.com/airsonic-advanced/airsonic-advanced/releases/tag/10.6.0-SNAPSHOT.20191122081836
- Or if you want the latest:  https://github.com/airsonic-advanced/airsonic-advanced/releases

The docker image for the succesor to this PR's merge is here: 
- `docker pull airsonicadvanced/airsonic-advanced:gitcommit-621ff1f764120b768c76043c0ab1b8ddec5d2c341`
And if you want the latest:
- `docker pull airsonicadvanced/airsonic-advanced:edge-latest`
- Peruse the images: https://hub.docker.com/r/airsonicadvanced/airsonic-advanced

Tak-MK:
Yup, I've compiled the .war from your repo on master branch, and it scans too fast, but after it finishes, airsonic got stuck and didn't respond. I've tried a few times but it still happens. If you want I can open an issue in your repo.
Also I don't understand about developing that much so I don't know which issue arise on that discussion, I suppose then this merge won't join the airsonic master branch?
Thanks a lot!

randomnicode:
yep, best to open the issue on the other fork

PR #1377, opened 2019-10-27
Convert to JSR310

randomnicode:
A relatively simple refactor to change from legacy Date/Calendar classes to the modern JSR 310 java.time.* classes etc.

The number of files touched is pretty large, but most changes are actually simple drop-in replacements (please review)

PR is in two commits:
- First swaps the instances/logic of Date, Calendar, DateFormats with the java time pseudo equivalents
- Second deals with the JDBC mapping

The commit breakdown is deliberately such so as to allow us to revert the second commit once we get a JDBC 4.2 compliant driver and/or if we decide to use some aspects of an ORM.

Notes:
- Had to leave some instances of Date in there due to upstream method contracts
- Chose Instant over other Java Time classes as our primary replacement. Here are my reasons I'm choosing to write here for posterity:
  - We currently store data in the DB as `timestamp without timezone`, effectively epoch. We use the `Date` Java class which internally holds epoch elapsed.
  - Can't use `LocalDateTime`: Time will jump around if the server moves from place to place. You might have 8am for one timestamp and then transport the server (or migrate the db) to another timezone and get another 8am timestamp because `LocalDateTime.now()` will only give the local time. Additionally not backwards compatible with the current `Date` classes.
  - Can't use `OffsetDateTime`: We won't retain the timezone info and lose it without changing the columns to record timezones also (`timestamp with timezone`). I suspect it'll effectively become UTC anyway.
  - Thus we should use Instant and change it to whatever OffsetDateTime we want when we need it given the local timezone info. The data in the DB is always UTC

PR #1386, opened 2019-11-01
Use less batching for certain SQL queries

randomnicode:
Effective diff: https://github.com/airsonic/airsonic/compare/ffd229e07d5387c0477de883478d0bbd6f1867e7...c2e3f0a

Some queries are doing things in batches of 1000, which means multiple connections are being opened to the DB for no apparent reason. These are costing multiple round-trip times, when a single query should suffice.

This replaces these batch loops with a single query where needed.


jvoisin:
Why the fuck where those queries done in a `for` loop in the first place :/
The effective diff looks good, trivial and useful to me, but I guess that we'll have to merge #1377 first.

randomnicode:
Don't know. Maybe Sindre thought it'd be faster with a file-based DB

fxthomas:
Looks good to me too.

Minor nitpick: can we get away with removing the min/max id checks? They appear useless for what we're doing here.

randomnicode:
Good point, done here: https://github.com/airsonic-advanced/airsonic-advanced/pull/28 with an additional commit, feel free to backport with the additional commit.

tesshucom:
I also noticed when I copied it, but left it as it was.

Because it predicted the existence of concerns pointed out in https://github.com/airsonic-advanced/airsonic-advanced/pull/28.

In fact, the optimal solution depends on the target DB implementation.

tesshucom:
I also noticed when I copied it (to the expunge method), but left it as it was.

The code reminds me of some concerns.
This concern seems to have already been explained at https://github.com/airsonic-advanced/airsonic-advanced/pull/28.

In fact, the optimal solution depends on the target DB implementation and amount of data.

randomnicode:
I agree with concerns voiced by @eharris and @tesshucom. However, the current implementation is too inflexible and thus slow for most people. The correct implementation is, as @tesshucom pointed out, data dependent, with this PR being the correct implementation for large to small libraries and a variant of the existing implementation as the correct thing to do for massive libraries. (Additionally, type of db also likely matters.)

I concur with the approach outlined by @eharris that we should make this configurable so folks are able to play around to find the "right" amount of aggressiveness for their own use case with some default built in to suit most folks. I'll likely post a followup PR if someone doesn't beat me to it.

randomnicode:
Follow up (backend + UI) posted here: https://github.com/airsonic-advanced/airsonic-advanced/pull/31
Along with some useful indices here: https://github.com/airsonic-advanced/airsonic-advanced/pull/32

Feel free to backport

PR #1387, opened 2019-11-01
Fix test queries for limits

randomnicode:
We're using a limit of 0 for some test queries and expecting all results, which might work for HSQLDB but is incorrect by ANSI standards. Things will break if we test against other compliant DBs.

Production code seems to put the limits correctly

This affixes the limit in the test queries to be more in line with what the production code is doing (setting it to max).


randomnicode:
Yeah, probably quirks of HSQLDB or the prepared statement is doing something funky with variables that shouldn't be there

PR #1389, opened 2019-11-02
[WIP] Move from maven to gradle

jooola:
Maven is slow, dirty and not handy, so I suggest to use Gradle as build tool.

This PR will handle the tools and documentation migration.

@airsonic/maintainers Is anyone against this change ?

I'll extract the sonos wsdl upgrade in an other bigger PR since the upgrade isn't completed yet.


jvoisin:
Since I don't know much about java nor its ecosystem, I don't have a strong oponion about this, however:

- We shouldn't bundle .jar files
- Why do we need a .bar and a .sh file?

jooola:
> We shouldn't bundle .jar files

Agreed, but here it the exception, the gradlew (gradle wrapper) scripts bind with this jar, and allow us to download, upgrade or do whatever you want with the gradle tool. With no system dependencies.
 
> Why do we need a .bar and a .sh file?

Bat is for Windows users, sh is for Unix users.

jvoisin:
I already have my operating system's package manager to install/mange/update build tools, shipping a .jar in git sounds super-wrong

jooola:
The jar is ~ 53kb. so its not that big.

And I somehow disagree, some tools cannot always be delivered by package manager. This seems to be such kind of tools. Still this is my first time with gradle, lets see if some pro have some advice regarding this.

fxthomas:
I did use Gradle for Android dev before, and they have this practice of bundling bootstrapper scripts that let us build things without a lot of dependencies. A bit like the `ensurepip` module in Python?

It usually works fairly well, especially since each Java project would have its own dependencies anyway, without relying on the local package manager.

jooola:
Ensurepip is the same idea. What I find handy is the set of features gradle comes with, like multi project management. Also that we can code in groovy within the build.gradle files.

muff1nman:
I am vehemently against this.

> Maven is slow, dirty and not handy
Disagreed. It is declarative, fast and there is no reason to add gradle's procedural nonsense.

muff1nman:
https://phauer.com/2018/moving-back-from-gradle-to-maven/

The article above shares many of my concerns.

jooola:
Ok I get all your points.

Was a nice run discovering Gradle though !

tesshucom:
Oh, it was closed.
2 points from experience.

 - Gradle exposed many bugs when migrating to JDK11 (contrary to image, weaker to the front line than Maven).
 - [Github security alert](https://help.github.com/en/github/visualizing-repository-data-with-graphs/listing-the-packages-that-a-repository-depends-on) 
build.gradle is not supported.

PR #1390, opened 2019-11-03
Update Sonos wsdl file

jooola:
Updated Sonos wsdl file from https://musicpartners.sonos.com/sites/default/files/Sonos.wsdl


jvoisin:
Marking every new method return null is something we're ok with?

jooola:
I didn't think much about this, I saw some other methods returning null, or doing nothing. So I did the same with the new methods. I guess while we are not using them its fine. Not sexy but fine.

jooola:
Ok just needs a little review

fxthomas:
This looks okay to me, it does not touch a lot of things and appears clean. I don't have any way of testing it since I don't use Sonos, but let's merge it?

PR #1400, opened 2019-11-08
Add some space below album list

pacija:
Trying to click "more" in "Random" album view, or ">" next to"Albums X-X" in "Recently added" quite often raises up play queue which pops over and hides the links. Adding a few em's of space below makes sure play queue don't accidentally pop up.

### DIFF
--- a/airsonic-main/src/main/webapp/style/default-without-mediaelement.css
+++ b/airsonic-main/src/main/webapp/style/default-without-mediaelement.css
@@ -155,6 +155,7 @@ body {
     padding:0;
     border:0;
     margin:0.75em;
+    padding-bottom: 5em;
     font-family: Verdana, Tahoma, Geneva, arial, sans-serif;
     font-size: 10.5pt;
     line-height: 1.5em;


jvoisin:
Can you please provide a before/after screenshot?

pacija:
There are two screenshots, named "before" and "after" at the following link:

https://oblak.mimar.rs/index.php/s/rt9jZMNaGtNdFtj

Notice there is more space below "Albums 1-40"

almet:
Here are the screenshot inline (before and after):
before
after

almet:
I've had this issue also. I wonder if another way to fix this would be to put a sort of timer, so that the play queue doesn't pops up instantly, but instead waits a bit to be sure to get the focus?

Or, maybe directly remove this behavior, making the user click in order to display this play queue. What do you think?

fxthomas:
Hmm, the hit box for the play queue seems a bit large, sure. I don't often see it as I generally have auto-hide disabled in the preferences.

I don't like adding space like this to the UI, how about changing this piece of code (in `playQueue.jsp`) instead so that the play queue toggle does not fire if you are too far from the bottom border?

```javascript
$(window).mouseleave(function (event) {
    if (event.clientY < 30) onHidePlayQueue();
});
$(window).mouseenter(function () {
    onShowPlayQueue();
});
```

@almet I also created #1421 with a quick design idea for a manual show/hide button, thanks for the idea!

eharris:
I kind of like the jsp solution a bit better.

PR #1406, opened 2019-11-13
Allow access to real alsa hardware by systemd

Whisprin:
Add `DeviceAllow=char-alsa rw` and disable `PrivateDevices` to allow access to alsa hardware

### DIFF
--- a/contrib/airsonic.service
+++ b/contrib/airsonic.service
@@ -26,7 +26,6 @@ Group=airsonic
 DevicePolicy=closed
 DeviceAllow=char-alsa rw
 NoNewPrivileges=yes
-#PrivateDevices=yes
 PrivateTmp=yes
 PrivateUsers=yes
 ProtectControlGroups=yes
@@ -37,6 +36,10 @@ RestrictNamespaces=yes
 RestrictRealtime=yes
 SystemCallFilter=~@clock @debug @module @mount @obsolete @privileged @reboot @setuid @swap
 ReadWritePaths=/var/airsonic
+
+# You can uncomment the following line if you're not using the jukebox
+# This will prevent airsonic from accessing any real (physical) devices
+#PrivateDevices=yes
 
 # You can change the following line to `strict` instead of `full`
 # if you don't want airsonic to be able to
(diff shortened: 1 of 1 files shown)


contrib/airsonic.service
```diff
@@ -24,8 +24,9 @@ Group=airsonic
 # See https://www.freedesktop.org/software/systemd/man/systemd.exec.html
 # for details
 DevicePolicy=closed
+DeviceAllow=char-alsa rw
 NoNewPrivileges=yes
-PrivateDevices=yes
+#PrivateDevices=yes
```
almet: If we were to remove this, should we directly remove it, rather than commenting-it-out?

almet:
For whoever is interested, here is the systemd doc about private devices: https://www.freedesktop.org/software/systemd/man/systemd.exec.html#PrivateDevices=

contrib/airsonic.service
Whisprin: Maybe add a comment stating that this is required for Jukebox only, similar to line 46-48: #ProtectHome=true

Whisprin:
Opened an issue for further discussion. With the changes proposed here I'm able to select an audio device from the Admin UI. However, it's still not possible to play music.

contrib/airsonic.service
muff1nman: Yes a comment would be good and moved into a separate section like `ProtectHome`

fxthomas:
I think we should:
* Setup the systemd unit for all features by default
* Comment on why these restrictions were lifted (so that interested people can re-enable them if they don't use the associated feature)

Also, what do you mean by "jukebox is a bit crashy"? Is this something that could be split in a separate issue, or is that related to the device restrictions?

Whisprin:
> Also, what do you mean by "jukebox is a bit crashy"? Is this something that could be split in a separate issue, or is that related to the device restrictions?

Yes, propably something that would go into a separate issue. I'm trying to make it reproducible and collect logs. In short: I use DSub to connect to the jukebox, music is playing, switching songs works - however, when I disconnect DSub and reconnect later: nothing is playing anymore
The only way to recover from this state is to restart Airsonic

contrib/airsonic.service
muff1nman: @Whisprin can you update your PR?

Whisprin: @muff1nman Updated the PR accordingly, sorry for the delay

PR #1432, opened 2019-12-05
Fix playback not starting after adding tracks to an empty play queue

### LINKED ISSUE #1407: Play button at bottom of window won't start playback of Play Queue
## Problem description

Using Web Player -- After adding tracks to the Play Queue, clicking the play button below the Play Queue does nothing.

## System information

 * **Airsonic version**: *10.5.0*
 * **Operating system**: *Linux Mint 19*
 * **Java version**: *openjdk 11.0.4, Tomcat/8.5.34*
* **Client**: *tested with Firefox 70.0.1, Vivaldi 2.5.1525.41, Chrome 78.0.3904.97*

fxthomas:
Should fix #1407. To reproduce:

* Start Airsonic 10.5.0.
* Empty the play queue.
* Click on the "+" button on an album.
* Click on the "play" button on the web player.
* Playback doesn't start.

### DIFF
--- a/airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
+++ b/airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
@@ -41,6 +41,9 @@
     // Stream URL of the media being played
     var currentStreamUrl = null;
 
+    // Current song being played
+    var currentSong = null;
+
     // Is autorepeat enabled?
     var repeatEnabled = false;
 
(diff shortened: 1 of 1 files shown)


muff1nman:
Hmm, while I'm sure this fixes the issue, something seems a bit off about interacting with mejs in the playqueue callback function. Ideally this function should be abstracted away from the inner workings of mejs.

Would it make sense to refactor the `loadMediaElementPlayer` and `skip` functions such that we can utilize those paradigms here? In particular I'm imagining you'll want to utilize most of the logic in skip except for the actual start of the song. Does that make sense?

fxthomas:
I'll... try to get something working in that direction. Probably not before a few weeks though, given that it's Christmas and all :wink:

fxthomas:
I pushed a new commit with something I came up with, but... I'm not convinced this is better. Let me know what you think.

fxthomas:
Actually, scratch that, it makes the rest of the playback not work correctly. You can still have a look at b4fc1a3, but I reverted it.

fxthomas:
With 6c66c8e it should work fine. Take your time reviewing/testing it, especially if you can try non-default player types (Cast, external). I'd rather not have the media player break again because of this :wink:

fxthomas:
It's been working fine for me for months now, I'll merge that soon.

PR #1433, opened 2019-12-05
Add 'add to play queue' buttons to the playlist page

### LINKED ISSUE #1226: [Feature Request] Add/append playlist to queue
Playlists should have an "add to queue" ("+") button like albums. Currently, there is no way to add a playlist to the queue, and then append another playlist to the queue. This can be done in the Subsonic Android app, for example. The only way it can be done in Airsonic is to click Play to add the first list, and then add each song from another playlist one-by-one using the "+" button.

fxthomas:
This should fix #1226.

image

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/ajax/PlayQueueService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/ajax/PlayQueueService.java
@@ -253,6 +253,33 @@ public class PlayQueueService {
     /**
      * @param index Start playing at this index, or play whole playlist if {@code null}.
      */
+    public PlayQueueInfo addPlaylist(int id, Integer index) throws Exception {
+        HttpServletRequest request = WebContextFactory.get().getHttpServletRequest();
+        HttpServletResponse response = WebContextFactory.get().getHttpServletResponse();
+
+        String username = securityService.getCurrentUsername(request);
+        boolean queueFollowingSongs = settingsService.getUserSettings(username).isQueueFollowingSongs();
+
+        List<MediaFile> files = playlistService.getFilesInPlaylist(id, true);
+        if (!files.isEmpty() && index != null) {
+            if (queueFollowingSongs) {
+                files = files.subList(index, files.size());
+            } else {
+                files = Arrays.asList(files.get(index));
+            }
+        }
+
+        // Remove non-present files
+        files.removeIf(file -> !file.isPresent());
+
+        // Add to the play queue
+        int[] ids = files.stream().mapToInt(f -> f.getId()).toArray();
+        return doAdd(request, response, ids, index);
+    }
+
+    /**
+     * @param index Start playing at this index, or play whole playlist if {@code null}.
+     */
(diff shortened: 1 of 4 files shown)


randomnicode:
Sorry I had intended to post a comment on this PR earlier and the reason it was rejected from the `airsonic-advanced` fork and not ported, but didn't get around to it until now.

I think this PR should be reverted on this fork, or fixed. The central method in the PR is erroneous and the only reason it hasn't been spotted is either no one really reviewed the PR, or the UI is masking the actual error in the method because it just sends nulls as an argument. Anytime you send a non null argument into the method, the error will become obvious.

airsonic-main/src/main/java/org/airsonic/player/ajax/PlayQueueService.java
```diff
@@ -250,6 +250,33 @@ public PlayQueueInfo playInternetRadio(int id, Integer index) throws Exception {
...
+            }
+        }
+
+        // Remove non-present files
+        files.removeIf(file -> !file.isPresent());
+
+        // Add to the play queue
+        int[] ids = files.stream().mapToInt(f -> f.getId()).toArray();
+        return doAdd(request, response, ids, index);
+    }
```
randomnicode: This method seems incorrect. It's not doing what it should be doing. You haven't discovered the issue yet because you're not using the arguments from DWR properly (it always passes in a null for an index).

The issues are:
- I think you've confused `index` and used the same variable for two separate tasks
  - There are two indices in question here
    - One passed into the method itself  `addPlaylist(id, index)`: This `index` implies which part of the _playlist_ to add. If you pass in 0, it should select the entire playlist to add. If you pass in 20, it'll add the songs in the playlist after index 20. And so on. You can see that from the `sublist()` bit you're doing
    - One passed into the method `doAdd(request, response, ids, index)` which you call at the end to do the actual adding into the _playqueue_. This `index` represents where in the _playqueue_ you want to add the given files.
  - You pass the same `index` to both places. Imagine if index is 10. You're telling the function to select the playlist's songs starting at index 10 _and_ also add it into the playqueue at its own index 10. These (_playqueue_ and _playlist_) are two separate entities. Each's index has nothing to do with the other and yet are being passed around interchangeably. NPE will be the most frequent error, though you'll run into plenty more if your terms are confused.
- I'm not sure why that whole bit about `queueFollowing` is needed. Looks like copy-paste error from another function. You don't need it. You're appending the playlist in its entirety or its sublist. There isn't a question of whether to add followup songs. The playlist itself has all the songs.
- The method signature implies that you can select sublists of the playlist. But your UI specifically says "Add All", and never sends any index (always a null), which means this method never actually is using the index. My suggestion is to just remove the index here. You can add the entire playlist or nothing at all. When you come up with another UI design to add sublists, you can expand it later (YAGNI)

fxthomas: Good catch, I'll send a new PR for that!

Issue #1391, opened 2019-11-03
Add code formatting tool


jooola:
I would be nice to add a code formatting tool in this projectn for IDE's and CI.

The question is which on should we use.
Here are some suggestion:
https://github.com/google/google-java-format
https://github.com/diffplug/spotless
https://github.com/jhipster/prettier-java

I would go for Google plugin if we want to start a code formatting dictatorship. But I'm good with this.

Prettier use nodejs, so this might not be a great idea.

jvoisin:
We already have  checkstyle #1121

jooola:
Does checkstyle format the files ? I mean it seems more like a linting tool, not a formatting tool.
I purely speaking code readability. Both could be combined though ?

jvoisin:
Both should be combined, yes :)

tesshucom:
Checkstyle and formatter are the foundations of joint development.
The most popular formatter-maven-plugin doesn't seem to be listed, but anything that has equivalent functionality is fine.

 - Generate configuration xml using Eclipse settings GUI.
 - We can use the xml with Maven
 - We can use the xml with IntelliJ

randomnicode:
Hmmm, what would the workflow be?

Who's going to commit the changes?

If travis is doing the formatting and committing changes during a build, then we need to get travis some creds. Additionally that runs the risk of causing some merge conflicts especially on stacked PRs. Consider:
- You build branch1 with changes to File 1
- You build branch2 on top of branch1 with more changes to File 1
- You submit both as stacked PRs
- You merge PR1 (on branch1) to master
- Travis does a formatting commit on master for File 1
- Now due to the formatting commit being on the same file, PR2 (on branch 2) has merge conflicts even though it was stacked on top of PR1 and there were no changes made by anyone else in the middle.

If the devs are going to do it prior to a commit, then we don't need an external formatter plugin. About every IDE comes with a formatter. Just need to import a formatting guideline export into the IDE.

I'm not against the idea, maybe I just don't understand it?

Also, just my opinion, but I do find our checkstyle to be too restrictive.

tesshucom:
We can modify thousands of code by simply testing existing features and fixing problems.
Even if someone does it locally, those who can predict what will happen in the review do not want to provide feedback upstream.

CheckStyle, Formatter, Findbugs and PMD are common enough tools that has existed since ancient times.
There are many tiny probrems that have been pointed out in reviews, that can be detected by tools.
[...]

PR #1434, opened 2019-12-06
Added more checkstyles

eharris:
Added checkstyles for unnecessary whitespace and brace placement


tesshucom:
Settings for white space, braces, and line breaks are the basic settings that should be set first.
And it is usually introduced in pairs with the formatter.
Because it is related to productivity deeply.

Before the merge, can the maintainer's common opinion (Airsonic policy) be expressed to the committer in #1391?
Such policy statements are often important in project management.
Because for third-party committers it is one of the criteria for determining whether it is worth spending time PRing the project .

tesshucom:
The question may have been ambiguous.

I do not object to the contents of CheckStyle at all.
The focus is here.

> And it is usually introduced in pairs with the formatter.

The purpose of coding rules are not about fulfilling aesthetics, but ultimately to improving productivity.
If the method is mistaken, the opposite effect is produced.

Therefore, 

 - please clarify whether you are willing to prepare a formatter or not.

Expecting external support in this case is only a procrastination of the problem.
(It is impractical to envisage external collaborators who spend time on such basic things for project.)
Or, if you ’re already working strategically based on some plan, it ’s best to provide a roadmap.

fxthomas:
I will not add a formatter tool myself.

If someone wants to create a PR, we can help them integrate in-progress things as we go. How does that sound?

tesshucom:
Although it may have been written somewhere, it seems to be a topic with a considerable difference in personal experience.
Therefore, I respect any stance that Airsonic maintainers do.
And it is a negative legacy and the responsibility is not the current maintainer.
I fully understand this.

In my sense, there are no projects that do not maintain checkstyle and formatter, especially in the business scene.
Even beginners understand how it works, and knowledge of setup is essential for intermediate and higher players who are at the core of the project.

Coding topics are rarely featured during reviews.
If this happens frequently, you can think of it as the responsibility of the senior engineer who failed to control tools, not the submitter of the code.
(What is the maintainer for maintaining?)

> I will not add a formatter tool myself.

If you think that maintaining the formatter is annoying, that's fine.
Just label ``wontfix`` and close #1391.

However, judging objectively, the current situation is not favorable.
If you just do a smallfix, you will not be bothered, but it will be bothersome to modify thousands of lines, make extensive modifications, or add new functions.
Although not statistically visible, in the long run, there will be fewer third party comitters  proposing thousands of lines of corrections or feature additions.

If care is not taken to reduce the burden of PR, they only fix bugs locally and develop new features.
There is no merit of wasting time for PR, for them.

In other words, I'm thinking about the future rather than the little trouble at the time of the last commit.
Whether Airsonic intends to accept  change and evolution (in a good sense) and prepares for it.
Or vice versa.

fxthomas:
If I understand correctly then, you think we would gain a lot by adding a formatter? In any case I think it should be done in another PR.

> If you think that maintaining the formatter is annoying, that's fine. Just label wontfix and close #1391.

I don't think it's annoying, far from it; I just don't have enough experience with Java to clearly see the benefits, and I've only used a linter, never a formatter in other languages I know more about : I would rather have somebody else who knows about this propose a good solution there, rather than adding a broken tool that becomes a burden to committers.

Airsonic as I see it doesn't have a Product Owners, Project Managers or even a single Maintainer. It's a community project, and this means that ideas or issues are discussed publicly as much as possible. Plus I'm relatively new to the team, so I don't want to close an idea unless there is a strong decision that we will _never_ do it nor accept patches. Here it's clearly not the case.

Does that make sense?

muff1nman:
Let's move discussion of a formatter tool over to #1391. Until #1391 is resolved, the checkstyle plugin is the current strategy.

PR #1446, opened 2019-12-24
Bump javax' servlet version

jvoisin:
Apparently, there is an XXE in the version that we're using (1.1.2)


muff1nman:
Changing a runtime dependency version like this is probably not a good idea. Its better to update the version of whatever provides the dependency.

PR #1372, opened 2019-10-24
Spring Boot 2 Upgrade: Part 6 - Upgrade to Spring 2.2.0

randomnicode:
This builds on #1364, #1367, #1368, #1369 and #1370, and upgrades Spring Boot to Spring Boot 2.2.0-RELEASE along with all dependencies.

~FAILURE NOTE: This PR is expected to fail until https://github.com/airsonic/airsonic/pull/1358 is merged. Spring Boot 2 needs those migrations to produce the correct SQL in HSQLDB.~ Should pass now pending discussion in https://github.com/airsonic/airsonic/pull/1358 (issue with HSQLDB driver 1.8.0.10 which has been reverted)

Note that as of Spring 2.2.0, Spring Boot has migrated its Java EE dependencies with `javax` group IDs to the equivalent Jakarta EE dependencies with all the old dependencies immediately deprecated. This commit reflects that change. More may be found on their Release Notes: https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.2-Release-Notes

I noticed subjective speed up with this upgrade, which is what Spring Boot team is trying to accomplish. Exact metrics are out of scope, but one of my machine claims the boot time reduced from ~35 sec to ~8 sec consistently. YMMV.

This is Step 6 of a series of PRs each of which will build on each other
- GitHub shows the entire list of changes (including previous PRs) until previous PRs in the series are merged. This PR's diff from previous PR #1370 is the effective list of changes in it (5 commits)

This effectively concludes the series for now. We need to migrate the tests to JUnit 5 to stop using the vintage engine and instead switch to jupiter.

Issue #69, opened 2016-05-12
Passwords Stored Insecurely 


issuemover631:
<a href="https://github.com/alexfornuto"><img src="https://avatars0.githubusercontent.com/u/2349184?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [alexfornuto](https://github.com/alexfornuto)**
_Thursday May 12, 2016 at 15:04 GMT_
_Originally opened as https://github.com/Libresonic/libresonic/issues/69_

----

Currently Subsonic (and thus LibreSonic) stores user names and passwords in `/var/subsonic/db/subsonic.script` insecurely:

```
INSERT INTO USER VALUES('$USER','enc:$HEXNUMBER',1234574,0,0,FALSE,'$EMAIL')
```

The value `$HEXNUMBER` can be trivially decoded to reveal a user's password. I, for one, am tired of having to qualify to every new user that they shouldn't make their password something they wouldn't want me to be able to see.

4oo4:
Using md5 in 2018 is inexcusable. jBcrypt looks like the way to go:

http://www.mindrot.org/projects/jBCrypt/
https://www.stubbornjava.com/posts/hashing-passwords-in-java-with-bcrypt

heyarne:
@4oo4 The subsonic API is quite bad and it really doesn't deserve to be called a rest api. It's basically a protocol invented on top of HTTP GET / 200 OK requests.

I think it would really have to be evaluated how people use airsonic currently. Which clients do they use, over which protocols do they interface with it etc. After we know that we could check whether old clients break if we do stuff like return addtional values in JSON responses, return proper status codes etc.

Any idea how to best assess that? I don't know of any good way to do polls in github issues. On the other hand I don't know how representative the share of users on IRC / reddit is.

4oo4:
@heyarne That's a good question, not sure where else you could reach users besides those places and here, and that might risk not being fully representative.

Would it be worth looking into breaking the API where needed for these kinds of things, but then include a 'legacy' option to fallback to the Subsonic API standard? I know that would be a lot of work and add quite a bit of complexity though since you're essentially designing a second API, I would be curious to hear what Airsonic and Subsonic-compatible app devs would think about that.

4oo4:
Since rewriting the API is a huge task (especially since you'd have to convince people to start using it), I realized that for now the best thing to do is use fail2ban since I noticed that authentication failures will give you something like this in `/var/airsonic/airsonic.log`:

```
[...]

PR #1453, opened 2020-01-03
Upgrade to Spring Boot 2.2

fxthomas:
This PR is based on the huge work done in #1372 (and other related PRs) by @randomnicode, but only does the minimal changes required to upgrade to Spring Boot 2.2.

Commit messages contain more info on why some changes were required when upgrading.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/security/GlobalSecurityConfig.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/security/GlobalSecurityConfig.java
@@ -25,7 +25,7 @@ import org.springframework.security.web.util.matcher.AntPathRequestMatcher;
 import java.security.SecureRandom;
 
 @Configuration
-@Order(SecurityProperties.ACCESS_OVERRIDE_ORDER)
+@Order(SecurityProperties.BASIC_AUTH_ORDER - 2)
 @EnableGlobalMethodSecurity(securedEnabled = true, prePostEnabled = true)
 public class GlobalSecurityConfig extends GlobalAuthenticationConfigurerAdapter {
 
--- a/airsonic-main/src/main/java/org/airsonic/player/security/GlobalSecurityConfig.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/security/GlobalSecurityConfig.java
@@ -18,11 +18,16 @@ import org.springframework.security.config.annotation.method.configuration.Enabl
 import org.springframework.security.config.annotation.web.builders.HttpSecurity;
 import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;
 import org.springframework.security.config.http.SessionCreationPolicy;
+import org.springframework.security.crypto.password.DelegatingPasswordEncoder;
+import org.springframework.security.crypto.password.NoOpPasswordEncoder;
+import org.springframework.security.crypto.password.PasswordEncoder;
 import org.springframework.security.web.authentication.UsernamePasswordAuthenticationFilter;
 import org.springframework.security.web.context.request.async.WebAsyncManagerIntegrationFilter;
 import org.springframework.security.web.util.matcher.AntPathRequestMatcher;
 
 import java.security.SecureRandom;
+import java.util.HashMap;
+import java.util.Map;
 
 @Configuration
 @Order(SecurityProperties.BASIC_AUTH_ORDER - 2)
@@ -71,6 +76,32 @@ public class GlobalSecurityConfig extends GlobalAuthenticationConfigurerAdapter 
             settingsService.save();
         }
         auth.authenticationProvider(new JWTAuthenticationProvider(jwtKey));
+    }
+
+    @Bean
+    public PasswordEncoder delegatingPasswordEncoder() {
+
+        // Spring Security 5 require storing the encoder id alongside the encoded password
+        // (e.g. "{md5}hash" for an MD5-encoded password hash), which differs from previous
+        // versions.
+        //
+        // Airsonic unfortunately stores passwords in plain-text, which is why we are setting
+        // the "no-op" (plain-text) password encoder as a default here. This default will be
+        // used when no encoder id is present.
+        //
+        // This means that legacy Airsonic passwords (stored simply as "password" in the db)
+        // will be matched like "{noop}password" and will be recognized successfully. In the
+        // future password encoding updates will be done here.
+
+        PasswordEncoder defaultEncoder = NoOpPasswordEncoder.getInstance();
+        String defaultIdForEncode = "noop";
+
+        Map<String, PasswordEncoder> encoders = new HashMap<>();
+        encoders.put(defaultIdForEncode, defaultEncoder);
+        DelegatingPasswordEncoder passworEncoder = new DelegatingPasswordEncoder(defaultIdForEncode, encoders);
+        passworEncoder.setDefaultPasswordEncoderForMatches(defaultEncoder);
+
+        return passworEncoder;
     }
 
     private static String generateRememberMeKey() {
--- a/airsonic-main/src/main/java/org/airsonic/player/security/GlobalSecurityConfig.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/security/GlobalSecurityConfig.java
@@ -13,7 +13,7 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.core.annotation.Order;
 import org.springframework.security.config.annotation.authentication.builders.AuthenticationManagerBuilder;
-import org.springframework.security.config.annotation.authentication.configurers.GlobalAuthenticationConfigurerAdapter;
+import org.springframework.security.config.annotation.authentication.configuration.GlobalAuthenticationConfigurerAdapter;
 import org.springframework.security.config.annotation.method.configuration.EnableGlobalMethodSecurity;
 import org.springframework.security.config.annotation.web.builders.HttpSecurity;
 import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;
--- a/airsonic-main/src/main/java/org/airsonic/player/spring/ServletConfiguration.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/spring/ServletConfiguration.java
@@ -6,7 +6,7 @@ import org.springframework.context.annotation.Configuration;
 import org.springframework.web.servlet.ViewResolver;
 import org.springframework.web.servlet.config.annotation.ContentNegotiationConfigurer;
 import org.springframework.web.servlet.config.annotation.PathMatchConfigurer;
-import org.springframework.web.servlet.config.annotation.WebMvcConfigurerAdapter;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
 import org.springframework.web.servlet.handler.SimpleUrlHandlerMapping;
 import org.springframework.web.servlet.view.InternalResourceViewResolver;
 import org.springframework.web.servlet.view.JstlView;
@@ -14,7 +14,7 @@ import org.springframework.web.servlet.view.JstlView;
 import java.util.Properties;
 
 @Configuration
-public class ServletConfiguration extends WebMvcConfigurerAdapter {
+public class ServletConfiguration implements WebMvcConfigurer {
 
     @Bean
     public SimpleUrlHandlerMapping podcastUrlMapping(PodcastController podcastController) {
--- a/airsonic-main/src/main/java/org/airsonic/player/spring/ServletConfiguration.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/spring/ServletConfiguration.java
@@ -4,6 +4,9 @@ import org.airsonic.player.controller.PodcastController;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.web.servlet.ViewResolver;
+import org.springframework.web.servlet.config.annotation.ContentNegotiationConfigurer;
+import org.springframework.web.servlet.config.annotation.PathMatchConfigurer;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurerAdapter;
 import org.springframework.web.servlet.handler.SimpleUrlHandlerMapping;
 import org.springframework.web.servlet.view.InternalResourceViewResolver;
 import org.springframework.web.servlet.view.JstlView;
@@ -11,7 +14,7 @@ import org.springframework.web.servlet.view.JstlView;
 import java.util.Properties;
 
 @Configuration
-public class ServletConfiguration {
+public class ServletConfiguration extends WebMvcConfigurerAdapter {
 
     @Bean
     public SimpleUrlHandlerMapping podcastUrlMapping(PodcastController podcastController) {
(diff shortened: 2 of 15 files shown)


airsonic-main/src/main/java/org/airsonic/player/spring/ServletConfiguration.java
```diff
@@ -4,14 +4,17 @@
...
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurerAdapter;
 import org.springframework.web.servlet.handler.SimpleUrlHandlerMapping;
 import org.springframework.web.servlet.view.InternalResourceViewResolver;
 import org.springframework.web.servlet.view.JstlView;
 
 import java.util.Properties;
 
 @Configuration
-public class ServletConfiguration {
+public class ServletConfiguration extends WebMvcConfigurerAdapter {
```
randomnicode: deprecated class. new mvc doesn't need an adapter, you can use the interface directly.

also you don't need to do this here. you can just enable it in your properties file.

https://github.com/airsonic-advanced/airsonic-advanced/commit/b4f286fd372dd7038801ba1029d59724da64cfe3

fxthomas: Good point about the interface.

On the other hand, since we're moving all the configuration to Java, it makes more sense to me to have everything there.

airsonic-main/src/main/java/org/airsonic/player/security/GlobalSecurityConfig.java
```diff
@@ -73,6 +78,32 @@ public void configureGlobal(AuthenticationManagerBuilder auth) throws Exception
...
+        PasswordEncoder defaultEncoder = NoOpPasswordEncoder.getInstance();
+        String defaultIdForEncode = "noop";
+
+        Map<String, PasswordEncoder> encoders = new HashMap<>();
+        encoders.put(defaultIdForEncode, defaultEncoder);
+        DelegatingPasswordEncoder passworEncoder = new DelegatingPasswordEncoder(defaultIdForEncode, encoders);
+        passworEncoder.setDefaultPasswordEncoderForMatches(defaultEncoder);
+
+        return passworEncoder;
+    }
```
randomnicode: This is upto you, but I think the easier and more correct way is to retain the default encoder and modify SecurityService to add "{noop}" to the password string retrieved from the db (see corresponding file in airsonic-advanced), so it's passed onto the correct delegate (which is already present in the default delegatingpasswordencoder).

Currently you're overriding the default password encoder that contains _all_ encoders (including noop) with something that only has a noop encoder.

you're effectively taking away all other password encoders, so the encoder is forced to use the only one there. instead i think you should take your implementation of passwords and route it to the correct place.

Additionally, please open an issue to migrate passwords to nonplaintext. I intend to submit a pr for that soon on airsonic-advanced.

airsonic-main/src/test/java/org/airsonic/player/api/jukebox/AbstractAirsonicRestApiJukeboxIntTest.java
```diff
@@ -46,16 +47,19 @@
...
+            return new BeanPostProcessor() {
+                @Override
+                public Object postProcessAfterInitialization(Object bean, String beanName) {
+                    if (bean instanceof PlayerDaoPlayQueueFactory) {
+                        PlayerDaoPlayQueueFactory temp = (PlayerDaoPlayQueueFactory) spy(bean);
+                        doReturn(spy(temp.createPlayQueue())).when(temp).createPlayQueue();
+                        bean = temp;
+                    }
+                    return bean;
+                }
```
randomnicode: Implementation note. I chose to implement it this way so we didn't have to enable bean overriding, so it's easier to spot when there are multiple instantiations of the same bean.

An alternate implementation is to of course turn on bean overriding for tests.

fxthomas: Turning on bean overriding did not seem to be very future-proof given that they pretty much deprecated it. I prefer this solution!

airsonic-main/src/main/java/org/airsonic/player/spring/ServletConfiguration.java
randomnicode: we're not moving everything to java. we're moving (have moved?) xml to java.

your application properties remain outside java.
your opinion can differ of course, but consider: this can be autoconfigured by spring boot with one line in the config provided specifically for this reason instead of manually implementing a spring class to override it.

Use the facility provided by the framework.

randomnicode:
So far as i can tell off the top of my head, you're missing at least the update to your spotify docker dependency, and i dont think i saw you update the spring boot maven plugin to package everything right (see airsonic-advanced)

airsonic-main/src/main/java/org/airsonic/player/security/GlobalSecurityConfig.java
fxthomas: Your implementation, for me, has the issue that you're mixing up an implementation detail (adding "{noop}" to the password) in the middle of the authentication code, instead of doing it in the Spring configuration.

For the issue see #69 (you could have searched :wink:).

airsonic-main/src/main/java/org/airsonic/player/spring/ServletConfiguration.java
fxthomas: On the other hand Java is type-checked, but the XML/properties files aren't. This is, to me, the main benefit to moving most of the configuration to Java. If the API changes for some reason, then at least the compiler has your back.

airsonic-main/src/main/java/org/airsonic/player/security/GlobalSecurityConfig.java
randomnicode: It is absolutely an implementation detail of how passwords are implemented _specifically in airsonic_. Why would you therefore modify the framework and override the bean the framework provides?

Does the default spring password encoder not have the capability to decode the given password? Yes it does. And more. You just need to route it correctly.

The issue is specifically with airsonic implementation, therefore it seems to me that you need to modify the implementation itself to route the password correctly instead of change the framework to meet the incomplete implementation.

Consider future upgrades. Say while migrating you have a mix of bcrypt and scrypt and noop passwords coexisting in the db, which you identify in a separate column. Your configuration will be of no use. Rerouting the passwords to their correct encoder, however, will.

Thank you, that does seem to be an old issue. The password migration PR shouldn't be hard, but I will read the issue for potential pitfalls everyone seems to be trying to point out.

fxthomas:
Ah, good catch. It's still missing a few things, but it works enough so that I can test-drive it on my instance (no issues for now)! :smile: 

The original PR was a good starting point, but it had a lot of refactoring and structure changes that should IMO have been discussed separately in a different PR, and not formulated as a forced requirement to accepting the Spring upgrade PR.

Especially, figuring out the `@EnableWebMvc` annotation gave me a lot of headaches, dammit why is the Spring/Spring Boot documentation so scattered around?

Anyway, with this we should be close to finally getting all these upgrades done.

airsonic-main/src/main/java/org/airsonic/player/spring/ServletConfiguration.java
randomnicode: ? I do not understand this comment.

Your bindings will fail if you provide an incorrect config property type. Literally all your tests (that rely on this behavior) will fail. Do you think type checking doesn't happen at binding?

Look, you can continue using this method if you want, my point is that if you're going to use a framework, then you might as well use it and rely on it, including the properties it provides for you. You're not programmatically doing anything exotic to justify maintaining custom class overriding code that now _you_ now have the burden of maintaining, instead of the framework.
The framework gives you specific properties for a particular reason, and I can guarantee they'll do a lot more checking than you will. Switching to Java is justified and valid if the value needs to be dynamically calculated, but none of that is true here.

I mean if you don't want to use the facilities provided by the framework, then you can also write all the wire injection stuff too. It's not hard, but the point is that now _you're_ going to maintain it. I hope my point makes it across.

randomnicode:
I believe it was in my original PR as a specific standalone commit: Don't explicitly annotate with EnableWebMvc, it'll turn off a lot of autoconfig.

airsonic-main/src/main/java/org/airsonic/player/security/GlobalSecurityConfig.java
fxthomas: Let me consider that some more, I'll have a closer look and see if there is a better way.

randomnicode:
> Anyway, with this we should be close to finally getting all these upgrades done.

No, this is _one_ minor upgrade. There are plenty more. JSR310, mediascanner, streams, ranges, passwords, hsqldb.

none of these even change logic and just retain feature parity. yet.

But goodluck with this one. :)

airsonic-main/src/main/java/org/airsonic/player/spring/ServletConfiguration.java
fxthomas: Spring seems to consider properties and code equally, to the point that a lot of documentation mention both at the same time, so it's not like they say "you guys should use properties" in our case.

If the property name changes or the property is removed for some reason, its behavior will silently start to fail. Of course the tests may pick it up, but then they will also fail at some other place with an unrelated error message. Also (in Airsonic we have .properties files in both `src/main` and `src/test`) you may need to maintain additional configuration files for integration tests.

randomnicode: Yeah, that's not accurate at all.

If it's documentation you're trying to argue, then here is the actual relevant section of the official Spring Boot documentation for this exact change you're trying to address along with the prescribed solution: https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-spring-mvc-pathmatch

You'll notice that it tells you to use the _properties_, and there is nary a mention of any code.

Leverage the framework you're using; you're not using raw MVC, you're using Boot.

Of course, on this minor matter, you can continue "javafying" if you wish. This point isn't really worth debating over more. I gave my opinion, presented the Spring documentation for it, obviously it's your PR fork to do as you'd like.

randomnicode:
> The original PR was a good starting point, but it had a lot of refactoring and structure changes that should IMO have been discussed separately in a different PR, and not formulated as a forced requirement to accepting the Spring upgrade PR.

I'm sorry, were they not separate PRs?

airsonic-main/src/main/java/org/airsonic/player/security/GlobalSecurityConfig.java
muff1nman: I haven't yet looked at this in detail yet, but isnt this irrelevant given that we have a custom UserDetailsService which doesn't use a PasswordEncoder at all?

randomnicode: No, if you use Spring's authentication (which you do), then you use its PasswordEncoder under the hood (which you do). The failing tests which @fxthomas no doubt observed alerted him to the issue leading to this change.

However, which PasswordEncoder you use is determined by a prefix to the password that, if we used Spring to encode prior to storage, is affixed to the password by Spring itself. You construct the user details with that password in your custom UserDetailsService (called SecurityService).

Because we don't use Spring to encode, and yet use it to authenticate and thus do password matching, Spring is at a loss at which encoder to forward the password to. It used to default to a noop strategy (thus is worked), but now the default is (I believe) bcrypt.

My proposal is, in lieu of migration, to affix the correct prefix that Spring itself would generate to the passwords. A bcrypt hash by Spring is stored as `{bcrypt}hash`, a noop would be stored as `{noop}password`; since we are storing only `password`, prefixing `{noop}` when feeding it to Spring (via UserDetails) lets it forward it to the correct encoder it has set up. The other encoders can also be redirected to with a different prefix. The other alternative implemented here is to change the Encoder bean itself and strip all encoders from it except the noop encoder (set it as default).

fxthomas: I'm still considering what to do here. We also store MD5-encoded password like `enc:<md5 hash>` in that field (so it's not no-op at all in most cases), so neither solution looks good for me.

Essentially if we want to migrate to newer algorithms in the future it would make sense to create a `PasswordEncoder` that:

* If the encoded password starts with `{`, uses the default encoder
* If the encoded password starts with `enc:`, uses the MD5 encoder
* Otherwise uses the no-op decoder

Another question I still need to investigate is : do we really need the no-op decoder? I think we don't (it only seems to be used by the default admin user, changed at first boot), but I haven't completely verified that. If this is the case, it's even better to remove it entirely (which could be done in another PR).

randomnicode: - I think you're mistaken: we do not store md5 hashes. That would render a particular form of auth mentioned in the topic you pointed to earlier (#69) inoperable (salt+token). We store the password itself in hex.
- The point with respect to additional "enc:" is well pointed, but, as currently constructed, it is opaque to the layer in question. "enc:" is just a DB persistence detail, completely abstracted away by the Dao; it isn't even seen by the auth code. The PasswordEncoder being used is absolutely the NoopEncoder, regardless of whatever and however you're storing things in the DB. You can of course make the Dao dumber and introduce a new Encoder that does what the Dao does, but after "decryption", it'll still be a NoopEncoder.
- Regardless of what you do or don't do in the DB, from my perspective, you need to migrate passwords into a consistent representation.

fxthomas: > I think you're mistaken: we do not store md5 hashes

Indeed, you're right. I must have had a brain fart... Thanks for keeping up!

I still think the idea of having the correct Encoder is better in the sense that it is "correct" ; I'm uneasy about having "{noop}enc:something", this will bite us somewhere later. I'd rather get it right first.

randomnicode: Again, I think you're losing something in abstraction, or I'm misunderstanding you.

> I'm uneasy about having "{noop}enc:something"

`{noop}enc:something` will never happen. `enc:` is a storage abstraction that only the dao is aware of. It never makes it out of the dao.

Maybe I can make it clearer:
- `plaintext` comes in
- `plaintext` goes to the dao to be stored
- dao stores it in the db as `enc:hex`, it really doesn't matter how the db stores it
- on retrieval (say for password checking), dao retrieves it and converts back to `plaintext`
- anything outside the dao (like the User object) never sees `enc:hex`, it only sees `plaintext`
- you now need to redirect `plaintext` to the correct password encoder for comparison, therefore you append `{noop}`

I hope that makes sense. How you store it (the domain of the dao) is different and distinct from how the password is compared (the domain of the password encoder). I could store it a million different ways and in a million different places, the comparison only takes whatever the User object has (which is always `plaintext`).

fxthomas: Aaah, I think I understand what you mean now! I assumed (without checking, which I should have) that the User object had a 1:1 mapping with how it was stored in the DB, which is clearly not the case from what you say.

pom.xml
```diff
@@ -18,7 +18,7 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <cxf.version>3.3.5</cxf.version>
         <jackson.version>2.10.3</jackson.version>
-        <tomcat.version>8.5.51</tomcat.version>
+        <tomcat.version>9.0.33</tomcat.version>
```
randomnicode, line 21: no one is going to remember to change the tomcat versions when they need to update spring boot versions.

just get rid of this property and use the in built one that the spring bom is providing for you for all tomcat dependencies

fxthomas:
@airsonic/maintainers I want to merge this next week, since this is a breaking change on `master` I thought you might like the heads up.

Among others it will also need a note on the release notes/documentation about the changed properties (I added a todo item in the release tracking issue for 11.0).

PR #1454, opened 2020-01-03
Turn off Airsonic banner during tests

fxthomas:
The Airsonic ASCII art banner pollutes the output multiple times when running the unit/integration tests, how about removing it? Anybody against it?


randomnicode:
yeah, i think it's a good idea, it just polluted the test logs.

i turned it off too sometime back https://github.com/airsonic-advanced/airsonic-advanced/pull/46

PR #1455, opened 2020-01-05
Fix the boldness of the main submenu

### LINKED ISSUE #1424: That top navigation restyling... :-\
## Problem description

This is kinda a nitpicky issue, I apologize.

Theme: 2 Minutes to Midnight

**10.2.1:**
Top navigation on version 10.2.1

**Later versions:**
Top navigation on current Airsonic

I'm not sure who thought this change was better, but aesthetically speaking, it's certainly not. Having all this text boldfaced only makes it more noticeable that the font size of all three lines is fairly similar and gives it a rather "clunky" appearance. Also, notice how much longer the category row is now. This text does not wrap around in any way, so now categories that were usable before ("By decade" / "By genre" / "All") are more likely to get pushed off the edge of the window, out of reach, depending on the browser window size.

### Steps to reproduce

This is a general appearance comment, not really a playback functionality issue.

## System information

 * **Airsonic version**: 10.5.0-RELEASE
 * **Operating system**: FreeBSD 11.2-RELEASE-p9
 * **Java version**: Apache Tomcat/8.5.43, java 1.8.0_192,
 * **Proxy server**: nginx 1.16.1
 * **Client**: *n/a -- seen in Waterfox, Chrome, and Safari*
 * **Language**: en-US

## Additional notes

Is the styling here part of the theme? If I can just edit a theme file myself and make things as I want here I'd be fine with that. Considering this change is appearing on all themes I've tried I assume it's not.

jvoisin:
This should close #1424

before:
before

after:
after

### DIFF
--- a/airsonic-main/src/main/webapp/WEB-INF/jsp/home.jsp
+++ b/airsonic-main/src/main/webapp/WEB-INF/jsp/home.jsp
@@ -38,7 +38,7 @@
     <h2>${model.welcomeSubtitle}</h2>
 </c:if>
 
-<h2>
+<p>
     <c:forTokens items="random newest starred highest frequent recent decade genre alphabetical" delims=" " var="cat" varStatus="loopStatus">
         <c:if test="${loopStatus.count > 1}">&nbsp;|&nbsp;</c:if>
         <sub:url var="url" value="home.view">
@@ -55,7 +55,7 @@
         </c:choose>
 
     </c:forTokens>
-</h2>
+</p>
 
 <%@ include file="homePager.jsp" %>
 


fxthomas:
Looks good to me.

Minor nit : isn't it better to use `<h2>` (since they are titles, semantically speaking) and style that in CSS?

jvoisin:
I was lazy and went for the easiest thing, since I didn't want to test it for every single theme :/
But you're right, smenatically, `h2` would be better.

fxthomas:
It's pretty much okay on most themes I could see, so merging it.

PR #1466, opened 2020-01-11
Fix two possible nullptr deref

jvoisin:


airsonic-main/src/main/java/org/airsonic/player/controller/CoverArtController.java
```diff
@@ -95,7 +95,9 @@ public void init() {
 
     public long getLastModified(HttpServletRequest request) {
         CoverArtRequest coverArtRequest = createCoverArtRequest(request);
-        //        LOG.info("getLastModified - " + coverArtRequest + ": " + new Date(result));
+        if (null == coverArtRequest) {
+            return -1L;
```
muff1nman: Why not throw an exception instead of returning a garbage value?

jvoisin: Because there are only two usages of `getLastModified`, both in `airsonic-main/src/main/java/org/airsonic/player/controller/SubsonicRESTController.java`: the first one is treating `-1` as a regular value, and the second one will spit an error is `-1` is returned. I think that this is cleaner than an exception.

randomnicode: -1 isn't a garbage value. It's part of the method contract and signifies that this value must be regenerated (it's in the javadoc of the method). I think this is the correct approach.

PR #1467, opened 2020-01-11
Clean up a bit the jaudiotagger wrapper

jvoisin:
- ID3v24 are now fully supported by jaudiotagger,
  there is no need to have special code to handle
  them
- Inline a couple of one-line'n'useless methods
- Replace isImageAvailable with getArtwork
- Make some methods static
- Specialize some exceptions

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/MediaFileService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/MediaFileService.java
@@ -624,11 +624,7 @@ public class MediaFileService {
         // Look for embedded images in audiofiles. (Only check first audio file encountered).
         for (File candidate : candidates) {
             if (parser.isApplicable(candidate)) {
-                if (parser.isImageAvailable(getMediaFile(candidate))) {
-                    return candidate;
-                } else {
-                    return null;
-                }
+                return JaudiotaggerParser.getArtwork(getMediaFile(candidate)) != null ? candidate : null;
             }
         }
         return null;
--- a/airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
@@ -27,10 +27,8 @@ import org.jaudiotagger.audio.AudioFile;
 import org.jaudiotagger.audio.AudioFileIO;
 import org.jaudiotagger.audio.AudioHeader;
 import org.jaudiotagger.tag.FieldKey;
+import org.jaudiotagger.tag.KeyNotFoundException;
 import org.jaudiotagger.tag.Tag;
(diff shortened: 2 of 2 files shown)


tesshucom:
> there is no need to have special code to handle them

No problem in this regard!
Before, I did a read survey of ID3v2.2 to ID3v2.4 separately from this PR.
 - Music Center for PC(2.1.0)
 - iTunes(4.1.0.52-12.9.6.3)

No problem with single / multibyte strings.

fxthomas:
> ID3v24 are now fully supported by jaudiotagger,
There is no version upgrade here, I assume that jaudiotagger was upgraded at some point in the past without actually changing the code?

jvoisin:
I would assume so too, yes.

tesshucom:
The library remains the same, 
but it is likely that the current code contains certain misconceptions.

The current redundant coding is from #1000.
This is to avoid the phenomenon that occurs when a certain 2.4 data is called.

___

Following this PR, I conducted additional personal investigations.
The content is a test to confirm that the data generated by the latest Music Center for PC (SONY), a wide range of iTunes (APPLE), and Mp3Tag can be read correctly.
That is, data with relatively high reliability was re-created and confirmed.

 - [v2.4](https://github.com/tesshucom/jpsonic/tree/master/jpsonic-main/src/test/resources/MEDIAS/Metadata/v2.4)
 - [v2.3](https://github.com/tesshucom/jpsonic/tree/master/jpsonic-main/src/test/resources/MEDIAS/Metadata/v2.3)
 - [v2.3+v1.1](https://github.com/tesshucom/jpsonic/tree/master/jpsonic-main/src/test/resources/MEDIAS/Metadata/v2.3%2Bv1.1)
 - [v2.3+v1.0](https://github.com/tesshucom/jpsonic/tree/master/jpsonic-main/src/test/resources/MEDIAS/Metadata/v2.3%2Bv1.0)
 - [v2.2](https://github.com/tesshucom/jpsonic/tree/master/jpsonic-main/src/test/resources/MEDIAS/Metadata/v2.2)

[JaudiotaggerParserTestCase](https://github.com/tesshucom/jpsonic/blob/master/jpsonic-main/src/test/java/org/airsonic/player/service/metadata/JaudiotaggerParserTestCase.java)

As a result, there is no problem removing #1000.

___

Strictly speaking, the #1000 phenomenon can recur if code is shortened.
We expect this to happen if it is generated by certain bad client software.
#1000 is considered a dirty method against certain data, rather than essentially a fix related to v2.4.

 - There is no problem as it is.
 - If the operating criteria is based on relatively reliable software output, it is redundant code.

You can copy the test data and case, but I was hesitant to PR because I created and confirmed the test data in Japanese and English.

muff1nman:
@tesshucom can I delegate you to review this one?

tesshucom:
OK.
Give me a few days to get to it.

PR #1468, opened 2020-01-12
Set limit in random track query itself instead of in connection

jvoisin:
This allows query plan optimization.

This is a direct backport of
https://github.com/airsonic-advanced/airsonic-advanced/commit/f84644ae3df80fe4918844b62035b6bffeeed824

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/dao/AbstractDao.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/dao/AbstractDao.java
@@ -102,16 +102,6 @@ public class AbstractDao {
         return result;
     }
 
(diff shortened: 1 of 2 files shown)


fxthomas:
I'm curious, do you have some sources about DB optimizing queries differently if passed a limit clause in the SQL instead of the JDBC params? It makes sense, kinda, but I never knew that was a thing.

eharris:
It really depends on the database in use and its optimizer, which then selects the query plan.  But in general, the more info you provide the optimizer, the better it does.  Putting the limit in the query itself ensures that the optimizer has that info.  Otherwise, some databases could potentially do things as stupidly as querying without any limit, retrieving all the rows, and then the driver discarding all but N of them when returning the results.

PR #1469, opened 2020-01-13
WIP: Add option to disable "not very secure" t+s REST API auth

fxthomas:
This is a very early attempt following discussion in #69, showing what I had in mind in https://github.com/airsonic/airsonic/issues/69#issuecomment-573786360.

Note that I had to explicitly enable HTTP Basic Auth so that DSub works for me (it transmits the `Authorization` header in addition to either `t+s` or `p` parameters but that wasn't taken into account by Airsonic), not sure if that's the right way or if I did something wrong in my configuration.

Also, in its current form this spams the logs with warnings about the auth method if used with an app that tries to supply `t+s` (as DSub does).

At least the branch is here so that folks can test it.


heyarne:
I just wanted to note that [ampache also removed the md5 token](https://github.com/ampache/ampache/wiki/API#subsonic-api). I don't know their motivation but it's a good sign that it's a reasonable decision.

fxthomas:
Well, I'll close this for now to avoid polluting the PR list with a test branch.

PR #1471, opened 2020-01-14
Remove sync locks everywhere in StatusService and TransferStatus

jvoisin:
This is a direct backport of
https://github.com/airsonic-advanced/airsonic-advanced/pull/52/commits/caa5f8346cc4597517e25847780a98af0f3432ec
without all the functional-style-changes.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/StatusService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/StatusService.java
@@ -28,6 +28,8 @@ import org.springframework.stereotype.Service;
 
 import java.io.File;
 import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
+
 
 /**
  * Provides services for maintaining the list of stream, download and upload statuses.
@@ -43,35 +45,27 @@ public class StatusService {
     @Autowired
     private MediaFileService mediaFileService;
 
-    private final List<TransferStatus> streamStatuses = new ArrayList<TransferStatus>();
-    private final List<TransferStatus> downloadStatuses = new ArrayList<TransferStatus>();
-    private final List<TransferStatus> uploadStatuses = new ArrayList<TransferStatus>();
-    private final List<PlayStatus> remotePlays = new ArrayList<PlayStatus>();
+    private final List<TransferStatus> streamStatuses = Collections.synchronizedList(new ArrayList<>());
+    private final List<TransferStatus> downloadStatuses = Collections.synchronizedList(new ArrayList<>());
+    private final List<TransferStatus> uploadStatuses = Collections.synchronizedList(new ArrayList<>());
+    private final List<PlayStatus> remotePlays = Collections.synchronizedList(new ArrayList<>());
 
     // Maps from player ID to latest inactive stream status.
-    private final Map<Integer, TransferStatus> inactiveStreamStatuses = new LinkedHashMap<Integer, TransferStatus>();
+    private final Map<Integer, TransferStatus> inactiveStreamStatuses = new ConcurrentHashMap<>();
 
-    public synchronized TransferStatus createStreamStatus(Player player) {
-        // Reuse existing status, if possible.
-        TransferStatus status = inactiveStreamStatuses.get(player.getId());
-        if (status != null) {
-            status.setActive(true);
-        } else {
-            status = createStatus(player, streamStatuses);
-        }
-        return status;
+    public TransferStatus createStreamStatus(Player player) {
+        return createStatus(player, streamStatuses);
     }
 
     public synchronized void removeStreamStatus(TransferStatus status) {
-        // Move it to the map of inactive statuses.
+        // Move it to the map of inactive statuses
         status.setActive(false);
         inactiveStreamStatuses.put(status.getPlayer().getId(), status);
         streamStatuses.remove(status);
     }
 
-    public synchronized List<TransferStatus> getAllStreamStatuses() {
-
-        List<TransferStatus> result = new ArrayList<TransferStatus>(streamStatuses);
+    public List<TransferStatus> getAllStreamStatuses() {
+        List<TransferStatus> result = new ArrayList<>(streamStatuses);
 
         // Add inactive status for those players that have no active status.
         Set<Integer> activePlayers = new HashSet<Integer>();
@@ -87,7 +81,7 @@ public class StatusService {
         return result;
     }
 
-    public synchronized List<TransferStatus> getStreamStatusesForPlayer(Player player) {
+    public List<TransferStatus> getStreamStatusesForPlayer(Player player) {
         List<TransferStatus> result = new ArrayList<TransferStatus>();
         for (TransferStatus status : streamStatuses) {
             if (status.getPlayer().getId().equals(player.getId())) {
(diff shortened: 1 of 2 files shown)


airsonic-main/src/main/java/org/airsonic/player/service/StatusService.java
```diff
@@ -43,35 +45,29 @@
...
+    public TransferStatus createStreamStatus(Player player) {
+        return createStatus(player, streamStatuses);
     }
 
     public synchronized void removeStreamStatus(TransferStatus status) {
         // Move it to the map of inactive statuses.
-        status.setActive(false);
-        inactiveStreamStatuses.put(status.getPlayer().getId(), status);
-        streamStatuses.remove(status);
+        inactiveStreamStatuses.compute(status.getPlayer().getId(), (k, v) -> {
```
fxthomas: Is it normal that neither `k` nor `v` is used? I'm not sure I understand the use of `compute` in that case.

jvoisin: Because it's what `compute` is expecting: https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#compute-K-java.util.function.BiFunction-

I copy-pasted this from @randomnicode 's code, but I can remove it, since it doesn't seem super-useful in this case.

randomnicode: Both `k` (the key to bind the new value to) and `v` (the old value, if any) are not really needed. The key-binding is implicit, and we discard the old value (if present). We just need to associate the key with the new value.

There is a reason, however, why `compute` is used and that's because of concurrency. `inactiveStreamStatuses` is a `ConcurrentHashMap`. Using `compute` guarantees atomicity with respect to incoming thread requests. In fact, this pull request seems incomplete to me because that is the very reason you can remove the synchronized lock on this entire method, entrusting the ConcurrentHashmap to do its job.

Here's what might happen without the compute AND the sync:
- thread1: removeStreamStatus(statusA)
- thread2: removeStreamStatus(statusB)
- thread1: statusA.setActive()
- thread2: statusB.setActive()
- thread2: inactiveStreamStatuses.put(id, statusB) <- preempted
- thread1:  inactiveStreamStatuses.put(id, statusA)
...

statusA is followed by statusB, you would expect statusB to be the final inactiveStream for that id, but statusA comes out bound at the end.

With syncs, you can achieve good order by locking every request out and making them wait in a serial order.

With concurrent maps, you can do even better by making only some requests wait (ids in the same hash bucket), and letting everything else transact concurrently.

Again, this PR seems incomplete to me. I likely made more modifications later? The current version of this class in master (or even in the pointed commit: https://github.com/airsonic-advanced/airsonic-advanced/commit/caa5f8346cc4597517e25847780a98af0f3432ec#diff-9cf7e6b3f2a25472d4d3e4220a8dcb71R63 ) does not have a sync in the method signature

fxthomas: > With concurrent maps, you can do even better by making only some requests wait (ids in the same hash bucket), and letting everything else transact concurrently.

Thanks, that is a good reason.

In any case, swapping an explicit `synchronized` with an implicit "you should know that compute is synchronized on the internal buckets so it's good for performance!" is not ideal.

I'm okay with either the explicit `synchronized` _or_ having a comment above it making sure that someone who reads this understands this.

randomnicode: > In any case, swapping an explicit synchronized with an implicit "you should know that compute is synchronized on the internal buckets so it's good for performance!" is not ideal.

It's not implicit, it's part of the `ConcurrentHashMap` contract itself and is written in the javadoc pretty clearly. There are no hidden gotchas here:
https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#compute-K-java.util.function.BiFunction-

Here's the explicit portion:
>Attempts to compute a mapping for the specified key and its current mapped value (or null if there is no current mapping). The entire method invocation is performed atomically. Some attempted update operations on this map by other threads may be blocked while computation is in progress, so the computation should be short and simple, and must not attempt to update any other mappings of this Map.

airsonic-main/src/main/java/org/airsonic/player/service/StatusService.java
```diff
@@ -43,35 +45,27 @@
...
-
-    public synchronized TransferStatus createStreamStatus(Player player) {
-        // Reuse existing status, if possible.
-        TransferStatus status = inactiveStreamStatuses.get(player.getId());
-        if (status != null) {
-            status.setActive(true);
-        } else {
-            status = createStatus(player, streamStatuses);
-        }
-        return status;
```
muff1nman: Sorry not quite following, why don't we need this check for existing `TransferStatus` for the given player?

randomnicode: This was a while back so I'm trying to refresh my memory, but I think the more pertinent question is why do you need this check in the first place?

Are you trying to maintain one stream status per player at all times?

That's already not going to happen: if your player hasn't been inactive, every stream transfer status creation will create a new transfer status and add it to the active list (because it won't find anything in the inactive map, and no one is checking the active list), leading to multiple statuses per player in the active list. That's just inconsistent behavior, gated on when a stream status for a player turns inactive for the first time.

Additionally think of the concurrency scenarios. A transfer status being used for a transfer of a file it marks, suddenly gets interrupted by another thread for another file. Now two threads are vying for the transfer status, and while the byte counts may be fine because of the atomic byte counts in there, the file path is certainly not accurate (the status's path will only reflect whatever the latest thread's file path is, even though the earlier thread is still using it for its own file).

I think this way was just cleaner. You can keep a map around for display of older inactive transfers, but always create a new one.

eharris:
I have't looked at this part of the code myself, but just from the description it made me wonder if the map might just be intended to provide info for the "What others are playing" notifications feature (basically the right sidebar).  If that's the case, then adding additional entries rather than replacing/reusing the old one would probably be undesirable.

randomnicode:
>  If that's the case, then adding additional entries rather than replacing/reusing the old one would probably be undesirable.

Perhaps my comment didn't make it explicit enough. You're already adding additional entries for a given player. A player absolutely can have multiple active stream statuses in the active list. Here's your scenario:

- Player 1 gets created and requests a song
- System checks for an inactive transfer status for the player (there is none) and TransferStatus 1 gets created and added to the active list
- Somewhere else while the song is still being played, Player 1 requests another song
- System checks for an inactive transfer status for the player (there is none) and TransferStatus 2 gets created and added to the active list
- Lather, rinse, repeat _n_ times, and each time a new TransferStatus gets created, because each time there is NOTHING in the inactive transfer list.
- After 10 songs, there are 10 transfer statuses for Player 1 in the active list. Nothing is getting reused.
- The "magic cutoff" is when the first stream goes inactive (whichever one does it first). Thereafter we reuse the last one that retired.

This sounds absolutely inconsistent to me and if there is a reason for it, I don't understand it.
Either try and maintain 1 status per player at all times (without any magic cut offs), or allow multiple active ones from the get go.

> if the map might just be intended to provide info for the "What others are playing" notifications feature

Yep, thus my comment:
>I think this way was just cleaner. You can keep a map around for display of older inactive transfers, but always create a new one.

eharris:
So why was this closed?

PR #1496, opened 2020-02-08
Remove jsonpath

jvoisin:


fxthomas:
Works fine for me, a good dependency is a dependency we can remove! :wink:

Issue #1498, opened 2020-02-09
Unable to upgrade MySQL Server to 8.0.18


tstreit:
[catalina.txt](https://github.com/airsonic/airsonic/files/4177693/catalina.txt)

## Problem description

*When attempting to connect a MySQL Server 8.0.18, war file deployment fails, catalina.txt file contains the output of tomcat console.*

### Steps to reproduce

*Upgrade mysql database to at least 8.0.18 or later.*

1. *Upgrade MySQL Server, in my case I compiled it from source.*
2. *Import  mysql-connector-java-8.0.18.jar in tomcat/lib directory*
3. *Restart tomcat.*

## System information

 * **Airsonic version**: *10.5.0-RELEASE – November 7, 2019*
 * **Operating system**: *Ubuntu 18.04.4 LTS (GNU/Linux 4.15.0-76-generic x86_64)*
 * **Java version**: *OpenJDK 64-Bit Server VM (build 25.242-b08, mixed mode)*
 * **Proxy server**: *nginx*
 * **Client**: *Chrome 79.0.3945.130, DSub on Android*
 * **Language**: *English*

## Additional notes

*Would like to be able to upgrade database server without restriction.*

randomnicode:
`tx_isolation` was renamed in MySQL 8.x, but the current airsonic version still comes packaged with an outdated driver (5.x).

Airsonic provides an ancient version of the mysql connector embedded in its war file here: https://github.com/airsonic/airsonic/blob/master/airsonic-main/pom.xml#L558-L563

The `runtime` scope means it's packaged with the jar.  Adding an additional jar to the container likely won't help, it seems in your case it was just ignored (it tried connecting with the 5.x driver).

Your options are:
- Plead with airsonic devs to update their dependencies. MySQL 8.x has been around for 2 years now and I'm surprised no one's complained yet. I guess either no one uses it on 8.x, or they tried, failed and moved to different options.
- Update the dependency yourself on your own fork
- Use a fork of airsonic with newer and more current dependencies, such as https://github.com/airsonic-advanced/airsonic-advanced

tstreit:
Thanks for the reply.

For now I just pointed my airsonic instance to another older database that is still supported, so it's working but not ideal. I did try to update the dependency myself but unfortunately I am not as familiar with the code base to do it properly.

[...]

PR #1511, opened 2020-02-11
Update mysql-connector-java to 8.0.19

fxthomas:
Airsonic does not work with MySQL 8, which I discovered recently lately during testing and development. I did not see any issue with upgrading to the latest version of the JDBC driver.

Tested with Docker:
```
docker run --name mysql -e MYSQL_DATABASE=airsonic -e MYSQL_ROOT_PASSWORD=123456 -p 3306:3306 -d mysql:8.0 --character-set-server=utf8 --collation-server=utf8_general_ci
```


randomnicode:
Same issue as #1498

PR #1537, opened 2020-02-16
Bump the mysql driver

jvoisin:


fxthomas:
For mysql only there is #1511, so I'm okay with it, let's merge one and close the other?

PR #1538, opened 2020-02-16
Added support for (multistage) docker based builds

pulsar256:
This PR allows for automated Docker Hub image builds and local builds which require just a docker dependency instead of JDK+Maven. 

# Usage

## Previous approach to build a docker image

```
mvn -P docker package
````

## Recommended approach after this PR

```
docker build -t airsonic .
````

### Automated image builds by Docker Hub

For integration with Docker Hub set up Automated Builds linked to the official Airsonic Github Repository and configure *one* build rule matching the release tags to mimic the current release/tagging vs image tagging scheme:

```
Source Type: Tag
Source: /^v([0-9.]+)$/
Docker-Tag: {\1}-RELEASE,latest
```

Docker Hub Configuration

# Open for discussion:

Should the `dockerfile-maven-plugin` / `docker` maven module be removed in favor of this approach in a subsequent PR? The plugin`s README.md states:  "At this point, we're not developing or accepting new features or even fixing non-critical bugs."

Signed-off-by: Paul Rogalinski-Pinter <paul@paul.vc>

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 4
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,44 @@
+################################################################################################################ builder
+
+FROM maven:3-jdk-8 as builder
+WORKDIR /build
+
+# this block will make docker cache the dependencies for subsequent builds of the same image.
+COPY pom.xml .
+COPY subsonic-rest-api/pom.xml ./subsonic-rest-api/pom.xml
+COPY integration-test/pom.xml ./integration-test/pom.xml
+COPY airsonic-sonos-api/pom.xml ./airsonic-sonos-api/pom.xml
+COPY airsonic-main/pom.xml ./airsonic-main/pom.xml
+RUN mvn de.qaware.maven:go-offline-maven-plugin:resolve-dependencies
+
+COPY . .
+RUN mvn package
+
+################################################################################################################ runtime
+
+FROM alpine:3.9 as runtime
+LABEL description="Airsonic is a free, web-based media streamer, providing ubiquitious access to your music." \
+      url="https://github.com/airsonic/airsonic"
+ENV AIRSONIC_PORT=4040 AIRSONIC_DIR=/airsonic CONTEXT_PATH=/ UPNP_PORT=4041 JVM_HEAP=256m
+
+WORKDIR $AIRSONIC_DIR
+RUN apk --no-cache add \
+    ffmpeg \
+    lame \
+    bash \
+    libressl \
+    fontconfig \
+    ttf-dejavu \
+    ca-certificates \
+    tini \
+    curl \
+    openjdk8-jre
+COPY install/docker/run.sh /usr/local/bin/run.sh
+RUN chmod +x /usr/local/bin/run.sh
+COPY --from=builder  /build/airsonic-main/target/airsonic.war airsonic.war
+EXPOSE $AIRSONIC_PORT
+EXPOSE $UPNP_PORT
+EXPOSE 1900/udp
+VOLUME $AIRSONIC_DIR/data $AIRSONIC_DIR/music $AIRSONIC_DIR/playlists $AIRSONIC_DIR/podcasts
+HEALTHCHECK --interval=15s --timeout=3s CMD wget -q http://localhost:"$AIRSONIC_PORT""$CONTEXT_PATH"rest/ping -O /dev/null || exit 1
+ENTRYPOINT ["tini", "--", "run.sh"]
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+.gitignore
+target/
+.idea
+Dockerfile
+docker-compose.yaml

### CODE REVIEW

**Comment by fxthomas** [comment]
Nice job! :+1:

File: Dockerfile [hunk 1]
```diff
@@ -0,0 +1,44 @@
...
+COPY airsonic-sonos-api/pom.xml ./airsonic-sonos-api/pom.xml
+COPY airsonic-main/pom.xml ./airsonic-main/pom.xml
+RUN mvn de.qaware.maven:go-offline-maven-plugin:resolve-dependencies
+
+COPY . .
+RUN mvn package
+
+################################################################################################################ runtime
+
+FROM alpine:3.9 as runtime
```
randomnicode: I'd just take the openjdk's jre-alpine base instead of taking an alpine base and installing openjdk's jre into it

Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+RUN apk --no-cache add \
+    ffmpeg \
+    lame \
+    bash \
+    libressl \
+    fontconfig \
+    ttf-dejavu \
+    ca-certificates \
+    tini \
+    curl \
```
randomnicode: is there need for curl?

don't remember what we use it for... you seem to be using wget for the healthchecks down below, maybe consider just taking it out?

.editorconfig
```diff
@@ -0,0 +1,7 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 4
```
randomnicode: separate pr?

randomnicode:
> Automated image builds by Docker Hub
For integration with Docker Hub set up Automated Builds linked to the official Airsonic Github Repository and configure one build rule matching the release tags to mimic the current release/tagging vs image tagging scheme:

You can also use Travis and add it to .travis.yml. This idea has been proposed before (and a PR was submitted), but the maintainers seem to want to prefer to do it manually.

> Should the dockerfile-maven-plugin / docker maven module be removed in favor of this approach

That would be my preference

.editorconfig
pulsar256: will do.

pulsar256:
> This idea has been proposed before (and a PR was submitted), but the maintainers seem to want to prefer to do it manually.

Is this still the case? The current docker docker image lags more than a week behind the official release.

fxthomas:
Right now we have a few integration tests that use Docker. They're not enabled by default in the `package` build, but Travis runs them.

pulsar256:
> Right now we have a few integration tests that use Docker. They're not enabled by default in the `package` build, but Travis runs them.
> 
> I don't think this Dockerfile runs them, which is something we want to address if we start using it for our default build on Dockerhub.

I had a brief look. The `integration-tests` profile does also activate the docker maven module, references the built image in the `cucumber_hooks` and runs 4 tests against the instantiated container. 

There are multiple approaches to resolve this dependency which will require a more detailed discussion not directly related to this PR as the old functionality is not yet removed. 

Thanks for pointing this out.

randomnicode:
> >I don't think this Dockerfile runs them, which is something we want to address if we start using it for our default build on Dockerhub.

>I had a brief look. The integration-tests profile does also activate the install/docker maven module, references the built image in the cucumber_hooks and runs 4 tests against the instantiated container.

The docker-client is a different dependency (for connecting to the docker image) than the dockerfile-maven-plugin (that builds it). The integration-tests module just need to build the image prior to the test execution, which you can easily substitute in lieu of the dockerfile-maven-plugin (exec-maven it, or use some other way). Additionally, I'm not sure what cucumber uses for its default lifecycle phase, but if it's as an integration test, then you should change your builder to use `mvn package verify` instead of `mvn package`.

Either way removal of the plugin is a separate PR. I'm also in favor of throwing away cucumber.

> Is this still the case? The current docker docker image lags more than a week behind the official release.

Indeed. The https://github.com/airsonic-advanced/airsonic-advanced fork automatically pushes docker images (and github releases) at merge if you want to see an example of that automation via travis

Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+# this block will make docker cache the dependencies for subsequent builds of the same image.
+COPY pom.xml .
+COPY subsonic-rest-api/pom.xml ./subsonic-rest-api/pom.xml
+COPY integration-test/pom.xml ./integration-test/pom.xml
+COPY airsonic-sonos-api/pom.xml ./airsonic-sonos-api/pom.xml
+COPY airsonic-main/pom.xml ./airsonic-main/pom.xml
+RUN mvn de.qaware.maven:go-offline-maven-plugin:resolve-dependencies
+
+COPY . .
+RUN mvn package
```
randomnicode: maybe change this to verify to ensure integration-tests (if any) get run?

randomnicode:
Any issues with the base image after removing curl?

Dockerfile
pulsar256: I would even argue to use `-DskipTests=true` here and leave the process of running integration tests and tagging a release to the CI pipeline. Running the integration tests here would result in a DinD scenario.

pulsar256:
> Any issues with the base image after removing curl?

none so far (running the latest build on my "production" server). I do not expect any libcurl dependencies within the code and health checks were already performed with wget as you already mentioned.

Dockerfile
randomnicode: oo, good point.

so then help me understand why a multistage build is needed. Is it just for stand-alone docker builds especially where you don't have/want mvn on the host?

Let's examine the build scenario:
- CI runs `mvn verify`
  - after `package`, the `pre-integration-test` phase needs and builds the docker container,   which, in turn, runs `mvn package` internally and compiles and packages again (even though an artifact already exists outside and you can just copy it)
  - the verify is done against the newly compiled/packaged artifact sitting inside the container

randomnicode:
> none so far (running the latest build on my "production" server). I do not expect any libcurl dependencies within the code and health checks were already performed with wget as you already mentioned.

great! in that case, i'm wondering if we can be even more aggressive and take out bash too (maybe another PR). are there scripts relying on bash?

Dockerfile
pulsar256: > "Is it just for stand-alone docker builds especially where you don't have/want mvn on the host?" 

In Short: yes. In that particular case for easy integration with Docker Hub. 

From personal experience / opinion:
Expecting / imposing dependencies to specific toolchains on the CI/CD build agent is prone to errors and the root of many failed builds. I like approaches based purely on docker such as Drone.io. So basically boiling down to just one and leaving the responsibility to use the right JDK or maven version required for building to the developers and not bother CI/CD operators with tasks to maintain possibly exotic toolchains. 

That said: my perspective is perhaps biased and affected by my daytime job. Take that opinion  with a grain of salt.

pulsar256:
> are there scripts relying on bash?

init.sh does shebang bash. Did not look for bashisms though.

pulsar256:
> Did not look for bashisms though.

it does use bash specific syntax to check the arguments passed to the script and reassembles the JAVA_OPTS environment using bashisms. This looks like sanity checking to me as the current usage is very well defined and does not require that funcionality. So in theory this could be removed so we can use plain old `sh` instead. 

This would be for another PR with arguably questionable benefits. Personally i like jumping into a running container for debugging purposes and always appreciate the existence of a more advanced shell. &lt;/OT&gt;

Dockerfile
jef: > Expecting / imposing dependencies to specific toolchains on the CI/CD build agent is prone to errors and the root of many failed builds.

I'd argue this is the sole purpose of using Docker and any CI/CD. This ensures that wherever Docker is running, it will work because it has the required toolchain. Like you said,

> not bother CI/CD operators with tasks to maintain possibly exotic toolchains

I'd agree here, but it's important that we make sure that we have all the required dependencies within the package and Docker image.

Dockerfile
jef: I'd also suggest using [adoptopenjdk/openjdk8-openj9](https://hub.docker.com/r/adoptopenjdk/openjdk8-openj9/).

- https://github.com/docker-library/official-images/pull/5710#issuecomment-483219593
- https://www.eclipse.org/openj9/oj9_performance.html

Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+    fontconfig \
+    ttf-dejavu \
+    ca-certificates \
+    tini
+COPY install/docker/run.sh /usr/local/bin/run.sh
+RUN chmod +x /usr/local/bin/run.sh
+COPY --from=builder  /build/airsonic-main/target/airsonic.war airsonic.war
+EXPOSE $AIRSONIC_PORT
+EXPOSE $UPNP_PORT
+EXPOSE 1900/udp
```
jef: > The EXPOSE instruction does not actually publish the port. It functions as a type of documentation between the person who builds the image and the person who runs the container, about which ports are intended to be published.

https://docs.docker.com/engine/reference/builder/#expose

I'd also suggest that we get rid of what's on L22 for ports and just expose what we know is certain here.

```suggestion
EXPOSE 4040
EXPOSE 4041
EXPOSE 1900/udp
```

Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+COPY . .
+RUN mvn package
+
+################################################################################################################ runtime
+
+FROM openjdk:8-jre-alpine as runtime
+LABEL description="Airsonic is a free, web-based media streamer, providing ubiquitious access to your music." \
+      url="https://github.com/airsonic/airsonic"
+ENV AIRSONIC_PORT=4040 AIRSONIC_DIR=/airsonic CONTEXT_PATH=/ UPNP_PORT=4041 JVM_HEAP=256m
```
jef:
```suggestion
ENV AIRSONIC_DIR=/airsonic
ENV CONTEXT_PATH=/
ENV JVM_HEAP=256m
```

Dockerfile
jef: Users will be able to expose what they want on their host, but keeping this as default inside the container won't interfere with other containers or the user's port mappings.

Example:

```shell script
docker run -it --rm \
    -p $MY_AIRSONIC_PORT:4040 \
    -e $MY_AIRSONIC_CONTEXT_PATH:/
    airsonic
```

Dockerfile
pulsar256: agree with the removal of unnecessary indirections (ports). Sure about the multiple lines? Each of them will commit a new layer to the docker image.

randomnicode: They won't.
ENVs don't commit new layers to docker image.

(Additionally I don't think the ENVs should be removed, see my comment below)

Dockerfile
randomnicode: I disagree about removing the ENVs. They're not there solely for the EXPOSE portion of the Dockerfile (which as noted is purely for optics). They're actually there to provide the artifact running inside the container some default and needed env variables.

In the example command provided above, sure the container published 4040  to host's $MY_AIRSONIC_PORT, but the problem is that the artifact inside the container doesn't know that it needs to bind to the container's 4040 (same with the UPNP port etc). It could be binding to the container's 8080, in which case the wrong port is being published by the container to host.

Dockerfile
randomnicode: > In Short: yes. In that particular case for easy integration with Docker Hub.

I see. Without running any tests?

Not arguing the utility of a standalone Dockerfile (which certainly has value), but I think it'd be nicer to have a CI/CD push non-failed builds only to Docker Hub.

But of course, that's just my subjective opinion.

randomnicode:
> This would be for another PR with arguably questionable benefits. Personally i like jumping into a running container for debugging purposes and always appreciate the existence of a more advanced shell.

Fair enough. Just wondering how aggressive we could've gotten.

Dockerfile
jef: Oh -- I didn't know they were used in code. That makes sense. Linking to what you've described.

https://github.com/airsonic/airsonic/blob/bd83229345976bd92446a02c30656adbb0de13a9/install/docker/run.sh#L19-L22

Thanks!

Dockerfile
jef: > But of course, that's just my subjective opinion.

I guess mine too 🤣

Dockerfile
pulsar256: agree with > JDK8, and I am familiar with adoptopenjdk. Perhaps we could go all the way to JDK11 since it has LTS?

Dockerfile
pulsar256: not sure we are on the same page. I do not intend to remove the unit tests which are run implicitly by the `package` goal. 

Skipping the tests does only make sense if the release (tagging) is triggered by something like `mvn release` which must run tests and ideally also the integration tests before tagging a release. This is currently out of this PR`s scope. 

May I resolve this thread or am I missing a point here?

Dockerfile
pulsar256: > ENVs don't commit new layers to docker image.

They do not add any size to the image, but they do add layers:

```
airsonic (feature/docker-build ✘)✹ ᐅ docker history 1e9339ec6607
IMAGE               CREATED              CREATED BY                                      SIZE                COMMENT
(...)
f33cbef52b9b        About a minute ago   /bin/sh -c #(nop)  ENV JVM_HEAP=256m            0B
69fc464b7099        About a minute ago   /bin/sh -c #(nop)  ENV UPNP_PORT=4041           0B
d0a91aa6ecfc        About a minute ago   /bin/sh -c #(nop)  ENV CONTEXT_PATH=/           0B
89e7e6bb414f        About a minute ago   /bin/sh -c #(nop)  ENV AIRSONIC_DIR=/airsonic   0B
0533c1173dc5        About a minute ago   /bin/sh -c #(nop)  ENV AIRSONIC_PORT=4040       0B
(...)
```

* thumbs up:  for readability and each `ENV` on a seperate line
* thumbs down: collapse `ENV` on one line.

Dockerfile
randomnicode: yes you may resolve this.

I was just pointing out that I don't think this could be part of the release process, it's an independent alternative way to build.

> I do not intend to remove the unit tests which are run implicitly by the package goal.

Why not? Since this is not part of any release, why have unit tests running? Or run unit but skip the integration-tests?

The release process that'll build it's own image, won't need this Dockerfile, because it already has access to the artifact (and doesn't need to compile it again). Its simplified Dockerfile merely copies the artifact into a base container.

Dockerfile
randomnicode: At some point in time we might have to convince the maintainers here to move to 11, but I don't know if that's something they'd be amenable to in this pr

Dockerfile
randomnicode: They're temporary intermediate images, they don't actually add layers.

Here's the official documentation:
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/
```
In older versions of Docker, it was important that you minimized the number of layers in your images to ensure they were performant. The following features were added to reduce this limitation:

Only the instructions RUN, COPY, ADD create layers. Other instructions create temporary intermediate images, and do not increase the size of the build.
```

Dockerfile
jef: Did we want to use the [`adoptopenjdk/openjdk8-openj9`](https://hub.docker.com/r/adoptopenjdk/openjdk8-openj9/) image? At least it stayed on Java 8. OpenJ9 is a faster, reliable, OSS alternative to Oracle's JVM.

Dockerfile
jef: Looks like this one is resolved.

Dockerfile
jef: If we split into new lines and keep the variables, looks like this can be resolved.

Dockerfile
pulsar256: I did run some tests, `adoptopenjdk/openjdk9:alpine` requires workarounds for https://github.com/AdoptOpenJDK/openjdk-docker/issues/75, `adoptopenjdk/openjdk11:alpine` and `adoptopenjdk/openjdk8-openj9:alpine` work fine. 

Going to use `adoptopenjdk/openjdk8-openj9:alpine` in this PR and run more tests with adoptopenjdk/openjdk11:alpine on my personal instance.

Dockerfile
pulsar256: adapted formatting, did the same to the VOLUME statements.

jef:
Love all the changes @pulsar256!

randomnicode:
After some consideration, I've actually changed my mind about the utility of this PR, for a particular reason. I think it has value beyond the stand-alone utility.

The current build process here just creates a war file (running unit tests against individual unpacked classes in the process), puts it inside a docker container, and runs some integration tests against the container.

While that's fine, there is a penalty overhead incurred for running a fully compressed and packaged war file (with dependencies), which is particularly more noticeable inside a docker container. The dependency artifacts inside a docker container should be in their non packaged form.

That necessitates two separate builds
- one for the docker image which creates unpackaged artifacts that are copied in to the docker image
- one for the war file release

Here is what Spring has tried to do to address the issue: https://spring.io/blog/2020/01/27/creating-docker-images-with-spring-boot-2-3-0-m1

In short:
- Change this docker file to package a layered war (might not be possible on this repo with an older version of Spring)
- Copy those dependencies over separately and use that as your new optimized docker image

tesshucom:
Very interesting.
I think there is definitely a useful scene.

However, packaging has many purposes and I think it must be considered a cost to pay in some cases.
There are various Java mechanisms using packages, but I think that now it makes sense to give a signature to a single release file even if you exclude them.

How about brushing up a simple and basic flow first, and doing the derivation idea in a separate issues?

randomnicode:
Parts of the ideas in https://github.com/airsonic/airsonic/pull/1538#issuecomment-591084508 were used in the formulation of this PR:

https://github.com/airsonic-advanced/airsonic-advanced/pull/315

It uses a multi-stage build to explode the war contents and layers them into the docker image

Issue #1421, opened 2019-11-28
Toggle play queue by clicking on a link/button


fxthomas:
What do y'all think of adding something like ▼ (Hide) on the right next to the song count/duration label, and make this clickable so that the user can manually hide/close the play queue?

Play queue shown:

![Screenshot from 2019-11-28 22-34-54](https://user-images.githubusercontent.com/613594/69832591-8279ba80-122f-11ea-8375-77adfa8f559a.png)

Play queue hidden:

![Screenshot from 2019-11-28 22-33-52](https://user-images.githubusercontent.com/613594/69832599-932a3080-122f-11ea-830b-6a076c80f352.png)

eharris:
I like it.  My only (minor) concern is for the impact when users have a narrower window.

fxthomas:
Yup, it's a bit unfortunate... In any case Airsonic is not very usable anyway if your width is less than, say, 800-900 px.

jvoisin:
Other players (soundcloud, …) are doing this, and I like it.
But I think that it should be a bit more visible.

fxthomas:
Thanks bot, I think we can close it :wink:

PR #1539, opened 2020-02-16
Toggle play queue with a button

fxthomas:
This follows #1421 by adding an up/down toggle button on the right-side of the player bar.

The toggle button is only displayed when auto-hide is not set (should it be displayed at all times?).

Play queue shown:

Screenshot from 2020-02-16 16-47-55

Play queue hidden:

Screenshot from 2020-02-16 16-47-43

### DIFF
--- a/airsonic-main/src/main/resources/org/airsonic/player/theme/default_dark.properties
+++ b/airsonic-main/src/main/resources/org/airsonic/player/theme/default_dark.properties
@@ -64,3 +64,5 @@ viewAsListImage = icons/default_dark/view_as_list.png
 volumeImage = icons/default_dark/volume.png
 alertImage = icons/default_dark/alert.svg
 checkImage = icons/default_dark/check.svg
+playQueueShow = icons/default_dark/chevrons-up.svg
+playQueueHide = icons/default_dark/chevrons-down.svg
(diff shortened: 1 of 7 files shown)


randomnicode:
>The toggle button is only displayed when auto-hide is not set (should it be displayed at all times?).

Is autohide even needed if this feature is present? We could conceivably just take it out.

fxthomas:
> Is autohide even needed if this feature is present? We could conceivably just take it out.

Good idea. I personally wouldn't mind, but I think it is better if we have some input from users other than ourselves, so I created a poll here : https://www.reddit.com/r/airsonic/comments/f5yeis/poll_how_do_you_use_the_airsonic_play_queue.

randomnicode:
I like this change. Could animate it too, but I'm in favor of it.

Don't have an opinion about the iconography (others more visually inclined may have better opinions), but overall this looks pretty good to me.

eharris:
I still like having the option of auto-hide, but the toggle is a good improvement.

randomnicode:
A few things since this was merged on [`airsonic-advanced`](https://https://github.com/airsonic-advanced/airsonic-advanced) a while back.
- The height of the toggle hide is sometimes wonky. You may need to re-account for some of the pixels with the framesets heights
- The frameset height setter option itself has a bug where you're not rounding the height to the nearest pixel. At least on Chrome, setting the rows to something like `"*, 270.33333333"` results in a weird and seemingly arbitrary height. This bug is present here on the airsonic fork: https://github.com/airsonic/airsonic/blob/master/airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp#L148
It was fixed in this commit: https://github.com/airsonic-advanced/airsonic-advanced/commit/3d5e8c6c6a4b2123e46abe690b7d668f6c1f5676 as part of a faster table rendering commit

PR #1459, opened 2020-01-07
Avoid exceptions if Lucene index is not ready on first startup

fxthomas:
If the library is large enough, Airsonic finishes loading the
application before the first (automatic) scan is completed. The
exception is then shown on the home page until the scan completes.

Fixes #1402.

PR #1550, opened 2020-02-25
Fix not to perform clean up during scan

tesshucom:
It is follow-up of #1459.

After 015c7c5 apply, performing a cleanup during scan will cause the following error on all subsequent file scans:

```
        at org.airsonic.player.service.search.IndexManager.index(IndexManager.java:145) ~[classes!/:10.6.0-SNAPSHOT]
        at org.airsonic.player.service.MediaScannerService.scanFile(MediaScannerService.java:237) [classes!/:10.6.0-SNAPSHOT]
        at org.airsonic.player.service.MediaScannerService.scanFile(MediaScannerService.java:244) [classes!/:10.6.0-SNAPSHOT]
        at org.airsonic.player.service.MediaScannerService.scanFile(MediaScannerService.java:244) [classes!/:10.6.0-SNAPSHOT]
        at org.airsonic.player.service.MediaScannerService.doScanLibrary(MediaScannerService.java:183) [classes!/:10.6.0-SNAPSHOT]
        at org.airsonic.player.service.MediaScannerService.access$000(MediaScannerService.java:50) [classes!/:10.6.0-SNAPSHOT]
        at org.airsonic.player.service.MediaScannerService$1.run(MediaScannerService.java:151) [classes!/:10.6.0-SNAPSHOT]
```

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/MusicFolderSettingsController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/MusicFolderSettingsController.java
@@ -101,29 +101,41 @@ public class MusicFolderSettingsController {
     }
 
 
-    private void expunge() {
+    private synchronized void expunge() {
 
-        // to be before dao#expunge
         MediaLibraryStatistics statistics = indexManager.getStatistics();
-        if (statistics != null) {
+
+        /*
+         * indexManager#expunge depends on DB delete flag.
+         * For consistency, clean of DB and Lucene must run in one block.
+         *
+         * Lucene's writing is exclusive.
+         * This process cannot be performed
+         * while during scan or scan has never been performed.
+         */
+        if (statistics != null && !mediaScannerService.isScanning()) {
+
+            // to be before dao#expunge
             LOG.debug("Cleaning search index...");
             indexManager.startIndexing();
             indexManager.expunge();
             indexManager.stopIndexing(statistics);
             LOG.debug("Search index cleanup complete.");
-        } else {
-            LOG.warn("Missing index statistics - index probably hasn't been created yet. Not expunging index.");
-        }
 
-        LOG.debug("Cleaning database...");
-        LOG.debug("Deleting non-present artists...");
-        artistDao.expunge();
-        LOG.debug("Deleting non-present albums...");
-        albumDao.expunge();
-        LOG.debug("Deleting non-present media files...");
-        mediaFileDao.expunge();
-        LOG.debug("Database cleanup complete.");
-        mediaFileDao.checkpoint();
+            // to be after indexManager#expunge
+            LOG.debug("Cleaning database...");
+            LOG.debug("Deleting non-present artists...");
+            artistDao.expunge();
+            LOG.debug("Deleting non-present albums...");
+            albumDao.expunge();
+            LOG.debug("Deleting non-present media files...");
+            mediaFileDao.expunge();
+            LOG.debug("Database cleanup complete.");
+            mediaFileDao.checkpoint();
+
+        } else {
+            LOG.warn("Index hasn't been created yet or during scanning. Plese execute clean up after scan is completed.");
+        }
     }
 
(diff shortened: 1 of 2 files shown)


airsonic-main/src/main/java/org/airsonic/player/service/search/IndexManager.java
```diff
@@ -313,8 +313,10 @@ private void stopIndexing(IndexType type, MediaLibraryStatistics statistics) {
                 }
             } catch (IndexNotFoundException e) {
                 LOG.debug("Index {} does not exist in {}, likely not yet created.", indexType.toString(), indexDirectory.getAbsolutePath());
+                return null;
             } catch (IOException e) {
                 LOG.warn("Failed to initialize SearcherManager.", e);
+                return null;
```
tesshucom: After the exception is caught, subsequent processing should not be performed.
As the logs show, it seems that references to already healthy Serchers are being dropped.

airsonic-main/src/main/java/org/airsonic/player/controller/MusicFolderSettingsController.java
```diff
@@ -105,14 +105,14 @@ private void expunge() {
 
         // to be before dao#expunge
         MediaLibraryStatistics statistics = indexManager.getStatistics();
-        if (statistics != null) {
+        if (statistics != null && !mediaScannerService.isScanning()) {
```
tesshucom: Executes exclusive control in more large block.
You can separate separate the statuses for log, but that's enough.

tesshucom:
After applying the patch, if you perform a cleanup during the scan, the index will not be deleted as follows:

```
2020-02-25 10:00:41.398  INFO --- o.a.p.s.MediaScannerService              : Scanned media library with 250 entries.
2020-02-25 10:00:50.279  WARN --- o.a.p.c.MusicFolderSettingsController    : Missing index statistics - index probably hasn't been created yet or during scanning. Not expunging index.
2020-02-25 10:01:14.180  INFO --- o.a.p.s.MediaScannerService              : Scanned media library with 500 entries.
```

tesshucom:
After applying the patch, if you perform a cleanup during the scan, the index will not be deleted as follows:

```
2020-02-25 10:00:41.398  INFO --- o.a.p.s.MediaScannerService              : Scanned media library with 250 entries.
2020-02-25 10:00:50.279  WARN --- o.a.p.c.MusicFolderSettingsController    : Missing index statistics - index probably hasn't been created yet or during scanning. Not expunging index.
2020-02-25 10:01:14.180  INFO --- o.a.p.s.MediaScannerService              : Scanned media library with 500 entries.
```

tesshucom:
It may be a bit controversial.
Better solution might be another.

 - Database cleanup can be performed concurrently with scanning. Since Subsonic. DB is good at such behavior.
 - Lucene Index must be exclusive if there is a writing thread.

This fix prevents malfunctions due to errors.
It is necessary in itself.

However, this causes the case of "clean up only the DB".
Unnecessary data in the Lucene index will not be deleted in the next Cleanup.
This is because the Lucene index Cleanup relies on the delete flag in the DB.

___

Even if unnecessary data remains in Lucene, it will not immediately cause any adverse effects.
But not consistent.

As a better suggestion, the checks can be improved a bit more.
I suggest that prevents the entire cleanup process from running during a scan.

Performing a complete cleanup process after scanning can help ensure consistency.

tesshucom:
Fix not to perform cleanup during scan.

The PRs related to data consistency between Lucene and DB are:

 - Move index meta information https://github.com/airsonic/airsonic/pull/1262
 - Expunge search index https://github.com/airsonic/airsonic/pull/1264

I think this PR will be a modification adjacent to the contents of these PRs.

Important internally.
But from the user's point of view, the functionality does not change.
If released at the same time as a Lucene update that has already been merged into master, the Index will be initialized.
Extra data will be gone.

jvoisin:
Please fix the style issue and we're good to go :)

tesshucom:
I revised that according to your advice.
The style check seems to be cleared on my local machine.

fxthomas:
Let's merge this!

Issue #1549, opened 2020-02-23
NPE on System check page


tesshucom:
## Problem description

An error page is displayed when open the system check page from the first launch to the end of the scan.

### Steps to reproduce

1. Start the server by specifying the music directory where the song exists.
2. Opens the system check page during the scan.
3. An error page is displayed.

```
  java.lang.NullPointerException
	at org.airsonic.player.controller.InternalHelpController.gatherIndexInfo(InternalHelpController.java:263)
	at org.airsonic.player.controller.InternalHelpController.handleRequestInternal(InternalHelpController.java:238)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.lang.reflect.Method.invoke(Unknown Source)
	at org.springframework.web.method.support.InvocableHandlerMethod.doInvoke(InvocableHandlerMethod.java:205)
	at org.springframework.web.method.support.InvocableHandlerMethod.invokeForRequest(InvocableHandlerMethod.java:133)
	at org.springframework.web.servlet.mvc.method.annotation.ServletInvocableHandlerMethod.invokeAndHandle(ServletInvocableHandlerMethod.java:97)
	at org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerAdapter.invokeHandlerMethod(RequestMappingHandlerAdapter.java:854)
	at org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerAdapter.handleInternal(RequestMappingHandlerAdapter.java:765)
	at org.springframework.web.servlet.mvc.method.AbstractHandlerMethodAdapter.handle(AbstractHandlerMethodAdapter.java:85)
	at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:967)
	at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:901)
	at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:970)
	at org.springframework.web.servlet.FrameworkServlet.doGet(FrameworkServlet.java:861)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:687)
	at org.springframework.web.servlet.FrameworkServlet.service(FrameworkServlet.java:846)
[...]

PR #1555, opened 2020-02-27
Avoid NPE on System check page

tesshucom:
#1549

When applied to the master, the display before scanning is as follows :

image

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/InternalHelpController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/InternalHelpController.java
@@ -35,6 +35,7 @@ import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.input.ReversedLinesFileReader;
 import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.index.IndexReader;
+import org.apache.lucene.search.IndexSearcher;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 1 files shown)


airsonic-main/src/main/java/org/airsonic/player/controller/InternalHelpController.java
```diff
@@ -257,14 +258,18 @@ private void gatherScanInfo(Map<String, Object> map) {
     private void gatherIndexInfo(Map<String, Object> map) {
         SortedMap<String, IndexStatistics> indexStats = new TreeMap<>();
         for (IndexType indexType : IndexType.values()) {
-            try (IndexReader reader = indexManager.getSearcher(indexType).getIndexReader()) {
```
tesshucom: IndexReader is closable, but is indirectly managed by Searcher.
Searcher resource management uses release functions.

airsonic-main/src/main/java/org/airsonic/player/controller/InternalHelpController.java
```diff
@@ -257,14 +258,18 @@ private void gatherScanInfo(Map<String, Object> map) {
...
+            stat.setName(indexType.name());
+            indexStats.put(indexType.name(), stat);
+            if (searcher != null) {
+                IndexReader reader = searcher.getIndexReader();
                 stat.setCount(reader.numDocs());
                 stat.setDeletedCount(reader.numDeletedDocs());
-                indexStats.put(indexType.name(), stat);
-            } catch (IOException e) {
-                LOG.debug("Unable to gather information about {} index", indexType.name(), e);
+                indexManager.release(IndexType.ALBUM, searcher);
```
tesshucom: Strictly speaking, it doesn't seem to need to be released here. 
But in the worst case it just does nothing.

tesshucom:
Off-topic.
Null-Return is a unkind method for those unfamiliar with handling.
when reviewing Lucene updates, we also considered making ``indexManager.getSearcher`` ```Optional```.
If you are concerned about these NPEs in the future, you can change them to ``Optional`` in another topic.

The current implementation takes into account the fact that the exception implementation is somewhat special and that all the parts used go up one block.

airsonic-main/src/main/java/org/airsonic/player/controller/InternalHelpController.java
fxthomas: Shouldn't we keep it under the try-with-resources block anyway if it's closeable?

tesshucom:
Yes, SearcherManager manages it automatically.
Get Searcher from SearcherManager and release Searcher after use.
When the index is updated, Reader is updated via SearcherManager.
These are done in IndexManager.

In legacy code, IndexReader was not closed at all, so it was rewritten to modern Lucene management.

tesshucom:
Yes, SearcherManager manages it automatically.

Get Searcher from SearcherManager and release Searcher after use.
When the index is updated, Reader is updated via SearcherManager.
No explicit close from outside is required.

airsonic-main/src/main/java/org/airsonic/player/controller/InternalHelpController.java
tesshucom: Fixed because IndexType was specified incorrectly.

IndexType.ALBUM -> indexType

fxthomas:
Let's merge it, the only failed thing here is the Tomcat CVE that was fixed earlier.

PR #1565, opened 2020-03-02
Disable now defunct chartlyrics lookups

eharris:


eharris:
Actually the conditional is absolutely required, since the compiler doesn't allow unreachable code.

randomnicode:
Was the fact that chartlyrics is down confirmed?

It doesn't seem to be.

Ping checks all pass, it is returning search results also.
It doesn't seem to have the most updated song selection, but that's not the same as saying it's down?

eharris:
You're right, it does seem to be back up now.  All I know is it was down (completely unresponsive) every time I checked it for a very long time.

randomnicode:
Maybe they saw this PR get merged and that spurred them into action ;)

fxthomas:
It appears to work fine for me (e.g. http://localhost:4040/lyrics.view?artistUtf8Hex=41432f4443&songUtf8Hex=426c61636b20496365 shows lyrics just fine on a dev server), let's revert it?

PR #1585, opened 2020-03-17
Change to new shuffle generator

eharris:
This replaces PR #1239, taking the original code and adding requested cleanups.


fxthomas:
Looks good to me.

PR #1594, opened 2020-03-28
Update commons-configuration2 2.6 -> 2.7 (fixes CVE)

fxthomas:
Fixes the current CVE error on master.


fxthomas:
The build still fails on Docker integration tests JDK11. I have no idea why, does anybody know?

tesshucom:
According to the message, the Docker plugin seems to be the cause...?
But #1503 seems to be fine.
Research is needed.

This error does not appear in my clone because cucomber is stopped.
(Not a static dependency, but a runtime error when running Docker.)

PR #1601, opened 2020-04-03
Change the default naming convention for podcasts

G-regL:
It used to take the filename from the episode URL, aren't very useful for most cases. Now it takes it's name from the channel title,  Episode publishing date, and episode title 'channelTitle - publishDate - episodeTitle'.

Guidelines for Contributing
---------------------------


airsonic-main/src/main/java/org/airsonic/player/service/PodcastService.java
```diff
@@ -657,10 +657,11 @@ private synchronized File getFile(PodcastChannel channel, PodcastEpisode episode
 
         File channelDir = getChannelDirectory(channel);
 
-        String filename = StringUtil.getUrlFile(episode.getUrl());
+        String filename = channel.getTitle() + " - " + episode.getPublishDate().toString() + " - " + episode.getTitle();
         if (filename == null) {
```
fxthomas, line 661: Isn't that never null in this case?

airsonic-main/src/main/java/org/airsonic/player/service/PodcastService.java
```diff
@@ -657,10 +657,11 @@ private synchronized File getFile(PodcastChannel channel, PodcastEpisode episode
 
         File channelDir = getChannelDirectory(channel);
 
-        String filename = StringUtil.getUrlFile(episode.getUrl());
+        String filename = channel.getTitle() + " - " + episode.getPublishDate().toString() + " - " + episode.getTitle();
```
fxthomas, line 660: I know it wasn't there before, but while we're at it, I think we should make sure that `episode.getTitle()` doesn't return something ridiculously long. Windows systems still have a lot of issues with long paths, so I can definitely see it being an issue here.

G-regL: Yeah, good point!

My suggestion would be to trim it once it's concatenated, so that, for the most part, only the episode title gets cut off if it's too long.

fxthomas: Sounds good to me!

airsonic-main/src/main/java/org/airsonic/player/service/PodcastService.java
fxthomas: In this case, let's remove the "if" case, less code means less bugs ;)

fxthomas:
Looks good to me! Can I trouble you with a small rebase so that Travis is happy? Then we can merge!

G-regL:
Yeah, against which branch?

fxthomas:
Ah, sorry, against `master`! ;)

G-regL:
Just did a rebase! Any chance this will get included in the next release?

fxthomas:
I think it's simple enough to be included in 11.0.

I tested it on my instance over the last week, and noticed that there might be two additional issues (sorry! :wink:):

* The default format for Date's `toString` is fairly long, locale-dependent, and doesn't allow sorting easily (something like `Sun Apr 19 14:08:21 CEST 2020`), I think it might be better to use an ISO format like `2020-04-19`
* Two episodes having the same title and date is unlikely, but it's going to be an issue with duplicate files. Can you add the episode ID to the filename after the date?

The final format might look like:

```
Channel Title - 2020-04-19 - 123 - Episode Title That May Be Very Long And Truncat.mp3
```

G-regL:
This latest commit should do that (at least based on my Google-fu, and ultra-basic level of Java knowledge 😄).

fxthomas:
Adding an integer from `getId` to a string may not work directly (maybe using `String.format` might be needed).

Can't test right now, will do when I can!

fxthomas:
Nevermind, this works fine! Let's merge it :wink:

PR #1604, opened 2020-04-05
Fix Docker errors on JDK11

fxthomas:


fxthomas:
\o/

randomnicode:
We do realize that this library is INACTIVE, right? It says it on its own webpage: https://github.com/spotify/docker-client

It's not even used internally by the people who wrote it themselves. And they're not accepting any bugfixes either.

I'd suggest moving to a different, maintained library?

fxthomas:
I plan on looking at what you proposed in https://github.com/airsonic/airsonic/issues/1240#issuecomment-595853829, but I'm not very familiar with this part of the build so I'm a little uncomfortable doing big changes without understanding what they do.

In the meantime, this is a quick fix. If nobody (@jvoisin @eharris?) sees any issue with this I'll go ahead and merge it.

randomnicode:
Without knowing the build error, I don't know if I'm in a position to comment, but looks okay superficially.

> I plan on looking at what you proposed in #1240 (comment), but I'm not very familiar with this part of the build so I'm a little uncomfortable doing big changes without understanding what they do.

Break stuff. Delete and start over.
Only way to learn to put it back together. Can't keep avoiding big stuff all the time. Or we'll just be relegated to minor version bumps.

The comment in #1240 does way more than replace this particular dependency. It removes and replaces all of cucumber with maven's own failsafe, which from my recollection wasn't amenable to @muff1nman for this codebase, so you'll need to have a conversation with him if you want to actually do it or risk having it stuck in PR purgatory here.

The code itself is like 3 files. Instead of the PR, I suggest looking at the repo directly at the integration-test module, which is like all of 3 files.

The key is instead of using these docker client libraries to communicate with the docker and stand up containers in code and then deal with actual communication with the containers also, just have maven set up a container. Then your actual test code becomes agnostic to whatever you're trying to connect to, it just cares about a url, that's all. It could be docker, it could be an external server whaever.

fxthomas:
We just have different ways of working, I take small steps and try to not break things in a public repository. But oh well, you will probably be insanely irritated by my comment, and we will probably have to agree to disagree :wink:

randomnicode:
> We just have different ways of working, I take small steps and try to not break things in a public repository.

We all break things. Even in public repos. No one tries to break things intentionally. The question really isn't about the breakage, because it happens and _will_ happen regardless. Regarding small steps, all big changes are made up of small steps too. There are just more of them and the rate of iteration is much faster.

The point becomes that fear of breakage shouldn't cause an aversion to innovation or slow the rate of progress, especially in source code, because the consequence of a misstep, especially with versioning system is quite minimal. If this was a product, it still wouldn't be a reason because there would be qa, staging and production environments with feature flags and split testing to test out the implications with. Even there, the _source_ innovation isn't stemmed. In fact, all those environments exist to gate the rollout, _without_ affecting the rate of progress at the source level.

> But oh well, you will probably be insanely irritated by my comment

Not at all. I think online tends to mask true temperaments. My apologies if I've given the impression that I'm irritated with your comments.

Let me know if you have questions about the changes in the integration module while you're looking at it, if you're having a difficult time understanding it or if you think there is a design decision I should've made differently.

fxthomas:
@eharris @jvoisin Thanks for the testing! Let's merge this!

fxthomas:
@randomnicode I did interpret you previous comments as being a tad annoyed, so I apologize for being a bit curt myself if that wasn't intended. Communication is indeed hard!

And while I know I tend to be very conservative in how I work, thanks for the lot of work you're doing with all the upgrades, this is really helpful in eventually moving things in the right direction.

PR #1613, opened 2020-04-13
Bump jquery from 3.4.0 to 3.5.0

fxthomas:
See https://github.com/airsonic/airsonic/issues/1438#issuecomment-612991587, jQuery 3.5.0 has been released and this fixes a CVE.

Issue #310, opened 2017-03-20
API endpoint for lyrics not working anymore


issuemover631:
<a href="https://github.com/Whisprin"><img src="https://avatars1.githubusercontent.com/u/3153136?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [Whisprin](https://github.com/Whisprin)**
_Monday Mar 20, 2017 at 21:31 GMT_
_Originally opened as https://github.com/Libresonic/libresonic/issues/308_

----

Libresonic uses [ChartLyrics](http://www.chartlyrics.com/) as a lyrics provider. The [API](http://api.chartlyrics.com/apiv1.asmx) isn't working anymore.

We need to find an alternative lyrics provider and replace ChartLyrics.
Ideally we implement multiple providers and give the user the option to enable, disable and order them.

The Android app [QuickLyrics](https://github.com/geecko86/QuickLyric) (written in Java) implements [many different providers](https://github.com/geecko86/QuickLyric/tree/master/QuickLyric/src/main/java/com/geecko/QuickLyric/lyrics).

PR #1614, opened 2020-04-14
Revert "Disable now defunct chartlyrics lookups"

fxthomas:
This reverts commit 2022e98df9784c1efc48e0adcb5ddc472f7901eb (from #1565).

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/ajax/LyricsService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/ajax/LyricsService.java
@@ -66,13 +66,6 @@ public class LyricsService {
      */
     public LyricsInfo getLyrics(String artist, String song) {
         LyricsInfo lyrics = new LyricsInfo();
-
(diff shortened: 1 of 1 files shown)


eharris:
Looks like jQuery has a vulnerability breaking the build.

Given the outdated/broken/poorly maintained dependency that Chartlyrics seems to have become (see #310, #1144, plus the original offline condition that instigated the PR that this reverts), perhaps rather than just re-enabling it, someone is interested in fixing/updating the lyrics fetch?

fxthomas:
jQuery is going to be fixed by https://github.com/airsonic/airsonic/pull/1613! :wink: 

For the short term, I would rather not remove the feature without warning since it seems to be at least working "sometimes", so I would like to keep it for now if you're okay with it?

For the future, I won't work on it myself (not really interested in the feature), but I'm thinking of advertising these kinds of "easy" issues a bit more (perhaps a post on Reddit and links on the documentation?) so that people who are new to the project can come and help.

fxthomas:
Let's merge this so we can get the 10.6 release over with!

PR #1617, opened 2020-04-15
Upgrade JRE from 8 to 11 in Docker image

fxthomas:


randomnicode:
Needs a little more work. Here's the issue:
Because the base container is different, it has a more updated version of the transcoders (lame and ffmpeg).
This shows up when it transcodes a particular stream into mp3 because the stream mentions its encoder's version. Because you're comparing it to a file that was put in lfs encoded with the old version of ffmpeg, it doesn't match the transcoded file at the spot where the file states its encoder's version

You may check this by using a hexeditor and compare the two files (expected and produced response). You'll notice they differ right after the ID3 header at the encoder version.

To fix, just replace the file in the lfs (expected) with the new file being generated.

Reference, here's the PR on airsonic-advanced: https://github.com/airsonic-advanced/airsonic-advanced/pull/120/files
Note it updates the files (it doesn't use lfs), but only for the files that need transcoding. The mp3 doesn't need to be transcoded and therefore does not use ffmpeg, thus its encoder version remains the same.

fxthomas:
@airsonic/maintainers what do you think of dropping JRE <11 for Airsonic 11.0? This will allow us to upgrade everything and profit from performance improvements. Any blocking points?

jvoisin:
What about Debian stable?

jooola:
```
jo@jofix: ~ $ sudo apt-cache policy openjdk-11-jre
[sudo] password for jo: 
openjdk-11-jre:
  Installed: 11.0.7+10-3~deb10u1
  Candidate: 11.0.7+10-3~deb10u1
  Version table:
     11.0.7+10-3 50
         50 http://deb.debian.org/debian sid/main amd64 Packages
 *** 11.0.7+10-3~deb10u1 500
        500 https://deb.debian.org/debian-security buster/updates/main amd64 Packages
        100 /var/lib/dpkg/status
     11.0.6+10-1~deb10u1 500
        500 https://deb.debian.org/debian buster/main amd64 Packages
```

fxthomas:
Added a [post on Reddit](https://www.reddit.com/r/airsonic/comments/gbku9g/poll_is_anybody_unable_to_upgrade_to_java_11/) so that we can gather more info.

From what I can see most distributions already jumped to it, but there might be some widely used NAS devices (for example) on old distributions. I don't want to hold off the upgrade indefinitely for them, but if the number of users is large enough that might be a reason to hold off the upgrade before 11.0 is released.

tesshucom:
> Any blocking points?

It's a very niche, but there are times when you intentionally use a 32bit-JVM. Performance improves but doubles memory so. However, I think upgrades are unavoidable and acceptable. The time for the transition is now neither early nor late. I think it's kind enough because you've notified the user in advance.

randomnicode:
I must be really confused here because I don't see how this PR has anything to do with bumping your minimum JRE requirements.

The whole point of the container is to isolate the outside from whatever's running inside.

> From what I can see most distributions already jumped to it, but there might be some widely used NAS devices (for example) on old distributions.

Doesn't matter. If they're running Docker, that's all they know or care about. The specifics of the JRE inside the container are irrelevant. You only talk to a port.

> It's a very niche, but there are times when you intentionally use a 32bit-JVM. Performance improves but doubles memory so. 

Again irrelevant. They're not running 32-bit JVM. They're running a container.

Bumping your Docker JRE has no effect on your war compilation and thus the bytecode being generated. Any JRE equal to or above your war compilation should be sufficient to run the software.

Again, to revisit:
- You compile in 8 (this PR doesn't change that).
- You _used_ to run that compiled war in a JRE 8 environment in the container. That should work fine
- You are now switching to running the same war in a JRE 11 environment in the container. That should work fine also
- Nothing else has changed

So someone explain to me why there's talk of minimum JRE versions being bumped in this PR and Reddits being posted?

tesshucom:
> You compile in 8 (this PR doesn't change that).

As long as this is maintained there is no opposition. I think it's a trivial fix.

fxthomas:
Probably just bad communication on my part, I'm trying to figure out if both this one and #1618 can be used safely. Sorry for the noise.

tesshucom:
The compiler version up needs notification.
The fact that asm cannot be updated is truly painful ...

PR #1618, opened 2020-04-15
Bump compiler version from Java 8 to Java 11

fxthomas:
I don't see any big risk in upgrading, will of course test the newly built packages on my instance.

The main risk I see is that we're effectively forcing everyone to switch to Java 11, and some platforms might still be on Java 8. Anybody sees an important platform where this is the case?


fxthomas:
Needs https://github.com/airsonic/airsonic/pull/1606 first apparently, rebased on it.

hboetes:
You could consider releasing binaries for both 11 and 8 and see what the download stats say. For example my tomcat uses v8, but v11 is already available on OpenBSD. Although I bet v11 tomcat will be released soon.

fxthomas:
@hboetes That's a good point, do you think there are some Tomcat users stuck on v8 for now?

I would rather avoid releasing two versions to be honest, that's a bit of a hassle. If we have to I'd rather delay the change and stay on JRE8 until things are looking good for JRE11.

hboetes:
Well I was stuck on v8, until I reported a bug and got it fixed. If you are going to keep using V8 nobody has the option to move forward. If you move to V11 some people might not be able to upgrade and you will get quite some grumpy users and bug reports. But for you maintaining 2 versions is a hassle.

In that case I'd say move to V11. Prod people to move on. :-)

fxthomas:
I haven't seen a lot of reports of compatibility issues about Java 11, and no issues on my instance for a few weeks. I guess we can move forward on this?

hboetes:
Sure, let's see what happens.

muff1nman:
I think we should probably keep java 1.8 target for a bit longer just to play it safe.

PR #1535, opened 2020-02-15
Bump cucumber from 4.2.0 to 5.3.0

jvoisin:
- Bump the version of cucumber
- Remove the deprecated cucumber-parallel plugin

PR #1619, opened 2020-04-15
Replace deprecated integration test plugins (Cucumber, Docker)

### LINKED ISSUE #1240: Integration tests: cucumber-jvm-parallel-plugin not supported

fxthomas:
This imports the relevant parts of airsonic-advanced/airsonic-advanced#113 from airsonic-advanced/airsonic-advanced@4601c7e4 in order to remove old and unmaintained integration test plugins for Maven.

Reviewed and tested on my machine, and the previous tests are kept as before.

Fixes #1240, replaces #1535.


fxthomas:
@randomnicode :wink:

integration-test/src/test/java/org/airsonic/test/StreamIT.java
```diff
@@ -0,0 +1,49 @@
...
+                .findAny()
+                .map(x -> x.getId())
+                .orElseThrow(() -> new RuntimeException("no media file id matched"));
+
+        byte[] fromServer = Scanner.getMediaFileData(mediaFileId);
+        String expectedBodyResource = String.format("/blobs/stream/" + file + "/responses/1.dat");
+        byte[] expected = IOUtils.toByteArray(StreamIT.class.getResourceAsStream(expectedBodyResource));
+        if (!StringUtils.isBlank(System.getProperty("IntegrationTestsOutputDirectory", ""))) {
+            Path fromServerPath = Paths.get(System.getProperty("IntegrationTestsOutputDirectory", ""), file + ".dat");
+            FileUtils.writeByteArrayToFile(fromServerPath.toFile(), fromServer);
```
randomnicode, line 45: Any reason you want to write this file to disk? Are you doing something with it afterwards? I think the in-memory assertion below is enough for the test, no?

integration-test/src/test/java/org/airsonic/test/Scanner.java
```diff
@@ -0,0 +1,137 @@
...
+
+        assertThat(response.getBody()).hasSize((int) response.getHeaders().getContentLength());
+        return response.getBody();
+    }
+
+    @JsonRootName(value = "subsonic-response")
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    public static class SubsonicResponse extends Response {
+    }
```
randomnicode, line 136: This wrapper added because I just wanted to make do with Jackson instead of adding additional JAXB dependencies to deal with the incoming xml

integration-test/pom.xml
```diff
@@ -124,84 +83,152 @@
...
+                    <execution>
+                        <configuration>
+                            <!-- Skip integration tests by adding -Dfailsafe.skip to the Maven command-line -->
+                            <skip>${failsafe.skip}</skip>
+                            <systemPropertyVariables>
+                                <DockerTestingContainer>${docker.testing.container}</DockerTestingContainer>
+                                <DockerTestingHost>${docker.testing.host}</DockerTestingHost>
+                                <!-- This is the folder on the host (not on the container), used in org.airsonic.test.Scanner -->
+                                <DockerTestingDefaultMusicFolder>${docker.testing.default-music-folder}</DockerTestingDefaultMusicFolder>
+                                <IntegrationTestsOutputDirectory>${project.build.directory}/data/responses</IntegrationTestsOutputDirectory>
```
randomnicode, line 223: Not sure this guy is needed? You can just retrieve responses and compare them in-memory

integration-test/pom.xml
```diff
@@ -124,84 +83,152 @@
...
+                        <id>create-host-music-folder-for-docker</id>
+                        <phase>process-resources</phase>
                         <configuration>
-                            <finalName>unwanted</finalName>
-                            <classifier>unwanted</classifier>
+                            <tasks>
+                                <delete dir="${docker.testing.default-music-folder}"/>
+                                <mkdir dir="${docker.testing.default-music-folder}"/>
+                                <delete dir="${docker.testing.output-folder}"/>
+                                <mkdir dir="${docker.testing.output-folder}"/>
```
randomnicode, line 200: 👍

integration-test/pom.xml
```diff
@@ -124,84 +83,152 @@
...
+                                            <healthy>true</healthy>
+                                        </wait>
+                                        <env>
+                                            <DatabaseConfigType>${env.DatabaseConfigType}</DatabaseConfigType>
+                                            <DatabaseConfigEmbedDriver>${env.DatabaseConfigEmbedDriver}</DatabaseConfigEmbedDriver>
+                                            <DatabaseConfigEmbedUrl>${env.DatabaseConfigEmbedUrl}</DatabaseConfigEmbedUrl>
+                                            <DatabaseConfigEmbedUsername>${env.DatabaseConfigEmbedUsername}</DatabaseConfigEmbedUsername>
+                                            <DatabaseConfigEmbedPassword>${env.DatabaseConfigEmbedPassword}</DatabaseConfigEmbedPassword>
+                                            <DatabaseUsertableQuote>${env.DatabaseUsertableQuote}</DatabaseUsertableQuote>
+                                        </env>
```
randomnicode, line 167: Need to be careful here.

airsonic-advanced can [read properties from the env](https://github.com/airsonic-advanced/airsonic-advanced/pull/38), which airsonic can't. I doubt these are actually setting anything inside the container. You're not using an external db anyway, so they aren't needed anyway (which is probably why your tests are fine).

randomnicode:
Finally, here are @muff1nman 's thoughts when I pitched this change earlier to @jvoisin: https://github.com/airsonic/airsonic/pull/1535#issuecomment-589992137

which makes me think he's not amenable to this change, so you might want to discuss with him if you haven't already.

tesshucom:
The BDD perspective, itself is important.
However, looking over the reported Airsonic bugs, it seems that BDD is no longer working.

 - Cucumber has a high learning curve and is currently not maintained by anyone. It is a tool with no backward compatibility and a high update risk, and the maintenance cost is higher than the scenario addition cost. I have no intention of disregarding BDD, but it cannot be used effectively without a dedicated engineer.
 - The successor to Cucumber is a product called Turnip. Highly compatible with Capybara (UI test automation tool). So even if you use a type of tool like Cucumber, there may be better options now.
 - Adjacent proposals such as independence of integration test project, automation of Docker image creation, Github action, etc. are running side by side. Making integration tests more portable is beneficial to them. Perhaps.
 - We want to move forward rather than stagnant. Even if Cucumber has good points, it needs to be revisited if it becomes a barrier to many progress. Unless the cucumber is catching something that cannot be replaced.

Realizing BDD is not the ultimate goal.
I think it is better to change to a system that has a higher detection rate against cost.
While having an ideal, It is important to recognize that the human resources on this project are limited.

fxthomas:
> which makes me think he's not amenable to this change

That does seem like what @muff1nman is saying, I'll add him as a reviewer.

From the little I've seen Cucumber seems hard to debug, so I've always avoided looking too much into it unless I really needed to. I understand the updated system in this PR a bit better.

I'm going to push for this, but if he's strongly against it I won't pursue this further.

integration-test/src/test/java/org/airsonic/test/StreamIT.java
fxthomas: Mostly because I ran into the issue with the updated files (b/c of ffmpeg/codec changes in the Docker image) and needed to get the outputs to disk. I think it's easier to debug if we keep it?

integration-test/src/test/java/org/airsonic/test/Scanner.java
fxthomas: Can't we use the existing classes?

integration-test/pom.xml
fxthomas: See my other comment.

integration-test/pom.xml
fxthomas: Good point, I'll remove them for now.

fxthomas:
Rebased on master.

integration-test/src/test/java/org/airsonic/test/StreamIT.java
randomnicode: okay sure. usually these "debug" kind of things are kept around only if they're uploaded somewhere as debugging artifacts for CI pipelines. The idea is that they're useless if the tests pass, and if they don't, then a dev needs to change code anywhere, so it's not too difficult to just dump it locally.

but if this is your preference, that's fine. it'll just create and recreate some files locally whenever you run tests, not a big issue unless you're worried about the write cycles of your disk.

integration-test/src/test/java/org/airsonic/test/Scanner.java
randomnicode: It does use the existing class (*extends* Response). You also need a _binding_ library that binds xml to Java classes. You may use cxf. I thought it easier (since it was already in the classpath) to just use Jackson. But you need some annotations to account for some xsd elements, especially since the code generation doesn't seem to create self-sufficient classes (like mark an _XmlRootElement_ on Response). If you want, you can use a JAXB binder other than Jackson

muff1nman:
Sorry for my quietness in project as of late, but to add in my thoughts here.

My original hope for the BDD stuff was that given enough of the fundamentals were flushed out, non-developers that were interested in seeing their bugs fixed could write feature files instead of having to write java tests. In hindsight this hope was maybe a bit farfetched, but it was worth a try.  I also found in other experiences that feature files were a great way to decouple tests from the details on how the tests were implemented.

At the end of the day the most important thing is that the tests get written. Since I was the only person working on the Cucumber stuff, I'd say throw it out. With that said, I hope the spirit of these integration tests can remain. Namely that we are black box testing the binary and dependencies (ffmpeg) in a form closest to what users are dealing with. That looks to be the case so :+1:

fxthomas:
No problem re:quietness, we've all got a life and honestly things are being a bit weird for everyone at the moment. Hopefully you're doing alright!

Thanks for the details on this, since we all agree I'll wrap up this PR and get it merged.

Issue #1625, opened 2020-04-17
java.lang.UnsupportedOperationException: Not implemented for this format


HolyMacarony:
## Problem description
**Thank you all for this great project, it worked great so far!
So i started the latest version 10.6.0 today and the library scan fails at about 26500/~86000 files with the following stack trace:**

```
2020-04-17 02:35:12.332  INFO --- o.a.p.s.MediaScannerService              : Scanned media library with 26500 entries.
2020-04-17 02:35:14.694 ERROR --- o.a.p.s.MediaScannerService              : Failed to scan media library.

java.lang.UnsupportedOperationException: Not implemented for this format
        at org.jaudiotagger.audio.generic.GenericTag.getValue(GenericTag.java:191) ~[jaudiotagger-2.2.5.jar!/:na]
        at org.jaudiotagger.tag.wav.WavTag.getValue(WavTag.java:242) ~[jaudiotagger-2.2.5.jar!/:na]
        at org.jaudiotagger.tag.wav.WavTag.getFirst(WavTag.java:247) ~[jaudiotagger-2.2.5.jar!/:na]
        at org.airsonic.player.service.metadata.JaudiotaggerParser.getTagField(JaudiotaggerParser.java:128) ~[classes!/:10.6.0-RELEASE]
        at org.airsonic.player.service.metadata.JaudiotaggerParser.getRawMetaData(JaudiotaggerParser.java:99) ~[classes!/:10.6.0-RELEASE]
        at org.airsonic.player.service.metadata.MetaDataParser.getMetaData(MetaDataParser.java:47) ~[classes!/:10.6.0-RELEASE]
        at org.airsonic.player.service.MediaFileService.createMediaFile(MediaFileService.java:537) ~[classes!/:10.6.0-RELEASE]
        at org.airsonic.player.service.MediaFileService.updateChildren(MediaFileService.java:397) ~[classes!/:10.6.0-RELEASE]
        at org.airsonic.player.service.MediaFileService.getChildrenOf(MediaFileService.java:193) ~[classes!/:10.6.0-RELEASE]
        at org.airsonic.player.service.MediaScannerService.scanFile(MediaScannerService.java:240) [classes!/:10.6.0-RELEASE]
        at org.airsonic.player.service.MediaScannerService.scanFile(MediaScannerService.java:244) [classes!/:10.6.0-RELEASE]
[...]

PR #1626, opened 2020-04-17
Revert escription of a exception in JaudiotaggerParser 

tesshucom:
This is a revert for the #1625.
Due to lack of verification, we will also revert exception descriptions other than those reported.

It is a revert (patch) to instant recover the deterioration that cannot be untraceable by the user.
Another new exception description, logging and test addition (refactoring) should be done in another PR.

Note : CHANGELOG.md

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/metadata/JaudiotaggerParser.java
@@ -27,7 +27,6 @@ import org.jaudiotagger.audio.AudioFile;
 import org.jaudiotagger.audio.AudioFileIO;
 import org.jaudiotagger.audio.AudioHeader;
 import org.jaudiotagger.tag.FieldKey;
-import org.jaudiotagger.tag.KeyNotFoundException;
 import org.jaudiotagger.tag.Tag;
 import org.jaudiotagger.tag.images.Artwork;
 import org.jaudiotagger.tag.reference.GenreTypes;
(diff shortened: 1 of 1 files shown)


tesshucom:
We have created a minimal patch based on legacy so that people in need can apply it.
Please add as necessary, create a more appropriate commit and merge.
There is no problem if you leave it as it is.

```
try{
}catch(~Exception e) {  <=Add!
}catch(~Exception e) {  <=Add!
}catch(~Exception e) {  <=Add!
}catch(~Tthrowable e) {
}
```
For Java, If want to separate logs, just add them in exeption level order.

I'm having deja vu.
(In another commit, I feel like I had a similar story with randomnicode and  fxthomas...)

___

 - The reason why the exception design of JaudiotaggerParser is high is that it is a specification that reflects the product characteristics.
 - It's not laziness that JaudiotaggerParser has a huge try block, it should be. Perhaps the intention is to invalidate the previous processing if a more serious error occurs in the subsequent processing.

tesshucom:
I think this way.

 - If it's a regressive repair, having a test is not ''urgent''. Because it have a very long track record already. Of course, there should be tests. But it doesn't exist in legacy, and it's quite difficult to achieve. If realized, it will overlap with the work of the Jaudiotagger project.
 - After applying this PR (that is, after reverting to the traditional implementation), any further exception design change requires valid explanation and testing. Because a defect track record has already been created.

So, I think it's okay if you include this PR simply when you release 10.6.1.
After that, you need to be alert when a similar PR comes.

PR #978, opened 2019-04-05
Revert cling-core

tesshucom:
This has already been described in Issues #975.
As a final workaround.

PR #1635, opened 2020-04-18
Fix DLNA issues in 10.6

### LINKED ISSUE #1634: DLNA not working in 10.6
From https://www.reddit.com/r/airsonic/comments/g2zdx0/dlna_and_106/:

> I have been testing 10.6, but no matter how I fiddle the DLNA server remains invisible. I run Airsonic in a war container in windows (sush), and normally use bubbleupnp too send the tracks to my end point a NAD C338.
>
> I tried both running 10.6 in the old 10.51 folder,letting it use the old database and also a completely fresh install. Bubble couldnt see either install. Replacing the war with the old one and touching nothing else, the airsonic server was back.
>
> Maybe I am missing something obvious?. For now i'm back too 10.51 which works fine in my setup.

fxthomas:
Fixes #1634.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/UPnPService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/UPnPService.java
@@ -19,10 +19,10 @@
  */
 package org.airsonic.player.service;
 
+import org.airsonic.player.service.upnp.ApacheUpnpServiceConfiguration;
 import org.airsonic.player.service.upnp.CustomContentDirectory;
 import org.airsonic.player.service.upnp.MSMediaReceiverRegistrarService;
 import org.airsonic.player.util.FileUtil;
-import org.fourthline.cling.DefaultUpnpServiceConfiguration;
 import org.fourthline.cling.UpnpService;
 import org.fourthline.cling.UpnpServiceImpl;
 import org.fourthline.cling.binding.annotations.AnnotationLocalServiceBinder;
@@ -122,7 +122,7 @@ public class UPnPService {
     }
 
     private synchronized void createService() {
-        upnpService = new UpnpServiceImpl(new DefaultUpnpServiceConfiguration(SettingsService.getDefaultUPnPPort()));
+        upnpService = new UpnpServiceImpl(new ApacheUpnpServiceConfiguration());
 
         // Asynch search for other devices (most importantly UPnP-enabled routers for port-mapping)
         upnpService.getControlPoint().search();
--- a/airsonic-main/src/main/java/org/airsonic/player/service/UPnPService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/UPnPService.java
@@ -122,7 +122,7 @@ public class UPnPService {
     }
 
     private synchronized void createService() {
-        upnpService = new UpnpServiceImpl(new ApacheUpnpServiceConfiguration());
+        upnpService = new UpnpServiceImpl(new ApacheUpnpServiceConfiguration(SettingsService.getDefaultUPnPPort()));
 
         // Asynch search for other devices (most importantly UPnP-enabled routers for port-mapping)
         upnpService.getControlPoint().search();
--- a/airsonic-main/src/main/java/org/airsonic/player/service/upnp/ApacheUpnpServiceConfiguration.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/upnp/ApacheUpnpServiceConfiguration.java
@@ -1,0 +1,51 @@
+/*
+ This file is part of Airsonic.
+
+ Airsonic is free software: you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+
+ Airsonic is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with Airsonic.  If not, see <http://www.gnu.org/licenses/>.
+
+ Copyright 2016 (C) Airsonic Authors
+ Based upon Subsonic, Copyright 2013 (C) Sindre Mehus
+ */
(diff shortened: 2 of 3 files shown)


fxthomas:
I don't use DLNA, so this was a bit of a wild shot on a technology I don't understand well yet.

From what I understand : Cling (the library we use for UPnP/DLNA) does some weird hacks in `HttpUrlConnection` which don't play well with servlet containers like Tomcat (whether embedded or external). A workaround is available in Cling 2.0 in the form of helper classes (which were used in Airsonic before d626463), but is removed in Cling 2.1.

Relevant issues: #1505, #1381, #978, #975.

tesshucom:
As for the core of the problem, it's accurate you presented.
Previously I made a similar fix in the clone.
(There are more UPnP / DLNA users in Japan than in other countries. Therefore a careful approach was taken.)

https://github.com/tesshucom/jpsonic/commit/24d54fd36d5f89e93cc0ade31d14b9959da7cb73

Ephemeral port is the default in Cling.
(To be honest, the concrete value is 0, but it makes sense that Airsonic has no default value.)
Subsonic, Libresonic and Airsonic have been like that until now.

As you know, this has changed in 10.6.0.
I did not follow this and fixed it to allow port changes without breaking the legacy default design.

Demonstrating the cases in which this has a net effect is time consuming, but still a lower risk to existing users.
Because it's been working that way until now.
I've taken the definitive way to avoid having to verify on Windows, which has a unique auto-port configuration feature.
(Since the settings differ depending on the version, you have to verify it individually each time trouble occurs!)

I don't care how Airsonic handles the UPnP default port.
However, it has not been verified whether the default port shows a different value from the legacy server on a particular OS.

If you don't want to have any new validations or issues, then a switch like the one above.

airsonic-main/src/main/java/org/airsonic/player/service/upnp/ApacheUpnpServiceConfiguration.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    public ApacheUpnpServiceConfiguration(int streamListenPort) {
+        super(streamListenPort);
+    }
+
+    @Override
+    public StreamClient createStreamClient() {
+        return new StreamClientImpl(new StreamClientConfigurationImpl(Executors.newCachedThreadPool()));
+    }
```
randomnicode, line 53: This is again more or less the same as the parent. The only difference is that the parent provides its own custom executor and we provide a generic one.

airsonic-main/src/main/java/org/airsonic/player/service/UPnPService.java
```diff
@@ -122,7 +122,7 @@ private void startService() {
     }
 
     private synchronized void createService() {
-        upnpService = new UpnpServiceImpl(new DefaultUpnpServiceConfiguration(SettingsService.getDefaultUPnPPort()));
+        upnpService = new UpnpServiceImpl(new ApacheUpnpServiceConfiguration(SettingsService.getDefaultUPnPPort()));
```
randomnicode, line 125: I do not see much of a difference between this custom class and the parent class except for an request executor. I'm not convinced we know the issue.

randomnicode:
Note, I'm not denying there isn't an issue, or even that this doesn't fix it, just that it seems to me that the underlying root cause isn't understood at all.

In my mind, you may as well just fix it with a much shorter version of the PR. An almost one liner by overriding the createDefaultExecutorService() anonymously instead of creating a whole new separate class file:

```
upnpService = new UpnpServiceImpl(new 
        DefaultUpnpServiceConfiguration(settingsService.getUPnpPort()) {
            protected ExecutorService createDefaultExecutorService() {
                return Executors.newCachedThreadPool();
            }
        });
```

This seems to be the only change you're effecting by creating a separate class and overriding a bunch of methods.

airsonic-main/src/main/java/org/airsonic/player/service/UPnPService.java
randomnicode: Scratch that, I think I see what you're saying: the packages are different for the similarly named the classes

airsonic-main/src/main/java/org/airsonic/player/service/upnp/ApacheUpnpServiceConfiguration.java
randomnicode: Nevermind, the packages are different

airsonic-main/src/main/java/org/airsonic/player/service/upnp/ApacheUpnpServiceConfiguration.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Override
+    public StreamServer createStreamServer(NetworkAddressFactory networkAddressFactory) {
+        return new StreamServerImpl(new StreamServerConfigurationImpl(networkAddressFactory.getStreamListenPort()));
+    }
```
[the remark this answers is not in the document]
randomnicode: Different packages

randomnicode:
Yeah, my apologies, I stand completely corrected. I see the issue you're pointing out now.

In that case, this looks perfectly fine.

One minor nit would be to switch the cached threadpool to whatever the class's parent (Cling's custom threadpool) is using. I haven't looked in any detail, but no reason to deviate from the parent? (famous last words...)

On a side note, as you pointed, Cling is now unmaintained, and perhaps we should switch or stop providing DLNA (which is kinda getting old). A more recent fork of cling seems to be here: https://github.com/jupnp/jupnp

tesshucom:
We are very aware of that.
What is happening now is a reinstatement to deal with the feature degradation Instantly.
Also, it is an issue pointed out by the user .
It makes sense to eliminate noise and treat it as an incident response.

Other improvements can be treated as separate commits.
Blurring the problem of degradation and how to deal with it is not appropriate.

The reason why long-standing Airsonic maintainers recommend splitting (although this can be overkill) is that they are wary of such incidents.

fxthomas:
@tesshucom I think I will leave the default port for now (although you're right, it's maybe a commonly-used port, hopefully it won't cause any issues). Thanks for looking at this, merging!

fxthomas:
@randomnicode For now Cling doesn't cause issues with Tomcat, but I would be very wary of Tomcat upgrades while we stick to this.

PR #1659, opened 2020-04-25
Add support for MediaSession API

fxthomas:
This adds the currently-playing song to the notification area, as well as enables support for using media keys to control Airsonic playback from another application:

Screenshot from 2020-04-25 11-14-42

This works on Chrome out of the box, and the latest Firefox betas if the following preferences are set
to `true`:

* `dom.media.mediasession.enabled`
* `media.hardwaremediakeys.enabled`

Note that the latest Firefox betas have a [bug on Linux](https://bugzilla.mozilla.org/show_bug.cgi?id=1631608) where the song metadata isn't updated in the notification area, but it looks like it'll be fixed soon.


airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
```diff
@@ -216,7 +216,11 @@
         if (currentSong) {
             updateWindowTitle(currentSong);
             <c:if test="${model.notify}">
-            showNotification(currentSong);
+            if ('mediaSession' in top.playQueue.navigator) {
```
randomnicode, line 219: minor nit, why check the frame's navigator?
why not just the parent's frame's?

randomnicode:
lgtm otherwise

airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
fxthomas: MediaSession needs to be updated from the same frame/iframe as the `<audio>` element, and I figured having the same object used everywhere would reduce the risk of misunderstanding?

randomnicode: so then why not just `navigator`?
```
if ('mediaSession' in navigator) {
```

airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
```diff
@@ -832,6 +836,27 @@
...
+        if ('mediaSession' in top.playQueue.navigator) {
+            var metadata = new MediaMetadata({
+                title: song.title,
+                artist: song.artist,
+                album: song.album,
+                artwork: [
+                    // Adding the '&size=' argument will not work here because
+                    // the JWT token used to expose this URL requires all
+                    // parameters to stay exactly the same.
+                    { src: song.remoteCoverArtUrl, type: 'image/jpeg' },
```
randomnicode, line 849: two things here:
- you're not guaranteed a jpeg here, are you? it could be a png...
- why access the remote url, when this user is local? you don't need to go through the jwt, you've already authed this user

airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
```diff
@@ -216,7 +216,11 @@
         if (currentSong) {
             updateWindowTitle(currentSong);
             <c:if test="${model.notify}">
-            showNotification(currentSong);
+            if ('mediaSession' in top.playQueue.navigator) {
+                showMediaSessionMetadata(currentSong);
+            } else {
+                showNotification(currentSong);
+            }
```
randomnicode, line 223: are these two mutually exclusive?

airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
fxthomas: The cover art controller always creates a JPEG when asked for resized cover art.
Good point re:using the local cover art.

airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
fxthomas: Not really, you're right.

airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
```diff
@@ -832,6 +835,28 @@
...
 
+    function showMediaSessionMetadata(song) {
+        if ('mediaSession' in top.playQueue.navigator) {
+            var metadata = new MediaMetadata({
+                title: song.title,
+                artist: song.artist,
+                album: song.album,
+                artwork: [
+                    // The "&size=" parameter always resizes the image to a JPEG image (even if the original is PNG).
+                    { src: song.coverArtUrl + "&size=96", sizes: '96x96', type: 'image/jpeg' },
```
randomnicode, line 846: Don't remember off the top of my head, but make sure there is a covertArtUrl property of song?

fxthomas: There is actually, so it's working as intended. My OS doesn't display images in notifications though, so I can't "really" test it ;)

fxthomas:
No issues for a while now, and with the latest Firefox betas this works great. I'll merge it soon.

Issue #1470, opened 2020-01-13
Updating byte counts deletes user roles temporarily, leading to race conditions


randomnicode:
While updating the streaming engine and making it faster, I came across this issue that might have been latent due to slower engine, but is particularly troublesome once the overall servicing of the requests is improved. You start running into race conditions.

Byte counts getting updated deletes user roles temporarily because the user is updated whole sum.
Furthermore, this isn't even done in a transaction.

A new song coming in that duration throws exception because it doesn't claim to have access (because the streaming role has been deleted).
Additionally if you shut down the server in the middle of a song byte count getting updated, you unexpectedly lose user roles for that user.

A fix involves updating only user's byte count columns and not touching anything else (like roles etc). A proposed fix is implemented here:
https://github.com/airsonic-advanced/airsonic-advanced/pull/58

jvoisin:
Wow, the "delete + insert" combo is really bad.
Shouldn't we change the `writeRole` function to first add the new roles, then delete the invalid ones?

randomnicode:
Yes, though I think there are two issues at play.

I think first and foremost you should segregate the updates. byteCounts getting updated should not touch the rest of the user (including roles). These two things are completely orthogonal. Probably shouldn't even be in the same table.

As far as updating user itself is concerned, you'd still have a race condition (but on the other/negative side):

Currently:
- User 1 has roles: 1, 2, 3
- Thread 1: Update the user with roles 1, 3
- Roles get deleted
- Thread 2: Check on role 1 _fails_ because all roles were deleted, even though 1 should be there
- Thread 1: add roles 1, 3 (but check has already failed)

If we add new roles and then delete invalid ones:
- User 1 has roles: 1, 2, 3
- Thread 1: Update the user with roles 1, 3
- Roles 1, 3 get added
- Thread 2: Check on role 2 _succeeds_ because it was _not_ deleted even though it should have
- Thread 1: delete role 2 (but check has already succeeded)

You could just put it in a transaction.

tesshucom:
https://github.com/airsonic/airsonic/issues/1191#issuecomment-531542235

There seems to be a lot of bug reports regarding transaction. :pensive:

randomnicode:
The comment is no longer relevant.
The default tx isolation level for HSQLDB now is actually `READ_COMMITTED`, and not `READ_UNCOMMITTED`.

[...]

PR #1662, opened 2020-04-25
WIP: Import media scanner parallelization improvements

fxthomas:
This is a successor of #1385, which I'm reviving because I'm interested in the performance improvements.

This is still a work in progress for now, as I'm unsure about the following things:

* The `markPresent`/`deleteMediaFiles` methods take a collection as input, but the size is unbounded. I'm not sure this will not cause performance issues with some databases on larger collections (particularly HSQLDB)?
* The time window between the creation of new albums/artists/files and their persistence on the db is much larger in this PR (because it's only persisted at the end of the process). I haven't yet checked if this doesn't cause concurrency issues with other parts of Airsonic.
* Memory usage should be okay (the database would probably use that memory anyway before, especially in the case of HSQLDB), but I haven't run benchmarks.
* The fork-join pool seems to be only used for persistence (not sure yet what `parallelStream` uses, will need to document myself on that).
* The logging is a bit verbose for my tastes (will probably put some things in `DEBUG`).


randomnicode:
> The fork-join pool seems to be only used for persistence

No. The entire thing runs on the created fork-join pool. You can see the initial CompletableFuture is taking the pool as an argument. Once it's in that thread, any subsequent forks are taken up by pool threads again. Try having it spit out a thread name while running a scan. You'll see the whole operation inside the pool.

> (not sure yet what parallelStream uses, will need to document myself on that).

Without knowing which parallelStream you're talking about,
 - Java's stream api by default uses the common pool
 - If it's within a fork-join pool, it just uses the same one.

> The time window between the creation of new albums/artists/files and their persistence on the db is much larger in this PR (because it's only persisted at the end of the process). I haven't yet checked if this doesn't cause concurrency issues with other parts of Airsonic.

Can you define "creation" here? Not sure I understand. Like creation in memory? Creation of a folder?

tesshucom:
> The `markPresent`/`deleteMediaFiles` methods take a collection as input, but the size is unbounded. I'm not sure this will not cause performance issues with some databases on larger collections (particularly HSQLDB)?

We don't know without confirmation that specifies the conditions. 

Java parallelism has historically had many problems. Not available in all versions. It is recommended to add it as an option instead of replacing the basic function. It's not bad to consider introducing parallelism.

tesshucom:
> The `markPresent`/`deleteMediaFiles` methods take a collection as input, but the size is unbounded. I'm not sure this will not cause performance issues with some databases on larger collections (particularly HSQLDB)?

We don't know without confirmation that specifies the conditions. (The story begins with the question, is Airsonic the only one running on the AP server?)

Java parallelism has historically had many problems. Not available in all versions. It is recommended to add it as an option instead of replacing the basic function. It's not bad to consider introducing parallelism.

tesshucom:
> The `markPresent`/`deleteMediaFiles` methods take a collection as input, but the size is unbounded. I'm not sure this will not cause performance issues with some databases on larger collections (particularly HSQLDB)?

We don't know without confirmation that specifies the conditions. (The story begins with the question, is Airsonic the only one running on the AP server?)

Java parallelism has historically had many problems. Not available in all versions. It is recommended to add it as an option instead of replacing the basic function. It's not bad to consider introducing parallelism. However, it is not the optimal solution in all situations.

fxthomas:
> Java parallelism has historically had many problems. Not available in all versions.

Would that be mitigated if we bumped the minimum JRE version to Java 11? Do you have more information on these potential issues?

> AP server?

I don't know the abbreviation, does that mean "application server"?

> It is recommended to add it as an option instead of replacing the basic function.

Setting the new `MediaScannerParallelism` property to "1" should be enough, right?

Setting it to the number of cores could cause issues on some large multi-core systems, there should probably be a limit.

> If it's within a fork-join pool, it just uses the same one.

I looked a bit more into that, if I understand correctly `doScanLibrary` is run on the `ForkJoinPool` using `runAsync`, the current pool is somehow remembered (as a thread-local variable?) and re-used by subsequent `parallelStreams`, even if the pool is not specified explicitly anywhere, right? 

I'll probably add a few comments to make sure this is very obvious for the next readers.

fxthomas:
> Can you define "creation" here? Not sure I understand. Like creation in memory? Creation of a folder?

I'm thinking that other parts could potentially call `createOrUpdateMediaFile` (like when you're browsing an album page) and race with the scanner (possibly, e.g. creating an album twice?). I'll check each one just to be sure.

tesshucom:
createOrUpdateMediaFile is completely flawed.
In addition to creating two records, the same record may be overwritten. Therefore, depending on the combination of id3 data that coexist in the same directory, it is possible to create an album that has an unexpected combination of year and genre. This is because FileStructure generates the album record from the album's firstChild only but id3 overwrites properties continuously depending on the name.

tesshucom:
There are several ways to solve this. (That's right?)

 - 1. The data record should be registered as it is read. (a lot of albums with the same name with different properties are registered)
 - 2. The file structure creates an album from the first child record. Therefore, the same specifications should be adopted for ID3.
 - 3. Existing implementation should be consulted. Although there are very few, the ID3 album has a search implementation using a composite key of the album name and artist.(Although there is a question mark in the comment)However, in this case, after all, it approaches the idea of ​​2 to guarantee uniqueness.

etc.

It's a clear bug, so I fixed it in the clone before. However, because the specifications have not been decided, we chose 2 while considering 3. Of course this is a tentative response and a temporary response to get out of the worst. (If you care about performance, it should be registered only once when FirstChild is confirmed and Album instance is created)

To be honest, this is a topic that deviates from parallel processing. But it's probably related to parallelism. 

There is too much uncertainty.
If we want to continue, we need to clarify the issues involved and agree on a change policy for the architect. If we muff1nman now for a review, it probably won't go on.

(Not against parallelization. I think that it should not be possible now to implement because there is a gray part.)

tesshucom:
> Setting the new MediaScannerParallelism property to "1" should be enough, right?

 - 1. Keep current implementation
 - 2. Single threading
 - 3. Multi threading

Two of these can be selected and implemented. The combination of 1 and 2 and the combination of 2 and 3 are different. 2 and 3 are selectable only if they are guaranteed to be perfect. 

Objectively, Airsonic has a high bug rate. From the user's point of view, there are more new problems than the number of additional functions. I don't think it's convincing enough to say that the combination of 1 and 2 and the combination of 2 and 3 are the same.

Even after careful review...
if you anticipate that a problem will occur and announce a workaround, you should consider methods 1 and 3. Of course, in the beginning, 3 is an option.

fxthomas:
Regarding exception safety, it might be safer to do it this way?

```diff
diff --git a/airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java b/airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
index f366e990..ca7d5a97 100644
--- a/airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java
@@ -181,8 +181,10 @@ public class MediaScannerService {
         CompletableFuture.runAsync(() -> doScanLibrary(pool), pool)
                 .thenRunAsync(() -> playlistService.importPlaylists(), pool)
                 .thenRunAsync(() -> mediaFileDao.checkpoint(), pool)
-                .thenRun(() -> pool.shutdown())
-                .thenRun(() -> setScanning(false));
+                .whenComplete((r, t) -> {
+                    pool.shutdown();
+                    setScanning(false);
+                });
```

If I am not wrong the `thenRun` method isn't executed when an exception occurs, so the pool will linger (and no further scanning will be possible) if that happens.

fxthomas:
Also, the following strikes me as potentially problematic (at least it changes how the scanning behaves) on directory structures that have files from multiple albums in the same directory:

* On one hand `compute` uses album name/artist to get an album "pending scan", on the other hand `getAlbumForFile` also falls back to the path if not found. Previously only `getAlbumForFile` was used.
* The album metadata is filled with the first encountered file. Previously this was (I presume?) the filesystem order, which is usually constant if not sorted. Now this is also a function of which scanning thread goes to the file first, so order might change between scans.

```java
Album album = albums.compute(file.getAlbumName() + "|" + artist, (k,v) -> {
    Album a = v;

    if (a == null) {
        a = albumDao.getAlbumForFile(file);
    }

    if (a == null) {
        a = new Album();
        a.setPath(file.getParentPath());
        a.setName(file.getAlbumName());
        a.setArtist(artist);
        a.setCreated(file.getChanged());
    }

    firstEncounter.set(!lastScanned.equals(a.getLastScanned()));

    if (file.getDurationSeconds() != null) {
        a.incrementDurationSeconds(file.getDurationSeconds());
    }
    if (file.isAudio()) {
        a.incrementSongCount();
    }

    a.setLastScanned(lastScanned);
    a.setPresent(true);

    return a;
});
```

tesshucom:
> Should we retain the original implementation in case MediaScannerParallelism is set to 1?

Yes. Rather, it is inevitable when trying to make a safe transition.

- Convert the processing block you want to parallelize into a function and rewrite it as a simple call
- Add optional parallel execution

That's it. Each can be planned as an independent task and can even be split into releases.
Limit risky updates. Of course, the design should allow for unit testing. It is hoped that tests that intentionally throw regular exceptions, Runnables, or InterruptedExceptions in a thread will be committed at the same time(We can see how much is taken into account). Designs that are difficult to test indicate design quality issues.

> Should we limit the number of threads to account for systems with a lot of cores? To what value?

I don't think we need to think about this yet. Make it as compact as possible.

> Regarding exception safety, it might be safer to do it this way?

Thread exceptions are best proven with tests. A profile is required at the final stage after checking each unit. Personally, Executor Service is more reliable. But I respect the desire to try new methods. It is OK if it is accompanied by appropriate verification. Lambda exception handling or async has been done by Spring's http client for quite some time. If you haven't seen it, it may be helpful.

tesshucom:
> Should we retain the original implementation in case MediaScannerParallelism is set to 1?

Yes. Rather, it is inevitable when trying to make a safe transition.

- Convert the processing block you want to parallelize into a function and rewrite it as a simple call
- Add optional parallel execution

That's it. Each can be planned as an independent task and can even be split into releases.
Limit risky updates. Of course, the design should allow for unit testing. It is hoped that tests that intentionally throw normal-exceptions, Runnables, or InterruptedExceptions in a thread will be committed at the same time(We can see how much is taken into account). Designs that are difficult to test indicate design quality issues. 

> Should we limit the number of threads to account for systems with a lot of cores? To what value?

I don't think we need to think about this yet. Make it as compact as possible.

> Regarding exception safety, it might be safer to do it this way?

Thread exceptions are best proven with tests. A profile is required at the final stage after checking each unit. Personally, ExecutorService is more reliable. (Even incorporating cancellation later is relatively easy.) But I respect the desire to try new methods. It is OK if it is accompanied by appropriate verification. Lambda exception handling or async has been done by Spring's http client for quite some time. (I don't know if it will be helpful). If you haven't seen it, the shortcut is to learn from existing processes.

___

#1708
#1470

Do you think this issue will be a blocker?
I'm not going to be mean ...  I just want to proceed carefully :pensive:
Parallelization probably has more impact than everyone think. If you feel the volume is large, it is a good idea to divide the phases in stages.

tesshucom:
> Regarding exception safety, it might be safer to do it this way?

It can be proved by a test. It should be test driven. Throw regular exceptions, runtimes, InterruptedException in subthreads, etc. If it is difficult to test, it indicates that there is a problem with the design quality. Finally profiling is needed.

> Should we retain the original implementation in case MediaScannerParallelism is set to 1?

That is the case if you are migrating at no risk. If you just want to parallelize, you don't have to stick to the past pull .

 - Make the block to be parallelized into a function and commit just to call it.
 - Incorporating parallelization flow as an option, and commit.

> Should we limit the number of threads to account for systems with a lot of cores? To what value?

Although there is a formula for the optimum value of threading, I do not think it is a very important option. Many implementations Pooling　default to the 2-5th place. As you can see on the forum, some people use the Raspberry Pi. Considering the performance of common rental VPS, recent NAS, etc., it is not a very important option.
At first you should focus on the bug-free transition.

tesshucom:
> Also, the following strikes me as potentially problematic (at least it changes how the scanning behaves) on directory structures that have files from multiple albums in the same directory:

Devil's proof was publicized to Airsonic not once or twice. If the architect has a big impact, you can ask the committer a few questions to get an idea of ​​how much consideration is being done...
In some cases it may be faster to stop or discard and redesign a new one... :pensive:

The point where fxthomas expresses concern is one of several points. It can be achieved by performing a complicated check, but it can also be achieved in a slightly cleaner way.
In general, the more complex the logic, the more difficult it is to verify.

> it is possible to present the design in advance which processing blocks are desired to be parallelized.

Threading is not limited to SQL-based design. MusicFolder/Artist/Album/Song etc. In some cases it will be easier. That's right?
(So to ask which processing blocks are going to be parallelized is valid.)


#1708 refers to a bug in MediaScannerService#updateAlbum. 
There is a line in the updateAlbum where a ``boolean firstEncounter`` is generated.
This bug can be fixed by overwriting genre, year, etc. only in the block created using firstEncounter.
When parallelizing, if you do parallelization in the scope where firstEncounter does not conflict, no problem will occur.

randomnicode:
sorry, I haven't been paying attention here for a bit. there are a lot of comments here, so I'll cursorily go over some:

> The markPresent/deleteMediaFiles methods take a collection as input, but the size is unbounded. I'm not sure this will not cause performance issues with some databases on larger collections (particularly HSQLDB)?

Do batched updates, but this is a DB issue, not a scanner issue.

> The time window between the creation of new albums/artists/files and their persistence on the db is much larger in this PR (because it's only persisted at the end of the process). I haven't yet checked if this doesn't cause concurrency issues with other parts of Airsonic.

Not sure I understand this. What concurrency issues? If it's not persisted in the DB, then the DB remains functioning with the data as it was before (as if the scan never happened).

> Memory usage should be okay (the database would probably use that memory anyway before, especially in the case of HSQLDB), but I haven't run benchmarks.

Already did a profile, and the results should be in one of the issue threads somewhere

>  Should we limit the number of threads to account for systems with a lot of cores? To what value?

? Why? More cores is a good thing, why try to reduce your own efficiency?
You don't want to be cpu-bound. IO-bound is another issue. I can see an argument for parallel threads taxing out your IO, but how're you going to figure this out for each machine?

In any case, the parallelism is specifiable as a var

>  Is Java 8 properly supported?

Again, I don't understand this question. What do you mean is Java 8 supported? You just compiled it? I see a green checkmark indicating the CI was able to compile it and run tests on it? Of course it's supported.

> Java parallelism has historically had many problems. Not available in all versions. 

I don't know what @tesshucom is talking about here. What historical problems? From the Java side? I don't know of any.
What do versions have to do with any of this? You're using a minimum of JRE 8 because that's your target compilation set in your pom. Any previous versions don't matter?

> In the first place, I'm worried about why parallelStream comes up in server programming. Is there any reason why you can't use Executor?

Ummm... Streams exist for a reason. I see no reason why you _wouldn't_ want to use them?
If you're asking for a general reason to prefer forkjoinpools in this scenario over executor threads, then there are plenty of articles everywhere online that can illuminate this for you.
This is a recursive task, very suitable for a lightweight, task sharing divide-and-conquer kind of alg.
Feel free to implement them in Executors with heavier complete thread stacks too. Maybe you can then profile the two and compare?
My guess is that the fork join implementation will win out (for this use case).

> Would that be mitigated if we bumped the minimum JRE version to Java 11? Do you have more information on these potential issues?

This is irrelevant. The fork join pool has been around since 1.7. Streams since 1.8. I don't know about what issues are being talked about here.

> Setting it to the number of cores could cause issues on some large multi-core systems, there should probably be a limit.

Again, not sure what issues you're expecting?

> I'm thinking that other parts could potentially call createOrUpdateMediaFile (like when you're browsing an album page) and race with the scanner (possibly, e.g. creating an album twice?). I'll check each one just to be sure.

You cannot create an album twice. Your DB has a uniqueness constraint. That key you see in the `compute` (with a "|") represents that constraint.

What are the issues you're expecting even if there is a race condition? I see nothing but a normal workflow of the app.
Say you start a media scan and create an album in memory, then another part of the system creates the album while the scanner is still running. Okay, so, what's the issue? When we decide to persist, it'll clobber the existing album. How is that different from any other album (we're clobbering all of them, or creating new ones if they don't exist in the db _at the time of persistence_)?

> createOrUpdateMediaFile is completely flawed.
In addition to creating two records, the same record may be overwritten.

It cannot create two records. There is a uniqueness constraint enforced at the DB level.
There is however a race condition that exists and is unfortunate. I'm still pondering how to solve and generate (even a pseudo)atomic upsert in a DB-agnostic manner (every DB seems to have its own way of upserting)

> If I am not wrong the thenRun method isn't executed when an exception occurs, so the pool will linger (and no further scanning will be possible) if that happens.

good point

> Also, I would rather have the pool be stored somewhere instead of just creating as a variable and forgetting it as soon as scanLibrary exits.

It _is_ stored, on the thread stack itself.

> Also, the following strikes me as potentially problematic (at least it changes how the scanning behaves) on directory structures that have files from multiple albums in the same directory:

I don't see what the actual issue is here?

> The album metadata is filled with the first encountered file. Previously this was (I presume?) the filesystem order, which is usually constant if not sorted. Now this is also a function of which scanning thread goes to the file first, so order might change between scans.

No. Java specifications specifically point out that you're not retrieving the files in any particular order (not "file system order" or whatever). So you're _by spec_ currently receiving a file in not any guaranteed order, and it can be but certainly not guaranteed to be constant. All album population now and in this PR is basically whatever file we come across first and _by spec_ random.

> Many implementations Pooling　default to the 2-5th place

No, Java's default common forkjoin pool actually ties the default parallelism to the number of cpu cores.
Since I'm using a custom pool, I did the same.

tesshucom:
At this stage, the bottom technical details don't really matter :relieved: It is more important to shed light on larger architects, design issues and concerns that need to be resolved in advance. Inconsistent speedup has no value.

randomnicode:
> At this stage, the bottom technical details don't really matter 😌 It is more important to shed light on larger architects, design issues and concerns that need to be resolved in advance. Inconsistent speedup has no value.

Just answering the technical questions that were posed (presumably for me since I'm the original author of this PR)

fxthomas:
There are multiple propositions that I want to validate re: consistency of the processing (and note that this also applies to the current scanning algorithm, but issues may appear more often when introducing parallelism, hence why I think it is important to check everything).

The notes below are the results of what I understand of the whole problem.

----

Right now we have multiple different (and conflicting) definitions of an "album", depending on the code:

1. An album may be a `media file -> (album name, album artist)` mapping:
    * In media scanner
    * In `createOrUpdateAlbum`
    * In `createMediaFile`
2. An album may be a `media file -> (album name, file artist)` mapping:
    * In media scanner as a fallback
    * In `AlbumDao.getAlbumForFile`
3. An album may be a `media file -> (album name, parent path)` mapping:
    * In `AlbumDao.getAlbumForFile` as a fallback

Issues will happen if mapping `album -> media files` has:

1. Multiple `album artist` values:
    * These should correspond to multiple albums in the database.
2. Multiple `artist` values:
     * `AlbumDao.getAlbumForFile` should use the same `album artist OR artist` logic?
     * If they all have the same `album artist`, this should not be an issue.
     * If they don't have the same `album artist`, see (1), they should be multiple albums.
3. Multiple `parent path` values:
     * Same album because of directory organizations like `album / disc 1 / file.mp3`.

In the case where a file has neither `album artist` nor `artist` metadata, things get a little trickier, and I suspect that this is where the `parent path` mapping in `getAlbumForFile` comes from. I think these should probably be ignored in a first scanning pass and merged into existing albums on a case-by-case basis, following this logic:

1. The directory doesn't have an "album" (all files in the directory have no metadata) : create a single album for the directory, with no artist.
2. The directory has exactly one (1) "album" (the files without metadata are corrupted/extra media/etc) : add them to the album.
3. The directory has more than one "albums" (it's a compilation of multiple sources) : create one album for all files.

----

What's displayed in the UI, on the other hand, are "media files" (of the "folder types"), which are often misconstrued as "albums" (because in the ideal case album <-> folder media files is bijective).

The main cases for this are:
* A "folder media file" has exactly one "album", we use it.
* A "folder media file" has more than one "album": which one do we display? Pick the first one? Using what order?

It actually seems to be a bit more complex than this. Because there is no clear relationship between "albums" and "folder media files", the `MediaFileService.createMediaFile` method will populate the parent media file by... listing children (in whatever order) and picking the first one it finds. It then re-parses the file metadata to populate the "folder media file" metadata, using the `media file -> (album name, album artist)` mapping.

In principle this should not be different between the parallel and non-parallel versions.

I think that's it for now. I'll investigate how these are implemented in Airsonic and propose a list of changes later!

randomnicode:
I'm trying to understand what is being said here (the whole bit about albums and what not) and how it pertains to the process of parallelization that is being done.

Whatever issues with definitions of albums and the semantics of what it means and how they should be changed isn't pertinent to the process of parallelization (which is the thesis and domain of this PR)

You simply have files coming in, either in series or in parallel, and you need to determine and update the album record for it. I don't know what album -> file mappings etc there are issues with.

So with the hope that this clears confusion, here is your logic flow for your scanners:
Currently:
1. A file comes in (picked at random from a directory)
2. If its got no album or artist or path (or its not audio), we return without any change
3. Otherwise we try and retrieve an album it matches from the DB (so we're updating the correct record)
    - The album name needs to match
    - Then either the artist name needs to match, or the path needs to match as the parent directory of the file.
4. If there are no matches, then that album doesn't exist in the DB, and we create a new one and fill in some cursory details from the media file.
5. We fill in a bunch of additional stuff from the file itself. Yes, each successive file from the same album will clobber these details filled in from previous one, but it doesn't really matter because the files are picked at "random". The album will have the details of the "last file" it encounters (which is random)
6. Then we persist the album in the db, so the next file coming in with the same album can retrieve it (step 3). We also index it (if it was the first file that had this album)

When you parallelize it:
1. A file comes in (picked at random from a directory), in parallel with other files
2. If its got no album or artist or path (or its not audio), we return without any change
3. Otherwise we try and retrieve an album it matches from the map of potential persistence records (so we're updating the correct record)
  - The album name needs to match
  - The artist name needs to match
  - Both of these are guaranteed to be non null
  - We apply this layer because it saves us having to save the albums into the db after every file coming in (Step 6 above)
  - At this point, if a parallel file has the same album and artist, you're guaranteed to progress atomically
4. If you find an album from this map, then some other file has retrieved it from the db before or created a new album (and updated with its details in both scenarios)
5. If you don't find it in the map, this is the first file to have that album. You need to search the DB for possible matches and if you don't find anything, create one (both similar to Step 3/4 of the previous method)
6. We fill in a bunch of additional stuff from the file itself. Yes, again each successive file from the same album will clobber these details filled in from previous one, but it doesn't really matter because the files are picked at "random". The album will have the details of the "last file" it encounters (which is random). It is all done atomically.
7. We do not persist. The map is enough. The next file comes in, in parallel and takes advantage of the entry in the map. Persistence happens at the end once all files are done.


Now what is the issue and tell me about an exact edge case that has everyone worried? Show me a precise example that is processed one way by one system and another way by a different system?

fxthomas:
> Now what is the issue and tell me about an exact edge case that has everyone worried?

I already noted some inconsistencies in the way media files are scanned, namely as to whether we choose name/artist, name/album artist or name/path to match albums.

In practice (at least on Linux, not sure on Windows) I've seen that a lot of filesystems always return files in the same order, which is then picked up by the Java API. Existing issues might not be very visible with the current single-threaded implementation, but with a parallel implementation however, you add an additional factor of which thread comes up first, and that might be much less consistent.

I will feel much better having them fixed, so I will probably create a couple of smaller PRs with unit tests before I'm sure this one doesn't break anything for those edge cases.

tesshucom:
That's a certainty approach.

In addition, it's better to redo the requirement definition. 

 - What we have to do is not to aim for a benchmark maximum.
 - What we must avoid is a server outage as reported at reddit. It's the worst of all incidents.

Inevitably, you need to determine the landing points that have the least impact.
That is why I insist on preserving the existing flow. It is also overkill to use more threads than the default. (At least at the first landing point) Minimize waste and minimize verification cases. Give some thought to why Java has an option to throttle the number of threads. This is to prevent ridiculous runaway or resource occupation of applications that are only part of the server service configuration.

I don't like to decide important specifications in a closed room. If you want to drive it, it's best to create a dedicated roadmap for issues and get agreement from muff1nman and other members on the requirements. Implementation is then.

After forking from Subsonic, you may recognize that it is the most risky feature addition. I felt that there are limits to the current bottom-up development.

randomnicode:
> I already noted some inconsistencies in the way media files are scanned, namely as to whether we choose name/artist, name/album artist or name/path to match albums.

This isn't an issue with _this_ PR. This inconsistency exists for the existing scanner as well as the parallel scanner. This PR does nothing to _introduce_ it or _fix_ it.

> In practice (at least on Linux, not sure on Windows) I've seen that a lot of filesystems always return files in the same order, which is then picked up by the Java API.

? Am I reading this right? You're specifically _choosing_ to ignore very specific guidance in the Java docs. Doesn't matter about in practice. You're relying on behavior that is explicitly pointed out in Javadocs and cautioned _against_ depending on specifically. You're introducing latent bugs by adhering to some notion of how it _will_ work when Java docs specifically say it _won't_. Consider the notion that the behavior, since it is told it won't behave the way you're saying, will change from JVM to JVM, or from version to version, or from system to system. You're thinking you're being consistent, but you're actually just putting in bugs.

Here's the ACTUAL wording from the Java docs, and they go out of their way to point this out:

> There is no guarantee that the name strings in the resulting array will appear in any specific order; they are not, in particular, guaranteed to appear in alphabetical order.

Please read and comprehend and understand why specifications (and documentation for it!) exist. Whatever behavior you're relying on that you think is good _for the sake of consistency_ or _observed in practice_ doesn't matter because the Java docs explicitly say otherwise and caution against depending on that behavior.

> Existing issues might not be very visible with the current single-threaded implementation, but with a parallel implementation however, you add an additional factor of which thread comes up first, and that might be much less consistent.

I'm still waiting for the issues theoretical or otherwise that exist on one platform and not the other. What issues are showing up with parallel scanning that won't show up for the older implementation? Have you encountered any? Has anyone?

> I will feel much better having them fixed, so I will probably create a couple of smaller PRs with unit tests before I'm sure this one doesn't break anything for those edge cases.

Fair enough. Good luck!

randomnicode:
> What we have to do is not to aim for a benchmark maximum.
What we must avoid is a server outage as reported at reddit. It's the worst of all incidents.

What maximum? What outrage?
This feature's been in use for more than half a year now publicly. I still haven't seen a bug report.

> Give some thought to why Java has an option to throttle the number of threads. This is to prevent ridiculous runaway or resource occupation of applications that are only part of the server service configuration.

Ummm, Java, by default, configures your fork join pool to be maximally related to the number of your (logical) CPU cores. You know that, right? It's trying to _MAXimize_ your parallel thread count, not _MINimize_ it.

> I don't like to decide important specifications in a closed room. 

This isn't a closed room. This is a public PR.

> After forking from Subsonic, you may recognize that it is the most risky feature addition. I felt that there are limits to the current bottom-up development.

_This_ is the most risky feature addition since the fork? Or are you talking in general?
If _this_ is it, then you folks should really take out the _"Be Bold!"_ part from your contribution docs?

tesshucom:
Is it supposed that only Airsonic is running on the physical server? I think there is a lack of consideration.
It means that the claimed specifications are not in the stage of taking initiative.

randomnicode:
> Is it supposed that only Airsonic is running on the physical server? I think there is a lack of consideration.

And that matters because? I'm telling you what Java runtime defaults are. Specified by the JRE implementation itself. My opinion has nothing to do with it. Would you consider them inconsiderate because they maximize use all the cores for each JVM?

These silly arguments are basically the counterpoint of what any decent modern development practices preach. Stop micro-optimizing. You're trying to make an issue when there isn't one.

Once you actually hit an issue, then have a conversation. Actually, better yet, you don't even need to because, oh look, there's even a knob for that! (you can specify the parallelism too!)

tesshucom:
Many of the topics you are using more characters than test code are topics for tuneup. Fix legacy and test it to make it work correctly. Do minimal parallelization with minimal implementation code. Fine adjustments for tuneup are possible later. Refactoring does not require much advanced technology if the correct behavior is guaranteed by testing. Just make a small code fix and compare the behavior. Just repeat that task. 
If anyone argues that such a flow would take too long, I would even doubt the ability to calculate the originally needed estimate. 

Some of the flaws in legacy aren't edge cases. It's a clear flaw from an engineer's point of view, and especially users who want to manage songs in a tidy manner are very sensitive to this kind of consistency. It's nonsense to overlook this and discuss parallelization. And it's unreasonable to break or modify the spec in the middle of parallelization. It's something an engineer who can't control the specifications does.

Many projects have failed to ensure data integrity. The Airsonic isn't perfect, but it's relatively good. Relatively. If we make it more robust, we can also consider adding new features. If you misunderstand your priorities, you will spend more time.

PR #1663, opened 2020-04-25
fix Jukebox API test under windows

Faeb35:
I would like to start contributing to Airsonic and Airsonic-Advanced. This is my first PR (I'm pretty new to Git as well). While setup IntelliJ and building Airsonic, I bumped in the issue, that the Jukebox API test failed under Windows because of the different used slashes ("\\" vs. "/"). 
Solution is stolen from airsonic-advanced.


randomnicode:
Usually it's helpful (and nice) to give reference to the commit or a PR that actually committed the change in the original repo. It is tremendously helpful for cross-referencing. If you Annotate or use git blame, it'll tell you what commit this originated in. In this case it's https://github.com/airsonic-advanced/airsonic-advanced/commit/d1c67b4

Welcome aboard! Good luck!

PR #1686, opened 2020-05-02
Add some null-checks on calls to mediaFileService.getMediaFile()

Yetangitu:
Missing null-checks on mediaFileService.getMediaFile() cause NPEs in the Subsonic API and elsewhere, add some checks to avoid these.

Especially the NPE in the Subsonic API is annoying since it leads to a loss of functionality, e.g. empty podcast listings in DSub etc.

```
ERROR --- o.a.p.s.LoggingExceptionResolver: 192.168.1.136:
An exception occurred while loading https://music.example.org/rest/getPodcasts.view?v=1.2.0&c=DSub&id=11&t=<hidden>&s=<hidden>&u=<hidden>
#011at org.airsonic.player.service.MediaFileService.getParentOf(MediaFileService.java:156) ~[classes!/:11.0.0-SNAPSHOT]
#011at org.airsonic.player.controller.SubsonicRESTController.createJaxbChild(SubsonicRESTController.java:1243) ~[classes!/:11.0.0-SNAPSHOT]
#011at org.airsonic.player.controller.SubsonicRESTController.createJaxbPodcastEpisode(SubsonicRESTController.java:1606) ~[classes!/:11.0.0-SNAPSHOT]
#011at org.airsonic.player.controller.SubsonicRESTController.getPodcasts(SubsonicRESTController.java:1572) ~[classes!/:11.0.0-SNAPSHOT]
...
```

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/SubsonicRESTController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/SubsonicRESTController.java
@@ -1716,6 +1716,10 @@ public class SubsonicRESTController {
 
         Bookmarks result = new Bookmarks();
         for (Bookmark bookmark : bookmarkService.getBookmarks(username)) {
+            MediaFile mediaFile = mediaFileService.getMediaFile(bookmark.getMediaFileId());
+            if (mediaFile == null) {
+                continue;
+            } 
             org.subsonic.restapi.Bookmark b = new org.subsonic.restapi.Bookmark();
             result.getBookmark().add(b);
             b.setPosition(bookmark.getPositionMillis());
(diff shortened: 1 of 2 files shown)


airsonic-main/src/main/java/org/airsonic/player/controller/SubsonicRESTController.java
```diff
@@ -1723,7 +1725,9 @@ public void getBookmarks(HttpServletRequest request, HttpServletResponse respons
             b.setChanged(jaxbWriter.convertDate(bookmark.getChanged()));
 
             MediaFile mediaFile = mediaFileService.getMediaFile(bookmark.getMediaFileId());
-            b.setEntry(createJaxbChild(player, mediaFile, username));
+            if (mediaFile != null) {
+                b.setEntry(createJaxbChild(player, mediaFile, username));
```
fxthomas, line 1729: I assume that `entry` is going to be null in this case, wouldn't it be better to skip it entirely with `continue`?

Yetangitu: I was of two minds here, thinking that empty bookmarks might give a warning of something being amiss in the database without crashing the system. Then again, people probably don't want to see empty bookmarks... therefore I moved the test up to avoid generating empty entries.

PR #1687, opened 2020-05-03
use String.format("%tF") instead of "%tY-%tm-%td" 

Yetangitu:
As to why java suddenly started complaining about `java.util.MissingFormatArgumentException: Format specifier '%tm'` errors I have no idea but the fact is that it did as of the 1st of may 2020:
```
Failed to download Podcast from https://example.com/1.mp3
java.util.MissingFormatArgumentException: Format specifier '%tm'
```

Given that the intention of this format string is to produce an ISO date (_YYYY-MM-DD_) and given that `String.format` has a special formatter for this which does _not_ produce this error it seems logical to use it.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/service/PodcastService.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/service/PodcastService.java
@@ -657,7 +657,7 @@ public class PodcastService {
 
         File channelDir = getChannelDirectory(channel);
 
-        String episodeDate = String.format("%tY-%tm-%td", episode.getPublishDate());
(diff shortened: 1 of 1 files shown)


fxthomas:
I think we both stumbled on the same thing, cf #1682 :wink:

fxthomas:
Your version sounds better though (didn't know that a specific ISO formatter existed)!

Yetangitu:
> Your version sounds better though (didn't know that a specific ISO formatter existed)!

Neither did I, until I started digging through the (java)docs.

Issue #711, opened 2018-03-05
Cover art files should use larger files first


eharris:
In my music library the cover art is stored in different filenames based on the name of the album.  So a standard name like cover.jpg in the "Cover art files" portion of the config usually doesn't get them, and instead the fallback to just jpg is needed.

Unfortunately I usually also have several sizes of the cover art, large, medium, small, and thumb.  It seems as though airsonic just uses whatever jpg it finds first, which sometimes happens to be the undesirable thumbnail or other smaller sizes, when it would be a much better heuristic to use the largest jpg.

One potential workaround that might help some of the time would be if there was a way to wildcard the filenames in the config, so I could specify something like "*_large.jpg" earlier in the list than the catch-all jpg match.  Is there a way to do that?

jooola:
You could maybe try to exclude using the Exclude pattern function : 
![image](https://user-images.githubusercontent.com/19195485/36978270-34c4217a-2084-11e8-9c87-cd3fac18fd23.png)
And I don't know if the cover art files options can accept wildcard `*` but did you tried to only add `_large.jpg`
![image](https://user-images.githubusercontent.com/19195485/36978328-701ba162-2084-11e8-9ad7-5ed300524d2c.png)

eharris:
I don't really want to exclude anything, because if there's nothing else better, I would still want to use the smaller size images.

I haven't tried just adding the ending suffix, I will try that out.

jooola:
Yes also maybe they are a precedence in the Cover Art files list. Unsure about that.

eharris:
It either looks like there is no precedence in the ordering of the files it selects for the cover art, or there is no pattern/stub matching.  I just wiped out my database to get it to re-detect cover art, and it is often picking any old .jpg in the directory rather than the files I have specified (assuming/hoping that there was precedence in the album art detection).

My config specifies the following for Cover Art Files:
```Front_Large Front_Small Back_Large Back_Small jpg gif png```

I want it to use the ```Front_Large.(gif|jpg|png)``` first, followed by the others in that order.

Unfortunately it is very difficult to try to test different configurations and what airsonic uses for the album art since there seems to be no reasonably reliable way to force airsonic to update the cover art other than wiping the DB (see issue #649).

Faeb35:
I'm also struggeling with this...
I have a lot of albums, with multiple images in the folders. As example I have the following images:
dr. dre & snoop dogg - the wash (soundtrack) - back.jpg
[...]

Issue #649, opened 2017-12-12
Refreshing Artwork & Thumbnails


h45rd:
Is there any way to force airsonic to search the folder structure for new artwork (in sequence of the _cover art files_-mask) if the currently used artwork has become unavailable?

Practical example: I had some random images in various folders that were picked as artwork instead of the actual artwork. 
Deleting these images and scanning & cleaning up the DB does not refresh the associated artwork or thumbnails.
Manually deleting the thumbnails and re-scanning / cleaning does not help either.

It seems that once airsonic has associated an image with an album, it does not search the folder structure for new images, even if the associated one becomes unavailable.
Tailing the log seems to confirm this behavior:

```
2017-12-12 07:55:18.004  WARN --- o.a.p.c.CoverArtController               : Failed to process cover art /music/Some Artist/(1984) Some Album/WronglyPickedImage.jpg: java.io.FileNotFoundException: /music/Some Artist/(1984) Some Album/WronglyPickedImage.jpg (No such file or directory)

2017-12-12 07:55:18.011  WARN --- o.a.p.c.CoverArtController               : Failed to create thumbnail for Media file 47217 - Some Album
java.lang.NullPointerException: null
```

Trying to debug this issue only lead me to [another person with the same issue](https://www.reddit.com/r/airsonic/comments/7ah114/album_art_not_updating).

So I guess right now the only option would be to delete the whole database and re-scan?

eharris:
I'm having this same issue.  I've tried forcing a rescan using "Scan media folders now", as well as deleting and re-adding the affected media folder.  Neither of these fixed the bad artwork that it had saved initially.

I haven't tried to delete and recreate the whole database as that would take forever since my media is on a network share.  And waiting hours for the database to rebuild just to fix wrong image art is not a reasonable solution anyway.

I am using 10.1.1.

eharris:
Update: I have now tried deleting a media folder that had the image problems, then doing a "Clean-up database" and then re-adding and then re-scanning the media folder, and this also will not get it to pick up the different artwork.

The only thing that worked was doing the same process as above, but also renaming that folder on disk so that it appeared to no longer be there, and then it seemed to finally pick up the different artwork.

[...]

PR #1706, opened 2020-05-10
Increased cover art quality

### LINKED ISSUE #1702: Cover art has significant compression artifacting
## Problem description

Wherever scaled down album art is shown, it has noticeable compression artifacting. This is especially visible in the home gallery view with the 160px images, but also when looking at an album at 300px 

### Steps to reproduce

Open anything that displays album art at below full size

## System information

 * **Airsonic version**: 10.6.2-RELEASE
 * **Operating system**: Ubuntu 16.04.6
 * **Java version**: 1.8.0
 * **Proxy server**: None
 * **Client**: Firefox/Chrome
 * **Language**: English

## Additional notes

Looking at getCachedImage in [CoverArtController.java](https://github.com/airsonic/airsonic/blob/165701bcdc85685cf07684946efc818ee73b2d57/airsonic-main/src/main/java/org/airsonic/player/controller/CoverArtController.java#L256), it seems like this is because ImageIO.write is used to save art, and that doesn't let you pick a quality.

Seems like a quick fix, and I should be able to submit a PR in a couple days

sam-masaki:
I replaced the ImageIO.write with an ImageWriter and set the jpeg quality to 95%. PNG compression isn't affected

fixes #1702

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/CoverArtController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/CoverArtController.java
@@ -41,7 +41,12 @@ import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.servlet.mvc.LastModified;
 
 import javax.annotation.PostConstruct;
+import javax.imageio.IIOImage;
 import javax.imageio.ImageIO;
+import javax.imageio.ImageWriteParam;
+import javax.imageio.ImageWriter;
+import javax.imageio.plugins.jpeg.JPEGImageWriteParam;
+import javax.imageio.stream.FileImageOutputStream;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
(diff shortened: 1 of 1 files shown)


tesshucom:
@eharris

There are various problems related to cover art, but ...

#649
#711

This PR doesn't seem to conflict with existing troubles, what do you think?

airsonic-main/src/main/java/org/airsonic/player/controller/CoverArtController.java
```diff
@@ -236,7 +241,16 @@ private void sendUnscaled(CoverArtRequest coverArtRequest, HttpServletResponse r
...
     private File getCachedImage(CoverArtRequest request, int size) throws IOException {
         String hash = DigestUtils.md5Hex(request.getKey());
-        String encoding = request.getCoverArt() != null ? "jpeg" : "png";
+        String encoding = null;
+        ImageWriteParam params = null;
+        if (request.getCoverArt() != null) {
+            encoding = "jpeg";
+            params = new JPEGImageWriteParam(null);
+            params.setCompressionMode(ImageWriteParam.MODE_EXPLICIT);
+            params.setCompressionQuality(0.95f);
```
eharris, line 250: I would think quality of 90 should be more than sufficient for pretty much everything.  We probably don't want to unnecessarily blow up the image size too dramatically.

eharris (changes requested):

Issue #1708, opened 2020-05-12
Data inconsistency may occur in albums with ID3 tags


tesshucom:
## Problem description

Data inconsistency occurs in certain data patterns.
The decision spec of record is ambiguous when albums with different properties exist.

- File structure is determined from FirstChild in the same directory
- The ID3 tag will be added, or overwritten if it has the same name.

The problem is the case of registering an album with the same name in the ID3 tag.

### Steps to reproduce

The details are given by the test.

## System information

 * **Airsonic version**: master and all past revisions
 * **Operating system**: all 
 * **Java version**: all 
 * **Proxy server**: all 
 * **Client**: all 
 * **Language**: all 

## Additional notes

Currently, the range of defects that this causes is small.
(Because there are few places where ID3 tag is used)
Clients using ID3 tag may be affected

tesshucom:
Will be resolved in #1709.

PR #1709, opened 2020-05-12
Add AlbumDaoTest

tesshucom:
#1708 related test cases.
This in itself does not affect the operation of the server.
As a prerequisite for verifying legacy behavior, you must first merge it into master before fixing any bugs.

The following tags are registered in the attached data.

directory | filename | artist | albumArtist | album | genre | year
-- | -- | -- | -- | -- | -- | --
Scan/ID3Album/sameName/... | fileA1.mp3 | artistA | albumArtistA | albumA | genreA | 2001
Scan/ID3Album/sameName/... | fileA2.mp3 | artistB | albumArtistB | albumA | genreB | 2002
Scan/ID3Album/noTagFirst/... | fileB.mp3 | <null> | <null> | <null> | <null> | <null>
Scan/ID3Album/noTagFirst/... | fileC.mp3 | artistC | albumArtistC | albumC | genreC | 2003
Scan/ID3Album/nameInReverse/... | fileD.mp3 | artistE | albumArtistE | albumE | genreE | 2004
Scan/ID3Album/nameInReverse/... | fileE.mp3 | artistD | albumArtistD | albumD | genreD | 2005


airsonic-main/src/test/java/org/airsonic/player/dao/AlbumDaoTest.java
```diff
@@ -0,0 +1,155 @@
...
+        List<Album> albumId3s = albumDao.getAlphabeticalAlbums(offset, count, false, false, sameName);
+        assertEquals(1, albumId3s.size());
+        Album albumId3 = albumId3s.get(0);
+
+        // [albumA albumArtistA genreB 2002]
+        // This combination is "a combination that should not exist.
+        assertEquals("albumA", albumId3.getName());
+        assertEquals("albumArtistA", albumId3.getArtist());
+        assertEquals("genreB", albumId3.getGenre()); // XXX Unintentional overwrite
+        assertEquals(Integer.valueOf(2002), albumId3.getYear()); // XXX Unintentional overwrite
```
tesshucom, line 80: The following records are generated in the album table.

``albumA/albumArtistA/genreB/2002``

Real files do not have this combination.

tesshucom:
Add a setting to suppress this test itself.

```
[ERROR] Failures: 
[ERROR]   AlbumDaoTest.testNameInReverse:129 expected:<albumArtist[E]> but was:<albumArtist[D]>
```
AlbumDaoTest.testNameInReverse returns different results, at least on Travis and Windows.

Locally, you can run them individually with the command:
`` mvn clean test -DfailIfNoTests=false  -Dtest=org/airsonic/player/dao/AlbumDaoTest.java ``

airsonic-main/src/test/java/org/airsonic/player/dao/AlbumDaoTest.java
fxthomas: Does this really happen with the current scanner? That's bad...

airsonic-main/src/test/java/org/airsonic/player/dao/AlbumDaoTest.java
```diff
@@ -0,0 +1,155 @@
...
+        assertNull(album.getMusicBrainzRecordingId());
+    }
+
+    @Test
+    public void testNoTagFirst() {
+
+        // File structure
+        List<MediaFile> albums = mediaFileDao.getAlphabeticalAlbums(offset, count, false, noTagFirst);
+        assertEquals(1, albums.size());
+        MediaFile album = albums.get(0);
```
fxthomas, line 92: Instead of using `get(0)`, how about selecting files by path? The order is not important here (and might be different in different platforms).

tesshucom: Uh.
I can't immediately think of the cause of this. As you say, let's rewrite because the latter case of assertNull is the subject.

airsonic-main/src/test/java/org/airsonic/player/dao/AlbumDaoTest.java
tesshucom: The fix method that has relatively little effect is as follows :

The following processing ....

https://github.com/airsonic/airsonic/blob/f930a8e06bd5d87f66f5713620f347576499588a/airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java#L292-L300

Put in this block.

https://github.com/airsonic/airsonic/blob/f930a8e06bd5d87f66f5713620f347576499588a/airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java#L306-L313

 - However, including other properties, you need to think carefully and add tests.
 - It's hard to tell if this is a simple bug fix. Of course, data inconsistencies can simply be eliminated. However, there are doubts as to whether this is the expected data specification, and honestly it is a blank page.

Currently, the number of processes that depend on ID3 is extremely small, but it is not without it.

https://github.com/airsonic/airsonic/blob/f930a8e06bd5d87f66f5713620f347576499588a/airsonic-main/src/main/java/org/airsonic/player/service/upnp/AlbumUpnpProcessor.java#L81-L94

``// TODO: correct artist?``

A little unclear.
Note that you should be aware of these effects if you change the Album data specification rather than eliminating inconsistencies.
Sonos doesn't take a good look.

airsonic-main/src/test/java/org/airsonic/player/dao/AlbumDaoTest.java
tesshucom: I have tracked down this issue and the cause is due to an HSQLDB update. I created this test before updating HSQLDB. Results will change before and after updating HSQLDB.

> and might be different in different platforms

With POSIX, at least the range of the alphabet seems to be the same ... I wonder a little :confused:
Methods such as getAlphabeticalAlbums and getChildrenOf that implicitly follow the record registration order may return results in a different order than before. Of course, I believe that all business logic is properly sorted after that. If not, it will have an impact.

As you pointed out, there is no direct relationship between this test case and its problem.

tesshucom: A little discussion in me ...

It has already been confirmed that some methods that do not guarantee the order of the results will change the order by the update of HSQLDB. 
But is this case valid ...

tesshucom:
> IMO it should really be part of the scanner tests.

@fxthomas

Shall we build a new test directory?
``dao/mediascanner``
(If there is a directory name you like, I follow it)

 - Tests that involve scans tend to be specially configured and are often difficult to integrate with existing test classes. Also, if you assume a specific directory or data, it is more appropriate to create it individually. Merging is possible if it does not interfere with the verification, but it is difficult to predict in advance. In the first place, no one could do that.
 - As you can see from Search-test, there are often cases where an independent case is created that does not correspond to a particular class. It is easier to store in a directory that is separate from the existing basic unit tests.
 - Scan testing is of high importance, and it is advisable to add additional tests to fix these defects. In other words, it will increase from now on.
 - We can suppress test by directories. If the scan speed is increased, the suppression may be removed, but the current situation has a negative aspect. Since it is sufficient to execute it from maven in directory units, there is no problem with confirmation itself even if it is started manually.

If everything is okay, we will start such management from this commit.

PR #1714, opened 2020-05-14
Save "add to playlist"-dialog size

Faeb35:
This pull request is a copy of https://github.com/airsonic-advanced/airsonic-advanced/pull/303

This pull request adds the ability to save the size of the dialog, showing when adding tracks to a playlist. I found it very annoying to change the size of this dialog every time.

The size is saved in cookies. This makes it very lightweight and also adds the possibility to have different sizes on different computers/browsers for the same user/player.

The size of the dialog, shown when adding from the album view, is stored in a different cookie then the dialog shown, when adding from the player queue. Even though the initial size is 300 width and 250 height for both (height was 350 for mediaMain and 220 for playQueue).

### DIFF
--- a/airsonic-main/src/main/webapp/WEB-INF/jsp/albumMain.jsp
+++ b/airsonic-main/src/main/webapp/WEB-INF/jsp/albumMain.jsp
@@ -27,12 +27,15 @@
             minBorder: 30
         });
 
-        $("#dialog-select-playlist").dialog({resizable: true, height: 350, autoOpen: false,
+        var dialogSize = getJQueryUiDialogPlaylistSize("albumMain");
+        $("#dialog-select-playlist").dialog({resizable: true, height: dialogSize.height, width: dialogSize.width, autoOpen: false,
             buttons: {
                 "<fmt:message key="common.cancel"/>": function() {
                     $(this).dialog("close");
                 }
-            }});
+            },
+            resizeStop: function (event, ui) { setJQueryUiDialogPlaylistSize("albumMain", ui.size) }
+        });
 
         <c:if test="${model.showArtistInfo}">
         loadArtistInfo();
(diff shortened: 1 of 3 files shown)


fxthomas:
Thanks, I'll give it a quick test on my instance and it should good to go.

As an aside (does not prevent merging) : the "shuffle radio" feature uses LocalStorage to persist settings instead of cookies, these two should probably use the same mechanism.

Faeb35:
@fxthomas Thanks for testing and your feedback!
About `ReferenceError: getJQueryUiDialogPlaylistSize is not defined`, can you try to hit the "compile" in Maven? I had this issue once as well and disapeared after compiling. I think a jsp page was not re-compiled.
Anyway I'll try to take a look at the LocalStorage. I think it would make sense to use the same technology.

fxthomas:
> Can you try to hit the "compile" in Maven?

I cleaned the compilation directory (packaged with `mvn clean package`) before installing it, so the JSPs were all recompiled.

> Anyway I'll try to take a look at the LocalStorage. I think it would make sense to use the same technology.

I personally don't care much which we choose, and it's not that much a big deal. If anything cookies are likely to be more widely supported?

Faeb35:
I just updated the utils.js to user LocalStorage, was actually even easier then cookies ;-)
I prefere it, as Airsonic already use LocalStorage on other places and then we don't need the extra cookie handling introduced (initially) with this pull request.

About the issue you're getting: I'm not able to reproduce this. I'm guessing, that your browser could have been caching the utils.js (still uses the old version). Could you try to use a the browser in private mode, press CRTL+F5 for refresh or enable the "Prevent caching" checkbox in the browser tools under network.?

fxthomas:
Yeah, probably a cache issue. No real proof, but I think frames often prevent the browser from correctly managing the cache when hitting Ctrl-Shift-R to reload the page (should get rid of them at some point...).

Issue #1209, opened 2019-08-13
[10.4.0] web radio not handling binary stream anymore


Daryes:
## Problem description

With 10.4.0, webradio using a direct link to a mp3 or any equivalent binary stream doesn't work anymore.
Also, this error is recorded in the log file :
```
2019-08-13 19:10:24.020 ERROR --- o.a.p.s.InternetRadioService             : Failed to retrieve sources for internet radio http://direct.franceinfo.fr/live/franceinfo-midfi.mp3.

org.airsonic.player.service.InternetRadioService$PlaylistFormatUnsupported: Unsupported playlist format http://direct.franceinfo.fr/live/franceinfo-midfi.mp3
        at org.airsonic.player.service.InternetRadioService.retrievePlaylist(InternetRadioService.java:243) ~[classes!/:10.4.0-RELEASE]
        at org.airsonic.player.service.InternetRadioService.retrieveInternetRadioSources(InternetRadioService.java:157) ~[classes!/:10.4.0-RELEASE]
        at org.airsonic.player.service.InternetRadioService.retrieveInternetRadioSources(InternetRadioService.java:135) ~[classes!/:10.4.0-RELEASE]
        at org.airsonic.player.service.InternetRadioService.getInternetRadioSources(InternetRadioService.java:110) ~[classes!/:10.4.0-RELEASE]
        at org.airsonic.player.ajax.PlayQueueService.convertInternetRadio(PlayQueueService.java:746) [classes!/:10.4.0-RELEASE]
        at org.airsonic.player.ajax.PlayQueueService.convert(PlayQueueService.java:685) [classes!/:10.4.0-RELEASE]
        at org.airsonic.player.ajax.PlayQueueService.convert(PlayQueueService.java:674) [classes!/:10.4.0-RELEASE]
        at org.airsonic.player.ajax.PlayQueueService.doPlayInternetRadio(PlayQueueService.java:459) [classes!/:10.4.0-RELEASE]
        at org.airsonic.player.ajax.PlayQueueService.playInternetRadio(PlayQueueService.java:255) [classes!/:10.4.0-RELEASE]
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:1.8.0_222]
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[na:1.8.0_222]
[...]

PR #1727, opened 2020-05-20
Update InternetRadioService to support direct binary urls

Faeb35:
This pull request is from directly ported from Airsonic-advanced:
https://github.com/airsonic-advanced/airsonic-advanced/pull/309 and https://github.com/airsonic-advanced/airsonic-advanced/pull/313.

Thanks a lot to @randomnicode for his support with this!

Fixes this issue airsonic/airsonic#1209
In the current release the specified radio url in the settings should point to a m3u-file containing urls to radio streams.
With this update you can now also directly specify the url to a binary radio stream.


Faeb35:
@fxthomas I just pushed the test as well. It seems to work, but please double check if it's fine. I'm not familiary with the mock framework and testing in general.

Issue #1425, opened 2019-12-03
Cleanup database doesn't clear files after removing music folder


fxthomas:
After removing a music folder, media files are left in the database.

If the file still exists on the filesystem, its entries are not removed from the Airsonic database even if you run "clean-up database" from the settings page.

This results in multiple entries for the same file, like this:

```
$ java -jar hsqldb-2.5.0/sqltool.jar --sql 'select id, type, artist, album, title, parent_path, present from media_file order by type;' --inlineRc 'url=jdbc:hsqldb:file:.run/db/airsonic,user=sa,password='
ID  TYPE       ARTIST                 ALBUM                                       TITLE                        PARENT_PATH                                                PRESENT
--  ---------  ---------------------  ------------------------------------------  ---------------------------  ---------------------------------------------------------  -------
 1  ALBUM      水曜日のカンパネラ              SUPERMAN                                    [null]                       /home/fx/Music                                             TRUE
21  ALBUM      水曜日のカンパネラ              SUPERMAN                                    [null]                       /home/fx/Music                                             TRUE
23  ALBUM      水曜日のカンパネラ              SUPERMAN                                    [null]                       /home/fx/Music2                                            TRUE
```

tesshucom:
@fxthomas 

Checked using IndexManagerTestCase.

IndexManagerTestCase is a test case to confirm the operation of deleting the index.
Therefore, the database is not expunged.

#### Run a test case to see the data

![image](https://user-images.githubusercontent.com/27724847/82745383-80e41600-9dbe-11ea-82d3-3611267eaba6.png)

#### Run a test case to see the data

Add the following to the end of the test case.
Same as the processing performed in MusicFolderSettingsController#expunge ().
```
        artistDao.expunge();
        albumDao.expunge();
        mediaFileDao.expunge();
        mediaFileDao.checkpoint();
```

tesshucom:
Commit the fix.
You may need to be a little careful.

tesshucom:
add to.
In addition to the media_file, album, and artist tables, scrutiny is required.
Bookmarks, playlists, etc.
Simultaneous solution is desirable.

tesshucom:
add to.
In addition to the media_file, album, and artist tables, scrutiny is required.
Bookmarks, playlists, etc.
Simultaneous solution is desirable.

[...]

PR #1732, opened 2020-05-24
Fix dao#expunge()

tesshucom:
Related to #1425.
It fixes the bug that residual data occurs and accumulates.

There are three commits.
 (1) Minor modification of related test case (does not affect processing)
 (2) Test case for this defect
 (3) Modification code and difference of (2)

May require additional validation.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/dao/MediaFileDao.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/dao/MediaFileDao.java
@@ -728,17 +728,17 @@ public class MediaFileDao extends AbstractDao {
     }
 
     public List<Integer> getArtistExpungeCandidates() {
-        return queryForInts("select id from media_file where media_file.type = ? and not present",
+        return queryForInts("select id from media_file where media_file.type = ? and (not present or folder not in (select path from music_folder))",
                 MediaFile.MediaType.DIRECTORY.name());
     }
 
     public List<Integer> getAlbumExpungeCandidates() {
-        return queryForInts("select id from media_file where media_file.type = ? and not present",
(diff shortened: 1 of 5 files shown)


airsonic-main/src/main/java/org/airsonic/player/dao/MediaFileDao.java
```diff
@@ -749,7 +749,7 @@ public void expunge() {
 
         final int batchSize = 1000;
         for (int id = minId; id <= maxId; id += batchSize) {
-            update("delete from media_file where id between ? and ? and not present", id, id + batchSize);
+            update("delete from media_file where id between ? and ? and (not present or folder not in (select path from music_folder))", id, id + batchSize);
```
tesshucom, line 752: The revision policy is common to all Dao.
MusicFolder data that does not exist at the time of expunge is considered to be deleted.
(User data for which a dead record has already occurred can be rescued)

One thing to keep in mind is whether the present flag needs to be updated when the MusicFolder is deleted.

 - It has nothing to do with most existing processes, it should be....should ..
 - Before applying this code, you have to intentionally change the music folder and create a dirty record to verify. (In some cases, there may be cases in which we cannot help....?) Proof is difficult only with unit tests.

tesshucom:
There is a problem with 4thline-repo and compilation fails on travis. (There is a possibility that it can be avoided if the cache is alive) Local validation may be the focus for some time.

tesshucom:
There are many bug reports that may occur due to residual data, or bug fixes that are difficult to verify if normal cleanup is not guaranteed.
Originally intended to improve media_file, it will be extend the scope of verification to all albums, artists, and Daos that have not yet been implemented expunge.

Issue #1685, opened 2020-05-02
NullPointerException in CoverArtController


wolkenschieber:
## Problem description

Looking into tomcat's log I'm seeing following NPE being thrown multiple times:
```
Mai 02 18:46:51 edoras tomcat9[23010]: 2020-05-02 18:46:51.393  WARN --- o.a.p.c.CoverArtController               : Failed to process cover art /some/media/file.mp3: java.lang.RuntimeException: java.lang.NullPointerException
Mai 02 18:46:51 edoras tomcat9[23010]: java.lang.RuntimeException: java.lang.NullPointerException
Mai 02 18:46:51 edoras tomcat9[23010]:         at org.airsonic.player.controller.CoverArtController.getImageInputStreamWithType(CoverArtController.java:302) ~[classes/:10.6.1-RELEASE]
Mai 02 18:46:51 edoras tomcat9[23010]:         at org.airsonic.player.controller.CoverArtController.getImageInputStream(CoverArtController.java:281) ~[classes/:10.6.1-RELEASE]
Mai 02 18:46:51 edoras tomcat9[23010]:         at org.airsonic.player.controller.CoverArtController.access$600(CoverArtController.java:64) ~[classes/:10.6.1-RELEASE]
Mai 02 18:46:51 edoras tomcat9[23010]:         at org.airsonic.player.controller.CoverArtController$CoverArtRequest.createImage(CoverArtController.java:387) ~[classes/:10.6.1-RELEASE]
Mai 02 18:46:51 edoras tomcat9[23010]:         at org.airsonic.player.controller.CoverArtController$PlaylistCoverArtRequest.createImage(CoverArtController.java:533) ~[classes/:10.6.1-RELEASE]
Mai 02 18:46:51 edoras tomcat9[23010]:         at org.airsonic.player.controller.CoverArtController.getCachedImage(CoverArtController.java:251) ~[classes/:10.6.1-RELEASE]
Mai 02 18:46:51 edoras tomcat9[23010]:         at org.airsonic.player.controller.CoverArtController.handleRequest(CoverArtController.java:129) ~[classes/:10.6.1-RELEASE]
Mai 02 18:46:51 edoras tomcat9[23010]:         at jdk.internal.reflect.GeneratedMethodAccessor452.invoke(Unknown Source) ~[na:na]
[...]

PR #1737, opened 2020-05-30
Issue #1685

wolkenschieber:
PR: Fixes Issue #1685 

NPE of issuze occurs, if artwork binary data has a size of 0. 

Introduced null-check in `CoverArtController` will return a default `InputStream` in that case. The log message now is ` WARN --- o.a.p.c.CoverArtController               : Failed to process cover art /tmp/file.mp3: ImageIO.read failed`, instead of a NullPointerException stack trace.

Signed-off-by: wolkenschieber <wolkenschieber@users.noreply.github.com>

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/CoverArtController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/CoverArtController.java
@@ -295,7 +295,11 @@ public class CoverArtController implements LastModified {
             try {
                 LOG.trace("Reading artwork from file {}", mediaFile);
                 artwork = jaudiotaggerParser.getArtwork(mediaFile);
-                is = new ByteArrayInputStream(artwork.getBinaryData());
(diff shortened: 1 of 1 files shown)


wolkenschieber:
Turned out as two part PR:

1. Fix issue #1685 
2. Fix build
    1. Update dependencies to fix CVEs
    2. Update owasp for Java 8 (https://github.com/jeremylong/DependencyCheck/issues/1574)

wolkenschieber:
@tesshucom Label `type: dependency-upgrade` would also fit here.

tesshucom:
Thank you. 
Multiple dependency problems exist at the same time, and some problems are entangled. cxf ant is one of them. Java8 support may be dropped in the next release.
I think it would be better to solve them separately and rebase. It is good to avoid mixing problems.
Can get organized, but we can't move forward because there are no active maintainers now.

PR #1738, opened 2020-06-01
Master sonos

simojenki:
PR1356 with the following;

- Sonos -> Airsonic 'Anonymous/One User' user link with music playing as user 'sonos'
- Sonos -> Airsonic 'Multi-user' link via link-code 
- Liquibase DDL version bumped for 11.0.0

See https://github.com/airsonic/airsonic-docs/pull/96 for documentation on how to configure


airsonic-sonos-api/pom.xml
```diff
@@ -39,7 +39,7 @@
                             <sourceRoot>${project.build.directory}/generated-sources/cxf</sourceRoot>
                             <wsdlOptions>
                                 <wsdlOption>
-                                    <wsdl>${basedir}/src/main/resources/sonos-1.1.wsdl</wsdl>
```
randomnicode, line 42: can probably delete this file

airsonic-main/src/main/java/org/airsonic/player/service/SecurityService.java
```diff
@@ -347,15 +395,111 @@ protected boolean isFileInFolder(String file, String folder) {
         return file.toUpperCase().startsWith(folder.toUpperCase());
     }
 
-    public void setSettingsService(SettingsService settingsService) {
-        this.settingsService = settingsService;
+
+    // =======================================================================================================
+    // Utilities for Sonos link.
```
randomnicode, line 400: These are too Sonos-specific. Shouldn't sonos service handle them?

airsonic-main/src/main/java/org/airsonic/player/service/SecurityService.java
```diff
@@ -347,15 +395,111 @@ protected boolean isFileInFolder(String file, String folder) {
...
+
+    public Authentication authenticate(String username, String password) {
+        UsernamePasswordAuthenticationToken authReq = new UsernamePasswordAuthenticationToken(username, password);
+        return authenticationManager.authenticate(authReq);
     }
 
-    public void setUserDao(UserDao userDao) {
-        this.userDao = userDao;
+    public void authenticate(String sonosLinkToken) throws SonosSoapFault.LoginUnauthorized {
+        SonosLink sonosLink = jwtSecurityService.verifySonosLink(sonosLinkToken);
```
randomnicode, line 408: There is an authenticator for jwts, jwtauthenticationprovider, why not use that?

Or inject an authmanager.

Should we be using the jwt service directly when it has no concept of auth by itself? (it verifies and decodes jwt, without any notion of auth)

airsonic-main/src/main/java/org/airsonic/player/controller/SonosLinkController.java
```diff
@@ -0,0 +1,101 @@
...
+
+        String household = securityService.getHousehold(linkCode);
+        if (household == null) {
+            Map<String,Object> model = newHashMap();
+            model.put("linkCode", linkCode);
+            return new ModelAndView("sonosLinkNotFound", "model", model);
+        }
+
+        try {
+            Authentication authenticate = securityService.authenticate(j_username, j_password);
```
randomnicode, line 82: why not use an auth manager here that is designated to handle this?

airsonic-main/src/main/java/org/airsonic/player/service/SecurityService.java
```diff
@@ -347,15 +395,111 @@ protected boolean isFileInFolder(String file, String folder) {
...
+        return jwtSecurityService.createSonosToken(link.getUsername(), link.getHouseholdid(), link.getLinkcode());
+    }
+
+    public SonosLink getSonosLink(String linkCode) {
+        return sonosLinkDao.findByLinkcode(linkCode);
+    }
+
+    private void setUser(String username) throws SonosSoapFault.LoginUnauthorized {
+        User user = getUserByName(username, true);
+        Authentication authentication = authenticate(user.getUsername(), user.getPassword());
```
randomnicode, line 492: do you need to auth here?
as in what good does fetching a password from the db and using the same password to authenticate against the same db do here? this will always be a positive match, no?

airsonic-main/src/main/java/org/airsonic/player/service/SecurityService.java
```diff
@@ -347,15 +395,111 @@ protected boolean isFileInFolder(String file, String folder) {
...
+        SonosLink sonosLink = sonosLinkDao.findByLinkcode(linkCode);
+        if (sonosLink != null && householdId.equals(sonosLink.getHouseholdid())) {
+            return buildToken(sonosLink);
+        } else {
+            return null;
+        }
+    }
+
+    private String buildToken(SonosLink link) {
+        return jwtSecurityService.createSonosToken(link.getUsername(), link.getHouseholdid(), link.getLinkcode());
```
randomnicode, line 483: i don't think we should have dedicated functions for sonos jwts.
its just a jwt with extra claims

airsonic-main/src/main/java/org/airsonic/player/service/SecurityService.java
```diff
@@ -55,11 +66,37 @@
...
+        Something wrong here, some circular ref, maybe rebuild responsibilities...
+
+        We are :
+           - RESTRequestParameterProcessingFilter need SecurityService cause is UserDetailsService.
+           - But SecurityService need AuthenticationManager they create in GlobalSecurityConfig
+           - But GlobalSecurityConfig create RESTRequestParameterProcessingFilter
+
+           So GlobalSecurityConfig -> RESTRequestParameterProcessingFilter -> SecurityService -> AuthenticationManager (but create in GlobalSecurityConfig)
+     */
+    private AuthenticationManager authenticationManager;
```
randomnicode, line 97: this bit is weird, segregation of responsibilities or maybe a clue stuff shouldn't be organized this way (security service shouldn't have all this?)

airsonic-main/src/main/java/org/airsonic/player/service/JWTSecurityService.java
```diff
@@ -80,4 +88,23 @@ public static DecodedJWT verify(String jwtKey, String token) {
...
         return verify(settingsService.getJWTKey(), credentials);
     }
+
+
+    /**
+     * Create an unexpiring token
+     */
+    public String createSonosToken(String username, String householdId, String linkCode) {
+        return JWT.create()
+                .withClaim(CLAIM_USERNAME, username)
```
randomnicode, line 98: jwt has sub field (subject) for this instead of using a custom "username" claim

airsonic-main/src/main/java/org/airsonic/player/service/JWTSecurityService.java
```diff
@@ -80,4 +88,23 @@ public static DecodedJWT verify(String jwtKey, String token) {
     public DecodedJWT verify(String credentials) {
         return verify(settingsService.getJWTKey(), credentials);
     }
+
+
+    /**
+     * Create an unexpiring token
```
randomnicode, line 94: maybe create an expiring one and refresh it from time to time?

airsonic-main/src/main/java/org/airsonic/player/domain/User.java
```diff
@@ -27,10 +27,14 @@
 public class User {
 
     public static final String USERNAME_GUEST = "guest";
+    public static final String USERNAME_SONOS = "sonos";
```
randomnicode, line 30: why not just use anonymous?

it is Anonymous Link after all

airsonic-main/src/main/resources/applicationContext-sonos.xml
```diff
@@ -0,0 +1,19 @@
...
+    <jaxws:endpoint id="sonosEndpoint" implementor="#sonosService" address="/Sonos">
+        <jaxws:outFaultInterceptors>
+            <bean class="org.airsonic.player.service.sonos.SonosFaultInterceptor"/>
+        </jaxws:outFaultInterceptors>
+        <jaxws:inInterceptors>
+            <ref bean="sonosLinkInterceptor"/>
+        </jaxws:inInterceptors>
+    </jaxws:endpoint>
+
+</beans>
```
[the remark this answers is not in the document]
simojenki: I looked at doing that, however this not a vanilla bean, it somehow gets initialised using the jaxws: xml namespace, which creates it as some sort of webservice bean.  I could see an obvious way to reproduce that in java based Spring config.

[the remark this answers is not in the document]
randomnicode: Airsonic already does that in SonosConfiguration, but that file was deleted in this PR.

Here is airsonic-advanced's copy of the file: https://github.com/airsonic-advanced/airsonic-advanced/blob/master/airsonic-main/src/main/java/org/airsonic/player/spring/SonosConfiguration.java which does it in Java

Thus my suggestion to review https://github.com/airsonic-advanced/airsonic-advanced/pull/318

Issue #804, opened 2018-09-26
Repeated songs in the play queue causes a loop


kauron:
## Problem description

When queuing the same song twice, after playing the second instance, the play queue continues from the first instance. The icon signaling the song that is now playing appears in both songs when any is playing, leading me to think that the position in the queue is not saved, the song that is being played is, leading to being unable to know the current position if the song is repeated.

### Steps to reproduce

1. Enqueue any song
2. Enqueue other songs (any number)
3. Enqueue the song from step 1 again
4. Play the song from step 1
5. Wait until the song from step 3 is played
6. The next song played will be the one after the song from step 1, and not from step 3 as expected.

## System information

 * **Airsonic version**: 10.1.2-RELEASE – July 28, 2018
 * **Operating system**: Arch Linux
 * **Proxy server**: None
 * **Client**: Chromium 69.0.3497.100

ToxicFrog:
This still affects 10.4.0, FYI.

fxthomas:
Probably caused by the fact that the play queue code uses media file `id` values as keys, but picks the first item when switching to a song by `id` (hence, when going to a second identical id, the first is played again).

tesshucom:
Seems to have already been reflected in the implementation. Please close if resolved.

fxthomas:
I don't think this is solved, I can still reproduce it on the latest `master`.

doggogoesbork:
even an easy button remove all duplicates within a playlist (that would have been added by mistake) would be helpful

PR #1751, opened 2020-06-11
Highlight current playing track

Faeb35:
I found it very hard to always have to find the current playing track in the player queue. Specially if you playing a lot of songs (100+) in the queue. I always had to scroll up and down to find the one with the tiny ">>" icon.

With some inspiration from Airsonic-Advanced I highlighted the current playing track with a darker background color. As this is a new color it probably have to be adjusted for other templates.
grafik

Additionally I added a scroll functionality. If you trigger one of the following things, it will automatically scroll to the current playing song. The four (actually 3.5) last played tracks will still be visible before the current track.

Things triggering the scoll:
- Open the player queue (with auto-hide or by pressing Q)
- Seeking in the track
- Start playing (not when pausing): when pressing space key*, when pressing the play button
- Play next track: when the track ends*, when pressing right arrow key*, when pressing the next track button
- Play previous track: when pressing left arrow key*, when pressing the previous track button

*Will not happen if your mouse is currently over the track list in the player queue. This behavior was chosen because it might be very anoying when you do tasks in the player queue an it suddenly starts to scroll/jump to the current track, when a song ends. And you also can skip/pause+play the current track by keyboard when doing tasks in the player queue without having it jumping around. This whole behavior is just considered as proposal. It's what best suits for me, but don't have to suit for everybodys needs. So let me know your opinion about this, but please give it a try first ;-)


airsonic-main/src/main/webapp/WEB-INF/jsp/playQueue.jsp
```diff
@@ -161,11 +165,11 @@
     }
 
     function initAutoHide() {
-        $(window).mouseleave(function (event) {
+        $(".playlistframe").mouseleave(function (event) {
```
Faeb35, line 168: This change was necessary as the eventlistener on the window seems to get triggered several times with Firefox. Also changed in #1740 (was necessary there as well).

tesshucom:
The idea itself is not particularly against.
I think it is better to fix the duplication of the song being played and the bug related to the playback order...? Cumulative additions of features over bugs, even simple ones, complicate things. If we fix a bug later, probably will need to fix this PR as well.
Even if this PR is a good one, it may result in leaving a bad root in future generations.

Faeb35:
You mean the bug, when you have a song multiple times in the play queue, that it marks every of the song as current and always plays the first one? Is there another bug?
I totally agree with what you say. 
I'll try to have a look at this. Lets fix that first.
Probably I can fix that in another PR and rebase this one afterwards. What do you thing? (I'm quit new to Git, need some guidance)

tesshucom:
> Probably I can fix that in another PR and rebase this one afterwards. What do you think?

I think it's a good idea.

[web-playback bags](https://github.com/airsonic/airsonic/issues?q=is%3Aopen+is%3Aissue+label%3A%22in%3A+web-playback%22+label%3A%22type%3A+bug%22+sort%3Aupdated-desc)

Intuitively, I thought #804 was directly related. If I rebase this PR after fixing #804, I think both features will be robust to the user.

tesshucom:
> Probably I can fix that in another PR and rebase this one afterwards. What do you think?

I think it's a good idea.

[web-playback bags](https://github.com/airsonic/airsonic/issues?q=is%3Aopen+is%3Aissue+label%3A%22in%3A+web-playback%22+label%3A%22type%3A+bug%22+sort%3Aupdated-desc)

Intuitively, I thought #804 was directly related. If you rebase this PR after fixing #804, I think both features will be robust to the user.

Faeb35:
#804 should be fixed with #1759
just rebased this change here

PR #1752, opened 2020-06-11
Remove hidden clickable area around user icon in top

Faeb35:
Around the username and the icon in the top frame is an area with a link where shouldn be a link:
grafik

Firstly discovered here https://github.com/airsonic-advanced/airsonic-advanced/issues/222
It also got fixed in Airsonic-Advanced, but with a different approach https://github.com/airsonic-advanced/airsonic-advanced/pull/239

### DIFF
--- a/airsonic-main/src/main/webapp/WEB-INF/jsp/top.jsp
+++ b/airsonic-main/src/main/webapp/WEB-INF/jsp/top.jsp
@@ -132,7 +132,7 @@
         </td>
 
         <td style="padding-left:15pt;padding-right:5pt;vertical-align: middle;width: 100%;text-align: center">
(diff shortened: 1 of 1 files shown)


tesshucom:
Doesn't there exist many others? This may be a side street, but I will leave a note.

 - Some buttons are nested with submit
 - Album art links
 - Playqueue when the status is hidden

If anyone can fix it, I agree. Or, if individual fixes are issued, link to this issue. (It is desirable that all are fixed by a uniform policy. Right?)

Faeb35:
As far as I know that's the only place where we having this issue.
The link in the top menu for the user setting is the only one which is standing separat and centered.

TheFlatworm:
If this is being looked at, can this icon be moved over right next to the LOG OUT button.  I believe that would be a much cleaner look.

Issue #1748, opened 2020-06-09
Cover art stretched when not square


rubenv:
## Problem description

Cover art, when not square, is being stretched and looks out of proportion.

### Steps to reproduce

Take any media file or podcast with a non-square image. It'll be stretched upon resize. Makes people look very weird.

## System information

 * **Airsonic version**: 10.5.0
 * **Operating system**: Linux
 * **Java version**: Apache Tomcat/8.5.43, java 1.8.0_212, Linux
 * **Proxy server**: Nginx
 * **Client**: Firefox
 * **Language**: EN

PR #1755, opened 2020-06-12
Crop image instead of strechting

Faeb35:
This PR is fixing #1748
Before resizing the image it will be croped first to be square. This is prevent it from being strechted.

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/controller/CoverArtController.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/controller/CoverArtController.java
@@ -333,8 +333,17 @@ public class CoverArtController implements LastModified {
     public static BufferedImage scale(BufferedImage image, int width, int height) {
         int w = image.getWidth();
         int h = image.getHeight();
-        BufferedImage thumb = image;
 
+        //crop image to correct size
+        if (w > h) {
+            int cropX = (w - h) - ((w - h) / 2);
+            image = image.getSubimage(cropX,0, h, h);
+        } else if (h > w) {
+            int cropY = (h - w) - ((h - w) / 2);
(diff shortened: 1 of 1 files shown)


Faeb35:
> I don't think cropping the image is the right thing to do, because that then presents the problem of what portion of the image is kept, and it is impossible to always make the right choice. 

It's always taking the center part of the image and croping the edges. I've seen this behavoir quite a lot in other implementations. Like when I'm using DSub the cover is used as background of the lockscreen. It's doing the same thing there.

I prefered this behavior, as it results in having all the covers looking equaly instead of having transparent spaces on two sides on some of the covers. I think usually the covers already should be almost square, so it shouldn't crop off too much and wouldn't even be noticable. But if you add some transparent space it's defenitly "sticks out" (not sure if this is the right therm in english).

eharris:
I disagree.  In my library, I have lots of albums that come from DVD-Audio sources, and cover images from DVD cases are nowhere close to square.  Applying your cropping method would chop off the entire title of the album in many cases.  The current method of stretching is vastly preferable to simple cropping because it maintains the full fidelity of the image, although at the cost of a distorted aspect ratio.

Faeb35:
Ok I see, we need to have all three options and a setting to choose from these options to make everybody happy :-)

tesshucom:
Which looks right depends on how the screen is implemented. If Width and Height are displayed as fixed, eharris's claim is valid. Faeb35's opinion, on the other hand, has the advantage of using images that preserve the original size ratio. Probably the user thinks either method is convenient.

eharris's opinion goes well with tiles. The Faeb35 method may be better for special situations like "zooming". Apart from selection, there is also the idea of ​​using different methods depending on the location.

PR #1759, opened 2020-06-17
Fix issue with multiple songs marked as current

Faeb35:
This pull request is fixing  #804

### DIFF
--- a/airsonic-main/src/main/java/org/airsonic/player/ajax/PlayQueueInfo.java
+++ b/airsonic-main/src/main/java/org/airsonic/player/ajax/PlayQueueInfo.java
@@ -31,6 +31,7 @@ import java.util.List;
 public class PlayQueueInfo {
 
     private final List<Entry> entries;
+    private final int index;
     private final boolean stopEnabled;
     private final boolean repeatEnabled;
     private final boolean shuffleRadioEnabled;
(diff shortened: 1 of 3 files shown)


airsonic-main/src/main/java/org/airsonic/player/domain/PlayQueue.java
```diff
@@ -399,9 +409,6 @@ public synchronized Status getStatus() {
      */
     public synchronized void setStatus(Status status) {
         this.status = status;
-        if (index == -1) {
```
Faeb35, line 402: I'm not quite sure if removal of this line could cause any side effects. I had to remove this to fix the following circumstances:
1. Play a song
2. Clear the playlist
3. Add an album (just click on the + on the cover or use the "Add All" link, don't press play) 
4. First song got marked as current, even though it's not the one which is currently played by the controls, the controls were still playing the last song which isn't in the playlist anymore after clearing it

The method addFilesAt is invoking setStatus which was always resetting the index to 0 even though it should be -1.

airsonic-main/src/main/java/org/airsonic/player/domain/PlayQueue.java
```diff
@@ -399,9 +409,6 @@ public synchronized Status getStatus() {
...
-            index = Math.max(0, Math.min(index, size() - 1));
-        }
     }
```
Faeb35, line 412: I'm not quite sure if removal of this line could cause any side effects. I had to remove this to fix the following circumstances:

    Play a song
    Clear the playlist
    Add an album (just click on the + on the cover or use the "Add All" link, don't press play)
    First song got marked as current, even though it's not the one which is currently played by the controls, the controls were still playing the last song which isn't in the playlist anymore after clearing it

The method addFilesAt is invoking setStatus which was always resetting the index to 0 even though it should be -1.

PR #1790, opened 2020-08-31
Fix CORS

tamland:
Currently CORS headers aren't always added to responses e.g. on authentication failure, which means client won't be able to read the error response.


randomnicode:
Hey this fork seems not actively being worked on, do you want to push this to [airsonic-advanced](https://github.com/airsonic-advanced/airsonic-advanced) instead?

